Issue #1, opened 2022-10-09
Web map with selectable tasks in mobile map view


ivangayton:
Working Web map in mobile view on Android device, from which users can:
- See available tasks
- Click on tasks, triggering a mechanism allowing them to select the task and launch ODK Collect with the appropriate project, app user, and form configured.

krtonga:
With the merge of the `fast-api` branch, this Issue is:

Ugly implemented in the UI. 

Known issues:
- popup buttons don't fit in their dialogs, and will crash the app if clicked. 
- popups do not change to only show actions for "logged in user" who has locked a task.
- task list below the map is very rough and does not contain all information. 
- qr codes and other task detail is not yet sent to the front end.

PR #12, opened 2022-10-23
Display task status

krtonga:
The following PR is related to the functionality needed for [Github issue #1: Web map with selectable tasks in mobile map view](https://github.com/hotosm/fmtm/issues/1). 

Features include:
- display a web map that supports 3 different task statuses matching color from Tasking Manager - unavailable, available and ready for validation (note: ready for validation is not yet visible from the normal user flow)
- allow logged in user to select a task (changing it's status from available to unavailable) and download the QR code
- logged in users should be able to download the qr codes for unavailable tasks that they selected, but not tasks that other users selected

Other enhancements include:
- Automatically zoom the map to the grid.geojson area
- Removal of unused methods

This does not include:
- mobile friendly map view


robsavoye:
Since this doesn't have a dependency on odk_fieldmap, to keep the code organized, can you put this in a different top level directory ? Preferably 'src'.

krtonga:
@robsavoye My last commit implements the following structural changes:

New file structure:
> `scripts`
>> `odk-fieldmap-original/utils` (includes @ivangayton odk/geojson scripts for project creation)

> `src`
>> `api` (currently empty)
>> `web` (current flask-web app)

PR #49, opened 2023-01-27
form creation handles xform reference to different geojson files and …

ivangayton:
…arbitrary row for select_from_file question.

Now the tasks2forms module finds and replaces the select_one_from_file no matter where it's placed, and additionally any references to information in the geojson (for pre-populating fields and fetching geometry) get updated with the appropriate filename to refer to the specific task the form is associated with.


krtonga:
Ivan's instructions to generate forms:

Steps:
1. Navigate to the fmtm git repo on your local
2. Activate the venv (source venv/bin/activate) (need to ensure you have openpyxl and segno installed, which should be the case if the requirements.txt has been installed by pip)
3. use tasks2forms, with two arguments:
  a. Root folder of the project
  b. The xlsform template
4. Now you should have a folder full of xforms
5. run forms2server with 4 arguments:
  a. Root folder of the project
  b. ODK server URL
  c. ODK server username
  d. ODK server password

krtonga (approved):

PR #50, opened 2023-01-27
Fixed problem with docker for api after test-server was merged

krtonga:


robsavoye:
This is needed so main works for us mere mortals.

PR #51, opened 2023-01-31
Script to download all attachments from all forms in an ODK project

ivangayton:
This pulls all of the submissions from all of the forms in an ODK project. For now it stores them as the zip files provided by ODK Central; I'll update this later to extract the zips into CSV (which will of course be multiple CSV if there are repeats in the form; that's why I didn't do it yet), and optionally collate all of the submissions in a project to a single CSV, which might be useful for viewing in QGIS.

### DIFF
--- a/old_scripts/odk_fieldmap_original/utils/dl_attachments.py
+++ b/old_scripts/odk_fieldmap_original/utils/dl_attachments.py
@@ -1,0 +1,62 @@
+# Copyright (c) 2020, 2021, 2022 Humanitarian OpenStreetMap Team
+# This file is part of FMTM.
+#
+#     FMTM is free software: you can redistribute it and/or modify
+#     it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


old_scripts/odk_fieldmap_original/utils/dl_attachments.py
```diff
@@ -0,0 +1,62 @@
+# Copyright (c) 2020, 2021, 2022 Humanitarian OpenStreetMap Team
```
krtonga, line 1: As Rob pointed out, these years are wrong.

krtonga (approved):

old_scripts/odk_fieldmap_original/utils/dl_attachments.py
robsavoye: While it's possible to fix all the copyright dates using a simple sed script, I usually just do it for the files I'm working on.

PR #52, opened 2023-01-31
Endpoints

robsavoye:
This are just stubbed out endpoints with no functionality yet. One of the big new sections are the /debug endpoints, which I'll be using to test the functionality of the backend. They will be deleted once the initial support is working.


src/backend/db/database.py
```diff
@@ -21,7 +21,7 @@
 from sqlalchemy.orm import sessionmaker
 
 # TODO don't hardcode this
-SQLALCHEMY_DATABASE_URL = 'postgresql+psycopg2://fmtm:fmtm@db:5432'
+SQLALCHEMY_DATABASE_URL = 'postgresql+psycopg2://fmtm:fmtm@localhost:5432/fmtm'
```
krtonga, line 24: This probably shouldn't be checked in, as it will not work with the docker deployment.

src/backend/debug/debug_crud.py
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2023 Humanitarian OpenStreetMap Team
```
krtonga, line 1: Why is this called debug?

krtonga (changes requested):

src/backend/db/database.py
robsavoye: actually it works just fine with postgresql in docker or not in docker. The backend will mostly likely only be talking to localhost anyway.

src/backend/debug/debug_crud.py
robsavoye: It's called debug cause it's only for debugging. Some of the higher level endpoints need to execute several backend support functions. This lets me test them in isolation while debugging their initial implementation. Once the support code is considered stable, each endpoint used for that will get deleted.

PR #53, opened 2023-02-02
Unzip, collate submissions, expand geopoints

ivangayton:
The dl-submissions module now:
- Unzips submissions from each form, 
- checks for CSV with more than a header row
- checks for an ODK GeoPoint-type column, expands it into 4 additional columns for easy import into QGIS or direct conversion to GIS files like GeoJSON
- Writes the individual CSV files to a specified directory (not in a Zip archive)
- Writes a collated CSV for the entire project
  - WARNING: this will probably do something horrible if there are different types of surveys in a project; it's only sensible at the moment if it's a single survey split amongst multiple geographies (as we're basically doing now with the FMTM)
- If any media attachments, creates the subdirectory specified in the Zip archive for them and writes them (again outside of the archive). 

## TODO:
- The collation and geopoint expansion happen automatically, should be flags to do or not.
- The function to simply download and write the Zip archives exists but is not accessible (again should be a flag, i.e. -z or whatever)
- The geopoint expansion _will_ go wrong with multiple survey types in a single project

### DIFF
--- a/old_scripts/odk_fieldmap_original/utils/dl_submissions.py
+++ b/old_scripts/odk_fieldmap_original/utils/dl_submissions.py
@@ -26,6 +26,16 @@
 
     And downloads all of the submissions from that server as CSV
 
+    TODO (KNOWN BUGS): 
+    - For now it expects a project with multiple form, but all basically
+      identical (a single ODK survey with different GeoJSON forms).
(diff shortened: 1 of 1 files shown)


robsavoye:
Since this is standalone, it won't break anything. This code does need to be merged into the actual backend, but that's another task.

PR #56, opened 2023-02-02
Update import paths after PRs merged

robsavoye:
These is needed to fix minor breakage.


ivangayton:
Hmm, this looks dangerous. Python's import module is a bit of a minefield. Should we wait to merge this until we have the full deployment sorted and documented?

robsavoye:
On 2/2/23 17:42, Ivan Gayton wrote:
> Hmm, this looks dangerous. Python's import module is a bit of a 
> minefield. Should we wait to merge this until we have the full 
> deployment sorted and documented?

   All the import paths got changed in the recent PRs, but most projects 
don't do relative imports either. While security is super important, and 
should be  considered during all stages of software development, for a 
few weeks we need to just get it working. There is nothing on the dev 
server but open data and open source anyway.

	- rob -
-- 
Senior Tech Lead
Humanitarian OpenStreetMap Team
https://www.hotosm.org

ivangayton:
Ok perfect!

PR #57, opened 2023-02-03
Frontend/project details

Mudi-business:
PR, on frontend project details page , theming and project structure.


emi420 (approved):
Theming is better managed and structure is more clear now, thanks @Mudi-business !

PR #62, opened 2023-02-06
Qrcode

robsavoye:
Implement creating a QR Code, and writing it to the database. Add initial support for odkconvert to talk to ODK Central.


src/backend/tasks/tasks_crud.py
```diff
@@ -150,6 +150,25 @@ def create_task_history_for_status_change(
...
 
+def create_QRCode(self, project_id=None, token=None, name=None):
+    """Get the QR Code for an app-user"""
+    base = "fixme"
+    url = f'{base}key/{token}/projects/{project_id}'
+    logger.info(f"Generating QR Code for app-user \"{name}\" for project {project_id}")
+    settings = {"general":
+                {"server_url":f'{base}key/{token}/projects/{project_id}',
+                 "form_update_mode":"manual",
+                 "basemap_source": "OpenStreetMap",
```
ivangayton, line 161: Last I checked, ODK Collect has a bug which makes long-press to select a location fail when OpenStreetMap is selected as the basemap. This is fine when choosing points with select_from_file, but if you have to add a new point, you can't do it anywhere except where your GPS thinks you are. 

Workaround: use MapBox as the basemap. It's the OSM features, so corresponds nicely to our extracts, but doesn't block long-press point creation. It's lighter anyway because it's vector tiles.

ivangayton (changes requested):

PR #78, opened 2023-02-16
Create all output files for ODK Central

robsavoye:
This implements the process of generating QR codes, app users, XForms, and data extracts needed for the ODK Central FMTM interface. To use this you must have first created the project, and uploaded the project polygon. Uploading the project polygon automatically generate the task polygons. This then uses what's in the task database table to access the task ID and the boundary, and generate all the unique files.


src/backend/central/central_routes.py
```diff
@@ -20,10 +20,12 @@
 from sqlalchemy.orm import Session
 import logging.config
 import json
+import epdb
```
ivangayton, line 23: Huh. Never come across this extended debugger before. No problem, but curious what you're using this for? It does add a dependency...

src/backend/models/enums.py
```diff
@@ -16,8 +16,23 @@
 #     along with FMTM.  If not, see <https:#www.gnu.org/licenses/>.
 #
 
-from enum import Enum, IntEnum
-
+from enum import Enum, IntEnum, StrEnum
+
+
+class category(StrEnum, Enum):
```
ivangayton, line 22: Should these specific categories eventually live in some kind of external file rather than within a class definition? I'm particularly thinking of users who aren't immersed in the standard OSM world (scientists come to mind), who may be looking for data that don't fall within these categories.

src/backend/projects/project_crud.py
```diff
@@ -19,19 +19,37 @@
...
+from json import loads, dumps, dump
 from typing import List
 from zipfile import ZipFile
+import epdb
 from fastapi import HTTPException, UploadFile
 from geoalchemy2.shape import to_shape
-from geojson_pydantic import FeatureCollection
-from shapely.geometry import mapping, shape
+import geoalchemy2
+import numpy as np
```
ivangayton, line 30: Well dang, there it is. I always try to avoid bringing in Numpy, Pandas, and friends as long as possible, but I guess we've hit that point where they inevitably show up!

src/backend/projects/project_crud.py
```diff
@@ -19,19 +19,37 @@
...
+from pyxform.xls2xform import get_xml_path, xls2xform_convert
+import xmltodict
+
+
+from geojson_pydantic import FeatureCollection, Feature
+from shapely.geometry import mapping, shape, Polygon
```
ivangayton, line 36: I strongly suspect we'll need to replace this with PostGIS functionality, but for now it's fine to get things into basic working shape.

src/backend/projects/project_crud.py
```diff
@@ -315,29 +396,175 @@ def update_project_with_upload(
...
 
-def create_task_grid(projectid: int, error_detail: str):
+def read_xlsforms(
+    db: Session,
+    directory: str,
+):
+    """Read the list of XLSForms from the disk"""
+    xlsforms = list()
+    for xls in os.listdir(directory):
+        if xls.endswith(".xls"):
```
ivangayton, line 410: Many, many people (including me) use .xlsx files for xlsforms. I forget exactly what it was, but I encountered some limitation in the .xls spec for some ODK data once (it may have been submissions rather than xlsforms), and since then I've avoided the .xls format altogether.

src/backend/projects/project_routes.py
```diff
@@ -71,8 +79,10 @@ async def read_project(project_id: int, db: Session = Depends(database.get_db)):
 @router.post("/delete/{project_id}")
 async def delete_project(project_id: int, db: Session = Depends(database.get_db)):
     """Delete a project from ODK Central and the local database"""
-    # FIXME: should check for errors
+    # FIXME: should check for error
     odkproject = central_crud.delete_odk_project(project_id)
+    # if not odkproject:
```
ivangayton, line 84: Nice to see some error handling here; very much appreciated. It's too easy to not bother dealing with such cases (an attempt to manage an asset that doesn't exist on a remote machine), and we've definitely been guilty of putting off this kind of housekeeping on this project so far. Kudos.

src/backend/projects/project_routes.py
```diff
@@ -87,23 +97,64 @@ async def create_project(
...
     project_id: int,
     project_name_prefix: str,
     task_type_prefix: str,
     upload: UploadFile,
     db: Session = Depends(database.get_db),
 ):
+    """This uploads the projecy boundary polygon to an existing project"""
     # TODO: consider replacing with this: https://stackoverflow.com/questions/73442335/how-to-upload-a-large-file-%e2%89%a53gb-to-fastapi-backend/73443824#73443824
-    project = project_crud.update_project_with_upload(
+    project = project_crud.update_project_with_zip(
```
ivangayton, line 121: I assume this is going away soon?

ivangayton (approved):

src/backend/projects/project_crud.py
ivangayton: @robsavoye this one is pretty important I think. A _lot_ of people have libraries of xlsforms in .xlsx format. Go ahead and merge, but please keep in mind that this could be a really frustrating limitation for anyone modifying forms or providing their own.

src/backend/central/central_routes.py
robsavoye: I use it all the time, as it lets me set a breakpoint in the backend, which you can't do from the command line. This will go away at some point, but right now there's a lot of debug left to do. Epdb is better than the default pdb. Kristen uses ipdb, epdb is just what I'm used to. It's a stable dependency, but not a critical one.

src/backend/models/enums.py
robsavoye: This seems to be how FastAPI works, I just did what already existed. This just adds the ones for xlsforms.

src/backend/projects/project_crud.py
robsavoye: Pandas we're not using geoalchemy and  numpy were already being used, just not by this file yet.

src/backend/projects/project_crud.py
robsavoye: Actually some geometry manipulation needs to be done outside of postgis.

src/backend/projects/project_crud.py
ivangayton: Ah, got it! Glad to hear we've not yet been invaded by black and white bears! (I'm sure they'll show up eventually)

src/backend/projects/project_crud.py
robsavoye: Interesting, as I've had the opposite results with xls vs xls. I can extend the code to deal with both.

src/backend/central/central_routes.py
ivangayton: Ok, fantastic.

src/backend/projects/project_routes.py
robsavoye: Not for a while on the backend, as it's what currently works. Now though with this PR there is a different process. which is automated. In the future it'll still be possible to upload the individual files, but the zip will be unneeded. But at this point there is no reason to depreciate working code.

src/backend/projects/project_routes.py
robsavoye: There's not a whole lot of error handling in this PR, that'll be the next pass. This was just getting it to work.

src/backend/projects/project_crud.py
robsavoye: I uses xmltodict instead of pandas.

src/backend/projects/project_routes.py
ivangayton: Nevertheless it's encouraging to see some of it happening already.

src/backend/projects/project_crud.py
ivangayton: I'd appreciate that; obviously working with either/or would be preferable.

src/backend/projects/project_crud.py
ivangayton: Really? I'm surprised by that; as far as I can see PostGIS can do pretty much any geoprocessing you like, and make you a tuna sandwich as well. If this is the case it has implications for my splitting algorithm bounty project...

PR #79, opened 2023-02-17
Fix breakage for docker-compose

robsavoye:
Incorporating odkconvert required changes to get the current main branch working again with docker-compose. With this patch, docker-compose works again, although still brings up the old Flask UI.


ivangayton (approved):
I can't assess this easily without a working deployment strategy, but it looks reasonable and it probably can't make it worse (and probably can make it better)!

PR #80, opened 2023-02-19
Look for both xlsx and xls files

robsavoye:
Get both spreadsheet versions.


ivangayton (approved):
This seems to contain a lot of authorization housekeeping as well as the change looking for either kind of Excel file! I'll assume those housekeeping changes are okay, and enthusiastically endorse this spreadsheet format addition!

PR #81, opened 2023-02-19
Add support for file I/O from database

robsavoye:
This patch allows for the download of both the project boundary polygon, as well as all the task boundaries as GeoJson. This can be used for display, or with frontend support, downloaded to disk. This patch also support upload a custom XLSForm to the database, adding it to the XLSForm library. If the name matches an existing XLSForm, then it gets updated instead.


ivangayton:
That's great! Hard-coding the form names was ok as a temporary measure, but this is much better.

PR #82, opened 2023-02-20
initial microfrontend

Mudi-business:
I have re restructured the project in modules the using module federation microfrontend architecture and I have finished the basic functionalities of openlayers map which will be displaying map, select task , update task status with little less styles I just focused more in functionality so far and there are some few bugs that I'm about to finish soon.


ivangayton (approved):
This is far too much for me to actually review in terms of its functionality. However, I can see that it's not stepping on anything else, it's safely confined it to its own files and directories. Therefore merging. Thanks for the great work Mohdi!

PR #85, opened 2023-02-20
List project and task geometries

robsavoye:
This was broken due to some database columns that I don't think are actually needed. With this patch it's possible to list projects and tasks again.


ivangayton (approved):
It looks like I'm saying pretty similar changes but the authorization infrastructure in the several of these pull requests, but if this is the most recent one and it's working, in it goes!

Frankly, I think it's likely that we're going to have a bit of cleanup to do after this batch of merges, but I think that's preferable to the alternative of a bunch of branches, and of course can always revert something if it turned out to be catastrophic.

PR #87, opened 2023-02-21
Format codebase using pre-commit hooks (black, ruff)

spwoodcock:
**This PR is entirely formatting related - no code was edited**

- In a previous PR I added a pre-commit config for pre-commit hooks. To use:
```bash
pip install pre-commit
pre-commit install
```
Then all of the files in any commit you make will have the hooks ran on them (formatted, linted, sorted) automatically, prior to committing.

- In this PR I ran black against all Python code, prettier for markdown and yaml files.
- The goal was to have a fully formatted base to start editing from, as we don't want to muddy commits with formatting old code further down the line.
- Everything apart from src/frontend was formatted (as I don't want to cause merge conflicts with those working on the frontend).


spwoodcock:
Oh, I forgot to mention! There is a commit in there to remove the src/web demo flask app, as we are making progress on the React frontend.

Issue #84, opened 2023-02-20
Improve Github Project Documentation


petya-kangalova:
[HOTOSM is taking part in Outreachy internship May to August 2023](https://www.outreachy.org/outreachy-may-2023-internship-cohort/communities/humanitarian-openstreetmap-team/). I will be adding suggested tasks for interns as issues and tagging with "Outreachy".

Suggested list of tasks for Outreachy contributors [can seperate  into individual issues]
- [ ] 1.1. Research best practices on Github product  documentation, including system and process documentation (ex. contributor guidelines, testing, translation, API documentation, etc.) For example, see example of API endpoint with description that is missing from other endpoints  https://github.com/hotosm/fmtm/blob/ccf3719663c11db599163b38e0f7889e277f529a/src/backend/projects/project_crud.py#L216 We would also need support with creating test cases for API endpoints.
- [ ] 1.2. Based on the research, suggest improvements to existing documentation starting with the [Field Mapping Tasking Manager Github repo](https://github.com/hotosm/fmtm).
- [ ] 1.3. Create new documentation where missing for [Field Mapping Tasking Manager Github repo](https://github.com/hotosm/fmtm).
- [ ] 1.4. Suggest a consistent documentation structure that can be replicated for other pinned repositories in the HOTOSM Github

harshithakolipaka:
hey, @petya-kangalova, I found the feedback form missing in the current documentation.So, I would like to add a user feedback form on the documentation and the product itself. This can help identify areas for improvement and can lead to a better overall user experience.

Ndacyayisenga-droid:
Hi @spwoodcock @petya-kangalova can I work on this issue?

spwoodcock:
Hi @harshithakolipaka, great idea! I'm not sure exactly where the feedback form should submit to, but @petya-kangalova might have a better idea.

@Ndacyayisenga-droid you can most definitely help by contributing anything you see missing from the docs!

Ndacyayisenga-droid:
@spwoodcock @petya-kangalova am going to do 1.1, 1.2 and 1.4 on google docs then share the link
I would probably send a PR for 1.3

Does that sound like a good Idea?

Ndacyayisenga-droid:
> [HOTOSM is taking part in Outreachy internship May to August 2023](https://www.outreachy.org/outreachy-may-2023-internship-cohort/communities/humanitarian-openstreetmap-team/). I will be adding suggested tasks for interns as issues and tagging with "Outreachy".
[...]

PR #90, opened 2023-02-22
Add Github workflow to deploy /docs --> wiki

spwoodcock:
- Triggers when there is a push to main (i.e. merged PR) and the `/docs` dir is changed.
- Syncs content of /docs with wiki, then pushes to wiki using service user access token.

**Note**: for this to work, we need to create a service user for the repo.
I propose a user `svcfmtm` with an email `fmtm@hotosm.org` (does this email exist?).
Then we need to create an access token for that user, and place it in repository secret variable `SVC_ACCESS_TOKEN`.

### DIFF
--- a/.github/workflows/wiki.yml
+++ b/.github/workflows/wiki.yml
@@ -1,0 +1,41 @@
+name: Publish Docs to Wiki
+
+on:
+  push:
+    paths:
+      - docs/**
+    branches: [main]
+
+env:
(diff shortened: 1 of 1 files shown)


spwoodcock:
I updated to use the default GITHUB_TOKEN instead of a service account - worth a test to see if it has the necessary permissions.
This PR can be merged.

spwoodcock:
Related to #84.
This workflow allows for the auto-deployment of docs to the wiki.
Docs can be updated via PR and traced easily using Git.

PR #91, opened 2023-02-22
Add 8081 to allowed origins

krtonga:
This adds 8081 to allowed origins so that react openlayers can be run locally and talk to a remote instance of the fmtm api.

### DIFF
--- a/src/backend/main.py
+++ b/src/backend/main.py
@@ -39,7 +39,8 @@ from .central import central_routes
 os.environ["OAUTHLIB_INSECURE_TRANSPORT"] = "1"
 
 # setup loggers
-log_file_path = path.join(os.path.dirname(os.path.abspath(__file__)), "logging.conf")
+log_file_path = path.join(os.path.dirname(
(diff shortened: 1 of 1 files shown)


spwoodcock:
Hi Kristen!

This is perfect for testing, but I think at some point we should make the CORS origins configurable via environment variable
What do you think?

Although providing many origins in a .env can be cumbersome:
```dotenv
BACKEND_CORS_ORIGINS='["http://localhost","http://localhost:8080","http://localhost:8080","http://localhost:8081"]'
```

I think we should eventually refactor the settings to use Pydantic BaseSettings & then add a validator for the CORS origins (we could provide them as a comma separated string, instead of with the square brackets:
```dotenv
BACKEND_CORS_ORIGINS=http://localhost,http://localhost:8080,http://localhost:8080,http://localhost:8081
```

robsavoye:
Eventually being the key word, but it sounds like this is needed now, and won't break anything.

Issue #76, opened 2023-02-15
DevOps: Set up CI workflow for testing


petya-kangalova:
Set up a Github action to test the software after every push to a branch.

Backend /done
Frontend /tbc

nrjadkry:
@varun2948 We have frontend tests integrated now. Right?

varun2948:
Integrated Frontend Test aswell

Issue #69, opened 2023-02-14
endpoint for listing projects


robsavoye:
There needs to be an endpoint to list all projects, and to search through existing projects.

spwoodcock:
Complete in #110

PR #93, opened 2023-02-22
Configure pytest to work with github workflow

spwoodcock:
Partially completes #76 (backend).
Frontend testing via workflows still required.


spwoodcock:
Hold on merging this, I need to make the tests work first!

spwoodcock:
PyTest should now work.
The problem is that the tests are all broken (starting from the conftest.py).
I'll create a separate PR fixing the tests.

spwoodcock:
Looks like from the latest run this fixed the issue.
The error is related to the broken test (instead of missing the pytest dependency).

spwoodcock:
- I also tacked onto this PR a commit that removed the old tests - they were all for the demo flask frontend & were failing.
- The PyTest workflow should now pass - I added a dummy test to make it work.

spwoodcock:
- There was a project.listProjects() called during app startup - attempting to call the odkcentral server & making tests fail - removed it.
- Partial implementation of #69 instead (not fully tested).

robsavoye:
There's a conflict with this patch.

spwoodcock:
Fixed - was due to the cors origins update.

PR #97, opened 2023-02-23
Microfrontend production deployment, odkcentral api access via proxy and self-signed certs (https)

spwoodcock:
- Didn't realise we also have the DNS entry for fmtm-api.hotosm.org, so I decided to place the API on that instead of subpath `/api`.
- Splitted API_DOMAIN and FRONTEND_DOMAIN env vars as a result.
- Also updated the default frontend build from `main` to `fmtm_openlayer_map`.


spwoodcock:
I'm hoping that merging this PR will also test the wiki.yaml workflow to build the docs.

spwoodcock:
Just a sec on merging this - I'm updating something on the frontend deploy (may as well include).

PR #100, opened 2023-02-24
Fix populating data from zip file

robsavoye:
Until the entire automated process is fully working, it's still necessary to populate the database from a zip file.


ivangayton (approved):
This looks like it's just reactivating Kristen's temporary system, which is fine for now!

PR #102, opened 2023-02-27
Update FastAPI init to use Pydantic BaseSettings

spwoodcock:
- Pydantic BaseSettings is the default way to load .env values with FastAPI
- It it more configurable, provides validation options, and more useful error messages.


spwoodcock:
I also added a couple of tweaks to the docker setup.

PR #109, opened 2023-02-28
Download submissions from ODK Central and convert them to OSM XML for JOSM

robsavoye:
This does the entire post data collection process of downloading and converting the data into a form suitable to edit in JOSM or QGIS.


src/backend/central/central_crud.py
```diff
@@ -23,10 +23,14 @@
 from sqlalchemy import table, column
 
 import os
+import csv
 from pyxform.xls2xform import xls2xform_convert
 import xmltodict
+import pathlib
```
ivangayton, line 29: I guess I'd better get with the times (I've not yet switched from os.path to pathlib, but it's fairly obviously a good idea!)

src/backend/central/central_crud.py
```diff
@@ -164,18 +195,72 @@ def generate_updated_xform(
...
+):
     """Upload a data file to Central"""
     filename = "test"
     xform.getMedia(project_id, xform_id, filename)
+
+def convert_csv(
+    filespec: str,
+    data: bytes,
+):
+    """Convert ODK CSV to OSM XML and GeoJson"""
```
ivangayton, line 230: Awesome, very happy that we're offering GeoJSON right away (not everyone wants to deal with OSM XML).

src/backend/central/central_routes.py
```diff
@@ -18,24 +18,23 @@
...
+from sqlalchemy.sql import text
+from sqlalchemy import (
+    select,
+    table,
+    column,
+    insert,
+    inspect,
+)
+
+import sqlalchemy
```
ivangayton, line 30: I'm slightly confused as to why you import a bunch of specific functions you'll use from a module, and then import the whole module anyway. No big deal, just curious why you do it that way?

src/backend/central/central_routes.py
```diff
@@ -65,28 +64,67 @@ async def create_appuser(
...
+        "projects", column("project_name_prefix"), column("xform_title"), column("id"), column("odkid")
+    )
+    where = f"id={project_id}"
+    sql = select(project).where(text(where))
+    result = db.execute(sql)
+    first = result.first()
+    if not first:
+        return {"error": "No such project!"}
+    # FIXME: this should be configurable
+    tmp = "/tmp"
```
ivangayton, line 91: Are you using /tmp in the local filesystem to store data?

ivangayton (approved):
Ok, looks good; couple of things I'm curious about (see comments) but nothing urgent and certainly nothing to wait on before merging.

robsavoye:
On 2/28/23 18:24, Ivan Gayton wrote:

> I guess I'd better get with the times (I've not yet switched from 
> os.path to pathlib, but it's fairly obviously a good idea!)

   This was needed to find the directory when odkconvert is installed as 
a pip package. The xforms.yaml file is needed for ODK->OSM conversion.

> Awesome, very happy that we're offering GeoJSON right away (not everyone 
> wants to deal with OSM XML).

   All data goes into the GeoJson file, as not everything is suitable 
for OSM.

> I'm slightly confused as to why you import a bunch of specific functions 
> you'll use from a module, and then import the whole module anyway. No 
> big deal, just curious why you do it that way?

   Cut & paste from existing code. We can trim it down later.

> Are you using /tmp in the local filesystem to store data?

  The pyxform function only supports writing disk files, so they had to 
go someplace. Submissions are now done without disk files with this 
patch. The can optionally write the CSV to disk though. Same for data 
extracts. They get written to disk for now, but pretty soon that will 
also all be done in memory. I try to avoid disk based temp files as much 
as possible.

	- rob -
-- 
Senior Tech Lead
Humanitarian OpenStreetMap Team
https://www.hotosm.org

PR #115, opened 2023-03-01
Test zips

ivangayton:
Added a directory with three zip files known to work with the Naivasha-era demo server.


spwoodcock (approved):
All zips look good to me - I added them into https://fmtm-api.hotosm.org successfully.

PR #120, opened 2023-03-02
Create OSM data extract, upload XForm and attachment to Central

robsavoye:
This patch uploads the XForm, modifies it's draft form status, uploads the data extract from OSM, and then publishes the result.


ivangayton (approved):
Far as I can tell this doesn't do anything dangerous for the parts of the API that Mohdi and others are using, but I had a little trouble figuring out a merge conflict, so I'm going to merge anyway and see how badly that goes.

PR #124, opened 2023-03-05
advance map, qrcodes and project details info

Mudi-business:


emi420 (commented):
Long PR but good work @Mudi-business ! 

My only comment is if it's needed to include both Bootstap and MUI?

Mudi-business:
well if you check `https://fmtm.hotosm.org/` you'll notice that the UI is a  bit not functioning well and I've notice it's because of the openlayers project inherits MUI components from main project but not having the MUI package installed in it  and currently I'm not sure if I can inherit packages from main directly but I'll look more into it and see what I can improve for that , but for bootstrap  I don't have any use for it now in openlayers project I just putted it there for the same reason  to avoid UI breaking down. but I'll will take a look at it more in depth if module` federation archtecture `allows to Inherit packages too then it won't be necessary to duplicate those packages in the openlayers project. @emi420

Issue #83, opened 2023-02-20
Explore adding AllContributors Bot


petya-kangalova:
**[HOTOSM is taking part in Outreachy internship May to August 2023](https://www.outreachy.org/outreachy-may-2023-internship-cohort/communities/humanitarian-openstreetmap-team/). I will be adding suggested tasks for interns as issues and tagging with "Outreachy".**

Is your feature request related to a problem? Please describe.
This is not related to specific FMTM feature. It will be useful to explore adding the All Contributor Bot to this repository: https://allcontributors.org/docs/en/bot/installation

Describe the solution you'd like
We would like to reward contributors to this repository, and not just those contributing with code.

Describe alternatives you've considered
I haven't see many alternatives to this- recently came across it! I would love to hear if others in the community have used it, installed in their repo? It looks like you need to tag people manually.

Ndacyayisenga-droid:
@petya-kangalova @spwoodcock can I work on this issue?

spwoodcock:
Hi @Ndacyayisenga-droid thanks for your willingness to contribute! Feel free to start and send a PR 😄

Ndacyayisenga-droid:
@spwoodcock can you take a look at mp PR please https://github.com/hotosm/fmtm/pull/132

AugustHottie:
Hi @petya-kangalova @spwoodcock it doesn't look like there's an Issue regarding rewarding open source contributors. Can I create the issue?

petya-kangalova:
@Ndacyayisenga-droid  thanks for the PR- will take a look by end of the week. @robsavoye 

@AugustHottie  good spot- created one here https://github.com/hotosm/fmtm/issues/135

spwoodcock:
Merged #132 as I thought it was a simple config, but the allcontributions bot didn't seem to run.
This needs to be re-vistited at some point.

spwoodcock:
This was my lack of understanding of how the bot works. It seems it's functioning: it will add a PR with a contributor if you comment in the correct format, see https://github.com/hotosm/fmtm/issues/135#issuecomment-1459496380

@all-contributors
please add @ivangayton for projectManagement code review ideas.
please add @robsavoye for maintenance mentoring code review ideas.
please add @krtonga for code docs tool ideas.
please add @biomassives for code.
please add @Mudi-business for code.
please add @petya-kangalova for docs eventOrganizing ideas.
please add @spwoodcock for code review infra ideas.

Sorry if I missed anyone!

[...]

PR #132, opened 2023-03-07
Add allcontributors bot

Ndacyayisenga-droid:
fixes https://github.com/hotosm/fmtm/issues/83

cc @spwoodcock @petya-kangalova @robsavoye


spwoodcock:
Thanks @Ndacyayisenga-droid.
I added the all-contributors app to the fmtm repo.
I updated the README to try and make it work, but couldn't on a cursory look.
I think we may need to re-open the issue and have another attempt.

PR #134, opened 2023-03-07
Update Contribution.md

AugustHottie:
I corrected a typo from line 4...."We welcome and encourage contributors of all skill levels and we are committed to making sure your participation is nclusive" I changed ''nclusive to inclusive". I believe that was what you meant to write.


spwoodcock:
Thanks @AugustHottie !
Merged, as single typo in docs.

PR #141, opened 2023-03-07
fix typo in the README.md

Ndacyayisenga-droid:
cc @spwoodcock


spwoodcock:
@Ndacyayisenga-droid thanks for the contributions! It would be great if you could merge these typos into a single commit and PR in future, to save reviewing each one 👍

PR #142, opened 2023-03-08
Add endpoint to import test data zips on first load

### LINKED ISSUE #138: Create endpoint to preload test zips (for debug/testing)
- When fmtm is first loaded there is no data available.
- It would be useful to pre-populate data for testing / debugging.
- Create an endpoint that loads zips from: https://github.com/hotosm/fmtm/tree/main/tests/project_files/test_zips and calls the upload methods manually.

Something like `API/load-test-data` would work, and could be called from the /docs page.

spwoodcock:
Fixes #138

- I made the debug endpoints conditional. They are omitted in production.
- I also moved the read_xlsforms method call into the API server startup.
- Created a new endpoint to automate the data import on first API load.
- Useful for testing when tearing down / starting up docker config.
- The 5 data zips commited to this repo and imported as tasks, with QR codes.
- To do this I moved the test data zips under /src/backend/test_data


spwoodcock:
Also added to the docs. PR ready.

PR #160, opened 2023-03-08
Add an empty line after every file with code

Ndacyayisenga-droid:
**Add an empty line after every file with code**

There are several reasons why this convention is followed:

1. Some programming languages and tools require the last line of a file to end with a newline character. Not having a newline character at the end of a file can cause issues when working with these tools.

2. The newline character is used as a delimiter to separate lines of code in a file. If the last line of a file doesn't end with a newline character, it can sometimes cause issues when concatenating files or processing them in other ways.

3. It can also make the file more readable, especially when working with text editors or command-line tools that display files. Having an empty line at the end of a file can make it easier to visually distinguish the end of the file from the last line of code.

cc @spwoodcock @petya-kangalova @robsavoye


robsavoye:
You mean making sure all files end with a newline, and having an empty last line, not a blank one ?

Ndacyayisenga-droid:
> You mean making sure all files end with a newline, and having an empty last line, not a blank one ?

@robsavoye ya sure

robsavoye:
Cool, cause otherwise git diff complains. :-)

PR #174, opened 2023-03-10
Branch docchanges

Linda-Njau:
@petya-kangalova  I've updated the dataflow.png to a more concise and user friendly flowchart. The content is the same, but the format is different. Should I resize the image to be bigger?. What do you think of the color addition?Lemme know if I should make any changes to it.


robsavoye:
The dataflow.png is exported from the [dia](https://wiki.gnome.org/Apps/Dia) tool, and the chart may change occasionally. The actual flowchart symbols are part of a standard, and the different shapes have different definitions. I do like the look of your modified version, the colors too, but would become out of` date if there are changes to the dataflow.dia file. It is entirely possible to have two versions of this file, but the modified one wouldn't get updated when the dataflow.png does. So maybe the prettier one we should make more high level, and then it wouldn't need to be updated if the original changes, probably by adding more detail.

Linda-Njau:
Thank you for the clarification. I’ve learnt something new! I’ll look into making it more high level with extra details.

PR #175, opened 2023-03-12
Frontend/more details

Mudi-business:
updated task status and resolved UI breaking


spwoodcock (changes requested):
Please don't modify the docker setup, it's required to automatically build and deploy for production

spwoodcock:
`.env.dev` in the subdirectories isn't required (if you need it for something else, please add to .gitignore). Variables injected by docker compose should be read from `.env` in the repo root.

Also the Dockerfile under each directory requires hardcoding the URL to the production values, due to how our building is configured with quay.io (build args cannot be passed).

Otherwise looks great to me!

spwoodcock (approved):

PR #182, opened 2023-03-12
Frontend/shared modules

Mudi-business:
removed unnecessary packages in fmtm_openlayers project  and created shared modules to avoid modules duplication in main


spwoodcock (commented):
I think you added the .env files back in, plus we should only need a single gitignore in the repo root.

Otherwise, there is a lot of code edited, so it's hard to review it all. The map stuff looks good though & great idea stripping the deps for the map frontend.

spwoodcock:
Also, the OSM_LOGIN_REDIRECT_URI variable in the .env.example was changed to OSM_LOGIN_REDIRECT_URL. It should definitely be URI, not URL. Please update.

spwoodcock:
And for organisation / visibility, would it be possible to put shared modules under src/frontend/shared?

Mudi-business:
These were the first commits and you'll probably see those in all three
PR's but I have deleted those files you should look on latest commits
you'll see a commit that *say removed unwanted files *in both PR's

Mudi-business:
> And for organisation / visibility, would it be possible to put shared modules under src/frontend/shared?



> And for organisation / visibility, would it be possible to put shared modules under src/frontend/shared?

shared is in `src/frontend/main/shared` it's in main project not in` src/frontend/shared`

Mudi-business:
> Also, the OSM_LOGIN_REDIRECT_URI variable in the .env.example was changed to OSM_LOGIN_REDIRECT_URL. It should definitely be URI, not URL. Please update.

I'm not sure about that I don't remember dealing with that file rather than just look at the configurations structure

Mudi-business:
> > And for organisation / visibility, would it be possible to put shared modules under src/frontend/shared?
> 
> > And for organisation / visibility, would it be possible to put shared modules under src/frontend/shared?
> 
> shared is in `src/frontend/main/shared` it's in main project not in` src/frontend/shared`

this is the correct path `src/frontend/main/src/shared`

spwoodcock:
Seems like it was this commit by @Roseford that changed the variable accidentally:  https://github.com/hotosm/fmtm/commit/380b1da4133ef15ceada845fa455bfb175ca1d81

The commit must have been merged into this PR. FYI @Roseford, URI stands for Uniform Resource Identifier and isn't a typo 😃 

(@Mudi-business could you please change it from URL to URI in this PR please?)

Everything else looks good, thanks for clarifying things!

Roseford:
Oh!!, so sorry about that.

Roseford:
Thanks for the explanation @spwoodcock, sorry about that.

spwoodcock:
No worries, thanks for contributing! 😄

Roseford:
I have reversed the changes I made @spwoodcock, please check out my pull request.

Mudi-business:
> Okay, in that case no need to update this PR @Mudi-business. We can resolve the merge conflict here.

I've already change  `OSM_LOGIN_REDIRECT_URL` to `OSM_LOGIN_REDIRECT_URI`

PR #190, opened 2023-03-14
Added more details to the docs

Roseford:


robsavoye:
Looks nice. Something I use when working with multiple branches is "git worktree", check it out! It's great when you have multiple branches you are working on in parallel. Also consider in a another PR adding something about after a PR has been approved and merged, to delete the branch both locally and remotely. Otherwise we get buried in dead branches we don't need.

PR #207, opened 2023-03-16
dev server & github actions for deployment

mahesh-naxa:


spwoodcock:
Hey @mahesh-naxa! Thanks for the input 😄 I see you made the docker-compose.prod.yml use environment variables. I hardcoded things as we can't include build args in the production build, due to restrictions on webhook triggers on quay.io (quay only accepts a dockerfile and context).

So dynamic variables will work for the prod deploy (as the deploy script includes --build with docker-compose), but the quay.io builds will fail.

I think we need to revert the changes for now from commit 9f845bd.

If you can think of a solution that allows us to trigger the deploy after the quay.io builds finish (i.e. wait for the build, then pull and deploy), then that would be the optimal solution. Any input on how to do this from @eternaltyro or @dakotabenjamin would be appreciated.

Quay.io does seem a little restrictive with not allowing build args. As an idea, would changing to GitHub Container Registry be a solution for this?

.github/workflows/dev_deploy.yml
spwoodcock, line 1: Looks good! It should be ok to uncomment the workflow_dispatch, allowing manual triggering if needed

spwoodcock (changes requested):

src/frontend/fmtm_openlayer_map/Dockerfile
spwoodcock, line 1: This is hardcoded here due to reasons in my comment. The dynamic dockerfile with variables is one directory level up: src/frontend/debug.dockerfile

dakotabenjamin:
> Perhaps we can use GitHub's [Repository Dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch) event trigger, to trigger the GitHub Action after the build is completed?

https://github.com/marketplace/actions/wait-on-check could be a solution, but I haven't looked too deeply at it

eternaltyro:
Quay allows for creating repository notifications - including a webhook POST request on build success (or failure). Maybe this is a good starting point: https://quay.io/repository/hotosm/galaxy-api/create-notification

Github can generally integrate well (and play nice with) Quay so this shouldn't be too convoluted to implement.

mahesh-naxa:
@spwoodcock i will look into it. yes can you add in my username, i will leave it in slack DM.

robsavoye:
Is this still needed with the change for the development server if it's using github actions ?

spwoodcock:
Yep, we need an action for prod and an action for dev. 

They are very similar, but have different triggers.

spwoodcock:
To get this deployed ASAP, I propose we make the webhook / deploy trigger another issue.

@mahesh-naxa could you remove the variable substitution as described above for now, so that the Quay builds will work.

Then this can be merged into development and the dev server started.

I will create a new issue about a potential POST success webhook from Quay to trigger the builds in the future. 

@eternaltyro do you know if it's possible to pass build args into the Quay build? As far as I can tell this isn't possible.

spwoodcock:
@mahesh-naxa if possible, could you please also rebase from the development branch and resolve merge conflicts 👍

mahesh-naxa:
hi @spwoodcock i have setup development server already and added in `deploy_dev` action in here https://github.com/hotosm/fmtm/pull/252. it's merged already and seems working. 
Only caveat is i have made local changes in (Dockerfile & Compose) in server (mainly to ARG & .env). development build will fail if there is conflicting changes in docker-compose or Dockerfile. 

I think we can close/discard this PR.

Have to think of a way that works for both development server and prod, we wouldn't be wanting to have separate docker stuff for each environment. (also we will soon have staging server as well)

If quay.io doesn't support build args, we might want to shift everything to github image registry and have image tagging based on branch names (main, development or even commit hash) & have that accessed in docker-compose.prod.yml through actions environment.

PR #218, opened 2023-03-18
Remove locale and instructions from the database

robsavoye:
Locale is unnecessary as it's available from the system, and since uses ODK Collect, there is no need for instructions lioke the Tasking Manager uses.


robsavoye:
This patch is a blocker for the soon to be submitted PR that implements CreateProject.

PR #230, opened 2023-03-21
Feat createproject

varun2948:
In this pull request i have worked on create project module where we can create project and upload geojson boundary although there is a sanity check remaining on the form which I will finish off later today.


Mudi-business:
Based on my review, I think everything looks good , Thank you for your wonderful work @varun2948

PR #232, opened 2023-03-21
ui update for auth

biomassives:


robsavoye:
Can you resubmit this into the development branch instead of main ?

robsavoye:
I'll close this since I hear there is about to be a new PR on the same issue, which will also be applied to the development branch.

PR #268, opened 2023-03-27
Feat xform_title added

varun2948:
I have  also done a simple fix on backend side aswell as per my need and getting reference from @nrjadkry and he approves that aswell for the form category thing


emi420:
Hi @varun2948 !

After "Create Project" and fill the form, when I click "Next" I receive the following error: 

`Uncaught TypeError: formCategory is undefined`

Do you know what could be wrong here?

varun2948:
Hi @emi420 i have added an initial state and stated the variable name now it should not be showing that issue. can you check it and let me know.

PR #281, opened 2023-03-28
Added comments to osm.py

neelimagoogly:
@robsavoye and @petya-kangalova please review it.

### DIFF
--- a/src/backend/auth/osm.py
+++ b/src/backend/auth/osm.py
@@ -1,19 +1,21 @@
+# Importing necessary libraries for defining a function and data types
 from typing import Union
-
 from fastapi import Header
 from osm_login_python.core import Auth
 from pydantic import BaseModel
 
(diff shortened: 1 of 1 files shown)


robsavoye:
Thanks for adding comments. Usually for python comments are in triple quotes in a line just under the function, class, or method. For standalone code those get picked up by pydoc. For FastAPI that won't happen, but we might as well follow standard procedure.

neelimagoogly:
> Thanks for adding comments. Usually for python comments are in triple quotes in a line just under the function, class, or method. For standalone code those get picked up by pydoc. For FastAPI that won't happen, but we might as well follow standard procedure.

Ok I'll make sure the changes are made according to this.Thank you

robsavoye:
I'm closing this as now this will expand into multiple issues as we go forward with Outreachy.

PR #322, opened 2023-04-02
Fixing minor error in README.md

mohammadareeb95:


robsavoye:
Line 73 is fine the way it was.

PR #326, opened 2023-04-04
Generate qr codes

varun2948:


robsavoye:
While this uses the Overpass support, in a different issue we'll want to renable the dbname. If that's a special value of "underpass", then it used the new HOT maintained database for the planet. That allows faster and larger data extracts.

PR #333, opened 2023-04-06
Fix odk collect qr codes

sijandh35:


robsavoye:
Interesting you you can upload it as a draft, that's broken in Central master branch.

PR #334, opened 2023-04-06
fix: added missed slash in the server url

sijandh35:


sijandh35:
I have missed slash in the server url, sorry for multiple pr for same issue.

PR #354, opened 2023-04-16
feat: basic typescript config will add more config

varun2948:


robsavoye:
Can you clean up thew conflicts ?

Mudi-business (approved):

Mudi-business:
I think it'll be nice to put up some comments of the new config that you've add up.

varun2948:
@Mudi-business there is added comments on the config I might filter out more when I adjust more config code

PR #356, opened 2023-04-17
Updated the docstrings

neelimagoogly:
As suggested by @robsavoye and @spwoodcock , I have started the work for adding the docstrings . @robsavoye , @petya-kangalova and @spwoodcock please review it.

### DIFF
--- a/scripts/odk_fieldmap_original/utils/aoi2odk.py
+++ b/scripts/odk_fieldmap_original/utils/aoi2odk.py
@@ -25,10 +25,17 @@ from geo_utils import get_extent_bbox, make_centroids, osm_json_to_geojson
 from overpass import query
 
 
-def get_buildings(aoi_file):
+def get_buildings(aoi_file: str) -> dict:
     """Given a GeoJSON AOI polygon, returns a centroid for every
     OSM building polygon within it.
+
+    Args:
+        aoi_file (str): Filepath of the AOI GeoJSON polygon.
+
+    Returns:
+        dict: JSON response from Overpass API containing building data.
     """
+
     (infilepath, extension) = os.path.splitext(aoi_file)
     extent = get_extent_bbox(aoi_file, extension)
     querystring = (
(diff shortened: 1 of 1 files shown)


neelimagoogly:
> Looks like a great start to me, thanks for the contribution @neelimagoogly!
> 
> For some extra validation of the docstrings it might be worth running the pre-commit hook for **ruff**
> 
> You can do this with:
> 
> ```shell
> pip install pre-commit
> pre-commit run --files src/backend/path/to/file.py
> ```
> 
> This will fix any formatting errors it can and output any remaining to the terminal.

Sure , I will follow this.

neelimagoogly:
> That's slightly funny, since this is Ivan's scripts which he uses himself, and aren't integrated into FMTM at all. This has been replaced by osm-fieldwork, but Ivan still uses the scripts the manual way as people have been using the FMTM concept, Much of the goal these last few months has been to replace the time consuming manual steps with a fully automated one process, which is now working in FMTM for the pre-mapping phase as of last week.
> 
> He'll probably appreciate comments like this, and it's a good place to make changes like this in a safe manner. I will ask Ivan to review this PR, as it's his code. I did notice a few change to the default datatype from None to dict() or something else. That could possibly break things. A lot of code will check for None, but now it may need to check for len(foo) > 0 instead. I think working on good code documentation is important, so thanks. I approve of tweaking the code parameters to have type info, instead of just **None**, but you'll need to make sure you can test changes. Having just worked through this exact issue myself yesterday when I made a similar change.
> 
> We'll need to decide on whether to use sphinx of mkdocs for class & method API doc strings pretty soon.

Thanks for the clarification and yes I understood what you said.

spwoodcock:
> That's slightly funny, since this is Ivan's scripts which he uses himself, and aren't integrated into FMTM at all. This has been replaced by osm-fieldwork, but Ivan still uses the scripts the manual way as people have been using the FMTM concept, Much of the goal these last few months has been to replace the time consuming manual steps with a fully automated one process, which is now working in FMTM for the pre-mapping phase as of last week.
> 
> He'll probably appreciate comments like this, and it's a good place to make changes like this in a safe manner. I will ask Ivan to review this PR, as it's his code. I did notice a few change to the default datatype from None to dict() or something else. That could possibly break things. A lot of code will check for None, but now it may need to check for len(foo) > 0 instead. I think working on good code documentation is important, so thanks. I approve of tweaking the code parameters to have type info, instead of just **None**, but you'll need to make sure you can test changes. Having just worked through this exact issue myself yesterday when I made a similar change.
> 
> We'll need to decide on whether to use sphinx of mkdocs for class & method API doc strings pretty soon.

I didn't check which files on my brief review 😅 But still, it's a good example of docstrings and type hints to continue from.

Agreed we don't want to be changing any logic here, just docstrings and type hints.

One other decision to make is if we would like to use Google, Numpy, or reStructuredText docstring formatting. It's purely a style choice, as other tools should work with any. My preference is Google (what @neelimagoogly used 👍), I think it's the most simple and clear.

Issue #165, opened 2023-03-08
Create FMTM Tutorial


robsavoye:
Create a tutorial on how to use FMTM for a mapping campaign.

Roseford:
Hi @robsavoye @petya-kangalova,
I would love to work on this, is there a particular format you would like the tutorial to be in?

Ndacyayisenga-droid:
@robsavoye @petya-kangalova where should I include this tutorial ?

Roseford:
@robsavoye @petya-kangalova 
I have written a tutorial on how to use the fmtm tool to create a mapping campaign, the link to the file is attached below
https://docs.google.com/document/d/1KMtCgIXCEulpDs19hkXe4bY-16Y9idjRghEc8Y981s4/edit?usp=sharing

Roseford:
Hi @robsavoye,
I have created an FMTM tutorial here https://github.com/Roseford/fmtm/blob/fixing_issues/docs/Mapping-Campaign-Using-FMTM.md
Please check it out.

Roseford:
Hi @spwoodcock,
I have created the FMTM tutorial at https://github.com/hotosm/fmtm/blob/main/docs/Mapping-Campaign-Using-FMTM.md do you think it would be a good idea to close this issue.

spwoodcock:
Looks good @Roseford! It just needs to be added to the _Sidebar.md to be visible - I'll update it when I'm back at my computer and close this.

suhana832:
can you assign this task to me ?

suhana832:
how many days I have ? to complete this task

ivangayton:
There's no hard time limit! But the first thing would be to reach out to @Roseford who's already been working on this, see what's been done so far, and ask where help is most needed.

ivangayton:
In my view, we could almost certainly use several tutorials, one very comprehensive video style for beginners, one text plus screenshots for more advanced people, and perhaps a slide deck for people trying to deliver the tutorial in person or via video chat.

suhana832:
should I start making it  y my own or collaborate with her ?

ivangayton:
I think you should ask what she prefers! But it also depends what you prefer to do.  Both would be okay; having two different tutorials is not a bad thing, and having one built by collaboration is also great. But it's probably best to first make sure that there isn't too much duplication of effort; it would be a shame for both of you to do the same work separately

suhana832:
I will make sure all of the conditions you mentioned. Thank you ☺️

[...]

PR #378, opened 2023-05-02
modified the mapping campaign file.

Roseford:
#165 @robsavoye,
I initially created a mapping campaign file and it was merged into main. With @petya-kangalova help, I found out it was not quite correct so I edited it but this time it was merged into development. Due to what I'm guessing is a merge conflict, the resulting document in both development and main are not quite correct. Here is a PR to the development branch with the corrected version. With your permission I would also like raise a PR to the main branch so that this particular conflict doesn't arise again. 
Please take a look at it and let me know what you think.


Roseford:
@robsavoye, I do not know if you noticed but I finally found a way to hard wrap the columns. I made use of an extension since the editor I'm using does not support hard wrapping yet. I also noticed that headings are not edited so the second sub topic in the file is quite lengthy, sorry about that. Apart from this setback, I believe the problem of hard wrapping for readability has been solved. What do you think?

robsavoye:
I think we just need shorter sub headers :-) The hard work wrap looks fine thanks. Much easier to read the  patches in github too. I can approve this for development, and can always apply it manually to main.

Roseford:
> I think we just need shorter sub headers :-) The hard work wrap looks fine thanks. Much easier to read the patches in github too. I can approve this for development, and can always apply it manually to main.

okay, Let me make the sub heading shorter.

PR #385, opened 2023-05-04
Select nodes or polygons for data extract

### LINKED ISSUE #374: Select nodes or polygons for data extract
Right now the default for all data extracts is to generate a geojson file with only nodes, ie... building centroids. The osm-fieldwork backend now supports generating either polygons or nodes. On the create project page, when specifying the category for the data extract should have an option to specify the output geometry.

nrjadkry:
This pull request addresses the issue of default data extract generation in the geojson format containing only nodes or building centroids. With this update, users can select either nodes or polygons as the output geometry for data extraction. The fmtm backend now supports both options, allowing users to extract the data in the format that best suits their needs.

To make use of this feature, users can now select the output geometry option when specifying the category for the data extract on the create project page. This update provides users with greater flexibility and control over the data extraction process.

- Closes #374

### DIFF
--- a/src/backend/projects/project_crud.py
+++ b/src/backend/projects/project_crud.py
@@ -678,6 +678,7 @@ def get_odk_id_for_project(
 def generate_appuser_files(
     db: Session,
     project_id: int,
+    extractPolygon: bool,
     upload: UploadFile,
     background_task_id: uuid.UUID,
 ):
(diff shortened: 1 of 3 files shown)


robsavoye:
I have a bug fix for the ODK_CENTRAL_SECURE problem in osm-fieldwork I'll push a new release to pypi soon.

PR #416, opened 2023-05-18
Update/generate osm data

nrjadkry:


robsavoye:
You'll want to use the new JSON support for submissions instead of CSVDump.

PR #533, opened 2023-06-26
Feat download submission json frontend

varun2948:


varun2948:
This PR is dependent with #532  so can you review it first then merge this @robsavoye

PR #569, opened 2023-07-05
Implement hackathon algorithm

nrjadkry:


robsavoye:
This is a huge patch, so will take a little while to properly review. I do notice when diffing that there are multiple embedded tabs for blank lines and at the end of some lines. Please clean those up while I work on reviewing this.

robsavoye:
It looks like you need to sync your fork with main. There's a few changes in main not in your patch.

robsavoye:
I'd prefer to see the large SQL queries be in an external *.sql file, instead of hardcoded into the source code.

nrjadkry:
@robsavoye I have moved the sql query to the seperate .sql file

robsavoye:
This is a huge patch. It looks ok, but to really know I'll approve it for the development branch, which is why we have one in case anything breaks.

Issue #567, opened 2023-07-05
Create Project needs a hashtag field


robsavoye:
After converting the ODK json file to OSM XML and conflating it, and pushing it to JOSM, it needs a hashtag. Currently the create project page doesn't have a place to enter the project hashtag. The hashtag is needed to generate statistics on FMTM data contributions. A hashtag column will need to be added to the database table.

varun2948:
@robsavoye Hashtag completed

PR #604, opened 2023-07-17
Feature/hashtag field project has hashtags field and projects can be filtered based on these values

Seckrel:
solution to #567
since hashtags fields are array they can support multiple hashtags per project


robsavoye:
Same file conflicts as the other PR, please fix and then I can approve.

Seckrel:
> Same file conflicts as the other PR, please fix and then I can approve.

Conflict fixed in all PRs

PR #615, opened 2023-07-19
Basic docs about database backup / migration

spwoodcock:


spwoodcock:
I tested the commands written in the docs on the prod server & backup/restore worked fine.

This is to be merged into `main` and included in the wiki.

PR #622, opened 2023-07-20
Feat update category

nrjadkry:


nrjadkry:
@robsavoye In this Pull request, I have created an api to update the category and form of the project. Also, I have implemented multithreading in the download submission for the project. Please have a look at this, especially in the multithreading

robsavoye:
So this creates a directory full of zip files ? Ideally the JSON (and later any images) fro each zip file needs to be combined into a single output file.

PR #640, opened 2023-07-27
upgrade osm-fieldwork version to 0.3.3

nrjadkry:


robsavoye:
Don't you also have to update the pdm.lock file ? I had to do that when I was testing locally.

nrjadkry:
@robsavoye I did update pdm.lock file too.  If you look into this pull request closely, you will see two files changes.

PR #662, opened 2023-08-03
Feat conflate data

nrjadkry:


robsavoye:
Can you fix the failing tests ?

robsavoye:
Since you can query the database for the list of XForms, I added an optional argument to OdkCentral.getAllSubmissions() to take a list of tasks to download submissions for. Without this argument, OdkCentral wil query Central for the list of XForms.

nrjadkry:
the failing tests are because of the osm-fieldwork. I guess this is GDAL issue.

ModuleNotFoundError: No module named 'osgeo'
This is the error message.

We might need to install gdal in this project, also because we will use image basemaps later. We might need gdal for this too.

robsavoye:
osm-fieldwork already installs gdal. If I install the PyPi package, everything works fine.

nrjadkry:
Okay. We might need a new release of osm-fieldwork, what do you think?

robsavoye:
Working on it now, just updating the test cases before doing a 0.3.4 release later (today, but it is a weekend...) This release will have the conversion working better, the bulk downloading, etc... all of which you'll be using. The current test cases are pretty lame, and don't really test anything, so I'm going to add some actual tests for the conversion process (since I just refactored it heavily), then I'll do the release. I did test with actual data files, but the tests should also do that to catch future mistakes.

nrjadkry:
@robsavoye tests have passed now.

nrjadkry:
> in src/backend/app/submission/submission_routes.py: removing the extra "" is a hack. :-) At one point it was required to use Osmfile.footer() to close the file, that's probably still being done in the backend someplace. Then I discovered a python class can have destructors, so now it's added automatically.
> 
> in src/backend/app/submission/submission_crud.py: I see you loop through all the tasks, and download each submission using getSubmissions(). This really needs to be multi-threaded. You can either adapt the new code in OdkCentral.getAllSubmissions(), or just call that method. Otherwise it' really slow, so better to do in parallel.
> 
> Also fix the failing tests.

where should I remove the extra "". Since it was producing an extra </osm>. I have removed that.

PR #676, opened 2023-08-07
Add GDAL to backend & refine pytest workflow

spwoodcock:
On request from @nrjadkry, added the GDAL Python bindings to the backend.
- I upgraded the backend dockerfile to use Debian Bookworm (over Bullseye).
- GDAL v3.6.2, which matches Debian Bookworm. Let me know if you need a newer version & we can review.

In order to make running pytest with GDAL easier, I updated the pytest Github action to run inside the backend container.
- As a consequence, this should significantly speed up running pytest, as all dependencies are already bundled and no longer require an install.
- This also means that the postgres service runs in a bridge network and can be accessed that way, instead of port mapping via localhost.

I also had to do some minor refactoring of conftest.py to make this work...


nrjadkry:
@spwoodcock The tests have failed asking for gdal-config.

spwoodcock:
This turning into a bit more work than I expected 😆

It should speed up our pytest runs by quite a bit though.

Ready to merge.

spwoodcock:
There will likely be a merge conflict with #675, so please merge this first & I will fix the other PR after.

PR #678, opened 2023-08-07
End work on postgis_snippets in this repo

ivangayton:
Merging the postgis task splitting work into development; it's all been in a separate directory inside scripts/postgis_snippets. It's not being moved into [fmtm-splitter](https://github.com/hotosm/fmtm-splitter) so won't be developed anymore here. 

Once this is merged, the postgis_snippets branch should be deleted, though the directory shouldn't be nuked too quickly because it contains a few things that might be useful later that aren't yet in fmtm-splitter.

@robsavoye in `/scripts/postgis_snippets/task_splitting/fmtm_task_splitting_for_buildings` you'll find the task splitting script specific to building mapping (distinct from the work in progress for road mapping) modularized into 4 scripts. The comments at the beginning of each script specify the inputs, outputs, and parameters.


robsavoye:
Since FMTM doesn't uses these files directly, I can approve this since it won't break anything. We will need to see how to update fmtm-splitter for the updated SQL files. We can leave the directory til after we switch to fmtm-splitter, and the branch can hang on for a while too. I think in the future fmtm-splitter will be a better project to debug SQL queries as then you don't to run FMTM to debug and improve SQL queries.

PR #687, opened 2023-08-09
fetch submission basic info

nrjadkry:
List submission returns only basic info. We will merge this only after a new release of osm-fieldwork.


nrjadkry:
@robsavoye 
No, we don't need to add anything more for this. I have added that already.
We don't need to release a new version just for this, whenever we release a new one, then we can merge this.

nrjadkry:
We can merge this now we have released a new version of osm-fieldwork.

PR #690, opened 2023-08-09
Add JOSM API to docker compose dev stack

spwoodcock:
Part misunderstanding, part fun little project.

I originally containerised JOSM in the hope to make the API accessible.

**Two Issues**

- JOSM is useless without the GUI component.

- JOSM has very strict and non-configurable remote access settings. It's accessible on 127.0.0.1:8111, which doesn't work with Docker (uses localhost).

**Solution**

- I made the JOSM GUI accessible via noVNC. This way we don't need to install Java or JOSM on our local machine for development. It runs entirely in a container, the GUI accessible on `localhost:8112`.

- I added nginx to the JOSM container, which is against the one-service-per-container ethos, but necessary in this case. Nginx reverse-proxies 127.0.0.1:8111 to localhost:80, making the JOSM API accessible by port mapping the container 8111:8111. Now we have remote JOSM available on `localhost:8111` for development.

**Next Steps**

- Tasking manager connects remotely to the JOSM API and configures the user session.
- It doesn't use anything fancy: just http calls using URL params.
- We can do the same from the FMTM frontend (making calls from the backend if needed too using `josm:80`).
- E.g. we can call a URL like: `http://127.0.0.1:8111/load_and_zoom?left=8.19&right=8.20&top=48.605&bottom=48.590&select=node413602999` to configure JOSM remotely.

**Note:** I added docs on how to run this in DEV-2: Backend.

### DIFF
--- a/docker-compose.josm.yml
+++ b/docker-compose.josm.yml
@@ -1,0 +1,54 @@
+# Copyright (c) 2022, 2023 Humanitarian OpenStreetMap Team
+# This file is part of FMTM.
+#
+#     FMTM is free software: you can redistribute it and/or modify
+#     it under the terms of the GNU General Public License as published by
+#     the Free Software Foundation, either version 3 of the License, or
+#     (at your option) any later version.
+#
+#     FMTM is distributed in the hope that it will be useful,
+#     but WITHOUT ANY WARRANTY; without even the implied warranty of
+#     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#     GNU General Public License for more details.
+#
+#     You should have received a copy of the GNU General Public License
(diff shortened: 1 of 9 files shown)


nrjadkry (approved):
This is great @spwoodcock .  So using this we wont need JOSM installed on the local machine.

robsavoye:
Hum... so you have JOSM in a container, but it works  normally for editing the data ?

spwoodcock:
@robsavoye yes it should have all the functionality of running locally, just via a web browser instead.

Give it a test and let me know 👍

spwoodcock:
Here's it running:

image

Then after copying `http://127.0.0.1:8111/load_and_zoom?left=8.19&right=8.20&top=48.605&bottom=48.590&select=node413602999` into the browser:

image

PR #696, opened 2023-08-10
Mbtiles for Project

nrjadkry:
This pull request has the feature to download the mbtiles for a project.

Here, I have created an api which accepts a project_id and the source for downloading mbtiles to use for odk collect.

Todo:
Since downloading tiles takes a large amount of time, we need to implement this in backgroud task.


nrjadkry:
Test cases have failed here.
tests could not find mercantile.
So, I have installed mercantile and still tests are failed.

robsavoye:
No, it doesn't, so it'll need to be added to the pyproject.toml file, and I'd have to push a new release.

nrjadkry:
I have added mercantile in the pyproject.toml and it's working fine in my local machine.
May be @spwoodcock knows, how does this works in the github action.

spwoodcock:
We need to add mercantile to the dependencies in the osm-fieldwork pyproject.toml via a PR.

Then when osm-fieldwork is installed, all required dependencies would also be.

robsavoye:
I assume I need to push a new osm-fieldwork release before approving this PR so I don't break the dev server ? The fix is in git now.

spwoodcock:
Yes I think so, or the dev server will get the same error missing the dependency.

This PR should also include the bump in osm-fieldwork version 😃

nrjadkry:
@robsavoye we would also need pySmartDL in osm-fieldwork, not just mercantile.

spwoodcock:
I'll take a look at adding that @nrjadkry 

Btw I think you might need to rebase to / merge development on this branch, as it's including the josm commits from another PR 👍

nrjadkry:
I pulled origin branch and got all these files which should be already in development.

spwoodcock:
I just released osm-fieldwork 0.3.5 with the required dependencies.

robsavoye:
The tests still fail looking for mercantile

spwoodcock:
This one is my fault actually.

I recently updated the pytest workflow to use an fmtm image, but I haven't made a workflow that builds the fmtm:ci image.
I'll push the image and re-run the actions.

robsavoye:
Thanks, cause I'd like to get this PR merged in, and possibly then update the prod server.

spwoodcock:
Pushed an new image & restarted the jobs.

This shouldn't be an issue unless dependencies change again.
(debating if it's worth making a new workflow for building the ci image, or if we just manually build and push if dependencies change)

PR #735, opened 2023-08-18
Remove manual signup/login (use OSM)

spwoodcock:
## Issues

**1**
We want to use OSM login for everything.
Manual login is a legacy feature we used before getting OSM login working.

**2**
Currently we default to OAUTHLIB_INSECURE_TRANSPORT=1 in the backend, which is insecure and could leak user credentials.

**3**
OSM login does not work during development, only production.

## Solution

- Removed login and signup code from frontend.
- Removed API endpoints for login and signup.
- Default to OSM login only during dev/prod.
- Update docs with OSM redirect url info for frontend.
- Remove default OAUTHLIB_INSECURE_TRANSPORT=True in backend & only set this if DEBUG=True.

Within this PR I also updated the OSM OAuth login flow slightly via the frontend.
It worked in production, but produced CORS errors during development.
I updated to use window.postMessage for secure communication between the popup and main window.
Now we can use OSM login during development / testing too.

## Additional Context

With the current setup, if a user attempt OSM login, but there is a manual user with the same name, then we get an error.

@nrjadkry is working on a migration for manual users, so we can keep them in dev/prod for the time being.

The login for dev/prod already defaults to OSM, so if any user finds themself locked out of manual login, they can contact us & we will migrate their entries to an OSM account user.
(this was the main concern of Ivan - not locking out any existing users)


robsavoye:
I know we have talked about making everyone have an OSM account. Definitively for the project managers, but for the mappers just trying to get their QR Code, this might be a problem. I know that in past mapathons getting people through getting their OSM account was time consuming, and considered to be a possible problem.

ivangayton:
We did make the decision that we should ask everyone to have an OSM account. I'm reluctant to reverse that. One of the key values of the FMTM is to ensure that field contributions to OSM are attributed to the field mappers, rather than to the validators or campaign managers.

I'm open to persuasion otherwise, but for the moment I still think we should ask people to have an OSM account to log into FMTM.

PR #737, opened 2023-08-18
Project creation: handle odkcentral failure, additional logs

spwoodcock:
Small unhandled exception when the ODKCentral address cannot be contacted.

Added additional logging to the project creation stages (3 endpoints) for debugging.
(I managed to complete the full project and task creation workflow successfully 🎉)


spwoodcock:
As this PR demonstrates, it's pretty tricky to view the diff of even tiny code edits when we don't have consistent formatting.

We should discuss the enforcement of pre-commit hooks, ideally using something like: https://pre-commit.ci

PR #753, opened 2023-08-23
Hotfixes

nrjadkry:
I was having permission issues for tiles directory. I have made few changes for that. Is this okay?? @spwoodcock 

We might need to update accordingly in the devlopment server too.


spwoodcock:
Looks good to me, thanks @nrjadkry

PR #762, opened 2023-08-24
Multithreading

### LINKED ISSUE #613: Implement multi-threading for data file creation
Right now, creating the data extracts and modified XForms for Central is sequential, so slow. Since each file to create has no dependencies on the others, we should refactor this to use a thread-pool & futures to create multiple files in parallel. There is a good working example in osm-fieldwork you can use as an example.
https://github.com/hotosm/osm-fieldwork/blob/main/osm_fieldwork/odk_merge.py

nrjadkry:
Closes #613


spwoodcock:
@all-contributors please add @nrjadkry for code ideas maintenance

PR #763, opened 2023-08-24
Update logging to be more feature rich, persist logs, JSON logs

spwoodcock:
## Issue

- Currently we use logging.basicConfig, which is fine, but not very configurable.
- Logs also don't currently persist after container restarts, which doesn't help debugging if multiple PRs and merged in succession (redeploy).

## Solution

I updated to use the loguru library, which is very feature rich and easy to use.

The setup:
- Log to terminal in a standardised way, hooking all downstream loggers.
- Add colour to logs if your terminal supports it.
- All log calls are non-blocking, using multithreading.
- Backtracing for more detailed logs on traceback.
- Generic error catching, so if an unhandled exception occurs, the server doesn't go down.
- Logging to rotating files in JSON format, located in a docker volume (these logs could be synced to S3 in the future and easily ingested / analysed on a dashboard).

## Notes

I made this PR after having to debug something - hopefully this makes it easier.


robsavoye:
How will this effect logging for osm,-fieldwork ? There's also a file conflict.

spwoodcock:
Closed in favour of  #768.

To answer your question, this should hook all downstream logging, regardless of how it's instantiated.

We could keep using standard lib logging for osm-fieldwork, or switch to loguru - either will be picked up by FMTM.

PR #785, opened 2023-08-29
update logging for generate files in a seperate file

nrjadkry:
We need seperate logging for a project creation process as we are showing the process in the terminal like box when creating a project.
As you have added loguru for logging. I made changes accordingly.

### DIFF
--- a/src/backend/app/projects/project_crud.py
+++ b/src/backend/app/projects/project_crud.py
@@ -1278,6 +1278,7 @@ def generate_appuser_files(
         - background_task_id: the task_id of the background task running this function.
     """
     try:
+        log_file = log.add(f"/tmp/{project_id}_generate.log", format="{message}")
         log.info(f"Starting generate_appuser_files for project {project_id}")
 
         # Get the project table contents.
@@ -1411,6 +1412,9 @@ def generate_appuser_files(
         update_background_task_status_in_database(
             db, background_task_id, 4
         )  # 4 is COMPLETED
+
+        # Stop logging in file for this project.
+        log.remove(log_file)
 
     except Exception as e:
         log.warning(str(e))


spwoodcock:
I'm not sure I understand, but sure we can have a separate log file.

You display the file content in a separate terminal somewhere? 

Anyway, it can be done. But instead of instantiating the log during every call of the endpoint, the log should be instantiated once at FastAPI startup then hooked / filtered in the endpoint. 

Here is a good example:
https://loguru.readthedocs.io/en/stable/resources/recipes.html#creating-independent-loggers-with-separate-set-of-handlers

So add the extra log in main.py to a file you choose, with enqueue=True to make this non-blocking for async and a filter on a specific task type =project_create

Then in the endpoint bind task=project_create to any logs that you wish to also add to this extra file.

## Note

Also be aware you will lose this log file on container restart. 

Not a problem if you scrape it regularly, but if not, you may wish to make the file in a persistent docker volume (the same as the existing log volume, just a different file name).

src/backend/app/main.py
```diff
@@ -134,6 +134,9 @@ def get_logger():
             # format=log_json_format, # JSON format func
         )
 
+        log.add("/opt/logs/create_project.log",
+                enqueue=True
```
spwoodcock, line 138: This needs to filter by the bound task, otherwise you will get all logs:

```python
log.add("/opt/logs/create_project.log",
    enqueue=True,
    filter=lambda record: record["extra"]["task"] == "create_project"
)
```

src/backend/app/projects/project_crud.py
```diff
@@ -1278,7 +1278,8 @@ def generate_appuser_files(
         - background_task_id: the task_id of the background task running this function.
     """
     try:
-        log.info(f"Starting generate_appuser_files for project {project_id}")
+        create_project_log = log.bind(task="create_project")
+        create_project_log.info(f"Starting generate_appuser_files for project {project_id}")
```
spwoodcock, line 1282: This is the only log currently. So you will just get a series of logs like:
```
Starting generate_appuser_files for project 1
Starting generate_appuser_files for project 2
etc
```

If that is what you are after, then all good.
Otherwise you may want to add a few more `project_log.info("stuff")`.

spwoodcock: Now I understand more what you are trying to do. You could probably update this to:

```python
project_log = log.bind(task="create_project", project_id=project_id)
```

Bind the project id, then you can filter the logs based on the project_id param later.

src/backend/app/main.py
spwoodcock: Also if these logs are made to be read in by the code, and not read by people directly, I would recommend JSON logging.

```python
log.add("/opt/logs/create_project.log",
    enqueue=True,
    serialize=True,
    filter=lambda record: record["extra"]["task"] == "create_project"
)
```

src/backend/app/projects/project_routes.py
```diff
@@ -694,7 +694,7 @@ async def generate_log(
             project_id, db
         )
 
-        with open(f"/tmp/{project_id}_generate.log", "r") as f:
+        with open("/opt/logs/create_project.log", "r") as f:
```
spwoodcock, line 697: Once these logs are in JSON format `serialize=True`, you can parse them as JSON.

```python
with open("/opt/logs/create_project.log", "r") as log_file:
    logs = [json.loads(line) for line in log_file]
    filtered_logs = [log for log in logs if log.get("extra", {}).get("project_id") == project_id]
    last_100_logs = filtered_logs[-50:]
```

Give this a go and if it doesn't work properly we could look into it together 👍

spwoodcock (changes requested):

nrjadkry:
@spwoodcock

spwoodcock:
Thanks @nrjadkry!

Appreciate your patience and time spent on this! 🙏 

I do have a few tiny comments, but in the interest of saving time I'll merge this and make a new PR with some minor modifications 👍

PR #792, opened 2023-09-01
Update Wiki --> GH Pages

spwoodcock:
## Includes
- Upgrade from wiki --> gh pages.
- Auto generation of openapi.json from FastAPI.
- Change of github workflow to publish pages, doxygen, swagger api docs.

Outside of the `docs` dir the main changes of note are:
- docs.yml workflow for publishing the docs.
- gen_openapi_json.py for generating the OpenAPI JSON using FastAPI.

## Notes

- Bundling the Swagger UI dist does add 10MB to the repo... (hence the huge +- on the diff).
  - This shouldn't change often / rarely ever.
  - Even if I only added to the gh-pages branch, it would still be part of the repo when cloning (unless depth=1).
- Some links to Wiki pages will inevitably be broken - they need to be updated.


robsavoye:
I see other PRs need this one first, so I'll prove it, but we do want to be able to update both the wiki and gh-pages.

spwoodcock:
Yes it does mean that - I can add the wiki workflow back in if needed

robsavoye:
Yes, we should update the wiki too. The gh-pages will have more docs, but the wiki is still useful.

PR #794, opened 2023-09-01
Project specific logging to additional separate file

spwoodcock:
Follow on from #785

Minor tweak to:
- Set log file rotation.
- Instantiate the log with `project_id=project_id and task="create_project"`
(we don't need to pass the logger between functions, as it uses a bind to differentiate it from the standard logs)

This will generate logs to terminal, `fmtm.log` (all logs), and then create_project specific logs to `create_project.json`.

### DIFF
--- a/src/backend/app/main.py
+++ b/src/backend/app/main.py
@@ -134,11 +134,15 @@ def get_logger():
             # format=log_json_format, # JSON format func
         )
 
-        log.add("/opt/logs/create_project.json", 
-                enqueue=True,
-                serialize=True,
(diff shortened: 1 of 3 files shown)


nrjadkry:
@spwoodcock 
This pull request is on main branch. 
Are we merging into production?

spwoodcock:
Thanks @nrjadkry, changed to `development`.

Could you please test this works for your needs?
If so, I don't see why we can't merge 👍

nrjadkry (approved):

nrjadkry:
Thanks @spwoodcock .
This is nice. We can merge it.

PR #803, opened 2023-09-05
Feat update schemas

nrjadkry:


src/backend/app/projects/project_schemas.py
```diff
@@ -50,26 +50,33 @@ class BETAProjectUpload(BaseModel):
...
-    description: str = None
-    num_contributors: int = None
-    total_tasks: int = None
-    tasks_mapped: int = None
-    tasks_validated: int = None
-    tasks_bad: int = None
-    hashtags: List[str] = None
-    organisation_id: int = None
-    organisation_logo: str = None
+    title: Optional[str]
```
spwoodcock, line 69: I haven't gone through all of the docs for Pydantic V2, but I think the syntax might be:

`title: Optional[str] = None` as Optional params still need a default value.

src/backend/app/organization/organization_crud.py
```diff
@@ -44,15 +44,12 @@ def generate_slug(text: str) -> str:
...
-
-    # Construct the SQL query with the case-insensitive search
-    query = f"SELECT * FROM organisations WHERE LOWER(name) LIKE LOWER('%{name}%') LIMIT 1"
-
-    # Execute the query and retrieve the result
-    result = db.execute(query)
-
-    # Fetch the first row of the result
-    db_organisation = result.fetchone()
+    # Use SQLAlchemy's query-building capabilities
```
spwoodcock, line 47: Try to avoid using SQLAlchemy, as we will remove it in the future.
The raw SQL equivalent is probably best 👍

src/backend/app/tasks/tasks_crud.py
```diff
@@ -50,10 +50,18 @@ async def get_task_count_in_project(db: Session, project_id: int):
 
 def get_task_lists(db: Session, project_id: int):
     """Get a list of tasks for a project."""
-    query = f"""select id from tasks where project_id = {project_id}"""
-    result = db.execute(query)
-    tasks = [task[0] for task in result.fetchall()]
-    return tasks
+    query = text("""
```
spwoodcock, line 53: Do we need the SQL wrapped in `text`? That locks it in a bit more to SQLAlchemy

spwoodcock (commented):
Looks great - good stuff 👍 
I can't test it right now, but I left a few comments

src/backend/app/tasks/tasks_crud.py
nrjadkry: We have used sqlalchemy db session. it was not required in the previous stage. But, it is required since your latest update.  It might be because of pydantic??

spwoodcock: No it's not related to pydantic, but there is no harm having the `text()` there for now if it works.
We can refactor when we remove SQLAlchemy.

PR #831, opened 2023-09-14
Feat Remove Microfrontend

varun2948:
I have removed fmtm_map micro frontend from docker and github actions please verify it @spwoodcock so that it wont break production and development and i have tested it in my local it worked fine.

Btw this PR have also added a feature onto it but you dont have to look all the changes you just have to look onto CONFIGS,DOCKER FILES, GITHUB ACTIONS FILE and ENV.


spwoodcock:
Awesome, so the microfrontend has been refactored to a single frontend now?

spwoodcock (approved):
Great refactor - looks good to me - thanks for checking 👍

varun2948:
> Awesome, so the microfrontend has been refactored to a single frontend now?

yeah just took some time to fix it cause it was making so much problem for any other contributors to work on frontend now it will be quite easy @spwoodcock

robsavoye:
Great idea to make the code similar and easier to maintain !

varun2948:
Just found a build issue on fmtm_main will look into it @spwoodcock

spwoodcock:
The deploy seemed to go ok after I just tweaked dockerfile.prod.

Is the error preventing a build for you?

PR #840, opened 2023-09-21
Feat New Create Project and Docker Build Changes

varun2948:


varun2948:
@spwoodcock i had some fix on the docker side can you check it and probably merge it whenever u will be available.

spwoodcock:
Looks like the `pre-commit` hook triggered fine.

It's hard to see your diff on this PR, but it should be good now it's run once!

Issue #856, opened 2023-09-27
Implement ODK workflow for accessing two geojson files in a form for Janakpur Pilot Project


susmina94:

robsavoye:
Two data extracts ? Then you'd need two XForms, one for each extract.

robsavoye:
In that case we'd need to add a YAML file to osm-fieldwork to extract both highways and buildings, and the merge the highway and buildings XLSForms together. Personally, I think it's work better to have two XForms and data extracts than overload one. Why does this have to be in the same XForm ?

nrjadkry:
We wont need to merge those forms. We are uploading a custom form for this Janakpur project as well as the data extracts.

But, for the future, we might need to explore merging two forms too.

nrjadkry:
We wont need to merge those forms. We are uploading a custom form for this Janakpur project as well as the data extracts.

But, for the future, we might need to explore merging two forms too.

robsavoye:
I think merging the 2 XForms and data extracts together is a bad idea. I assume this will be supported by uploading a custom XForm and data extract ? I'm curious about the mapper data flow to map both of these in the same XForm.

nrjadkry:
Acutally, mapper wants to map a buildings. But to know which exact buildings, selecting a road is also provided here. So, two data extracts have two be uploaded.
@susmina94 knows more why such a form is designed.

robsavoye:
I'd just use a basemap of the highways, and then focus on mapping buildings, rather than making a larger and messier data extract.

robsavoye:
The basemapper program can make an imagery from Google Hybrid, which has highway names.

spwoodcock:
I agree we should avoid overcomplicating things.

Would downloading imagery with highway tags solve the issue here @susmina94 ?

I'm not sure I entirely understand the requirements.
Does the user want to map both buildings and highways at the same time?
Or is highway information only required to improve the accuracy of building tags?

spwoodcock:
Closing.
Will eventually be replaced by https://github.com/hotosm/fmtm/issues/509

susmina94:
@spwoodcock  @robsavoye 
We actually want the users to select a road segment while mapping a building so that we can have the road segment id in the same set of submissions.

The road data extract is not just for the visual inspection for the users but they would have to instead either select the road so that we can map the road id or manually enter the road id by viewing the basemap with road ids.

I still have to see what Rob actually meant with the feature collection dataset approach.
[...]

PR #858, opened 2023-09-28
Upload multiple geojson for janakpur

nrjadkry:
We needed to upload two different geojson files in th odk for the Janakpur. 
To adjust this, I have created a seperate api through which we can upload two seperate data extracts files for buildings and roads. We did not want to make changes  in the workflowin the frontend. So, we will be doing this via api.

### DIFF
--- a/src/backend/app/central/central_crud.py
+++ b/src/backend/app/central/central_crud.py
@@ -636,20 +636,18 @@ def create_odk_xform_for_janakpur(
 
     if result != 200 and result != 409:
         return result
-    data = f"/tmp/{title}.geojson"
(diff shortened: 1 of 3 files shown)


spwoodcock:
Thanks @nrjadkry 👍

I think we should resolve the discussion [here](https://github.com/hotosm/fmtm/issues/856).

But as this doesn't require any changes to osm-fieldwork and only adds new endpoints (plus it's a lot of work done, thank you!), I will merge for the time being.

We can re-assess how we want to handle this after Janakpur. Does that sound ok @robsavoye?

PR #860, opened 2023-09-28
Upload multiple geojson for janakpur

nrjadkry:


spwoodcock:
For easier refactoring in the future, any client specific 'quick and dirty' edits would be best placed under a separate router.

IMO we should have `janakpur_crud.py` and `janakpur_routes.py` to handle this, as we will likely remove the specific endpoints in the future (and integrate a more generic way).

spwoodcock:
I'll merge this now as you probably need the changes for tomorrow 👍

PR #861, opened 2023-09-28
Feat-project-details-mapper-ui

varun2948:


spwoodcock:
Do you need this for tomorrow? 

I will merge immediate if so.

Otherwise I will check the code in a bit more detail 👍

varun2948:
You can check it in details

spwoodcock:
- I had to create a project to review this & in the process I tweaked both Alembic & the project creation workflow.
- Tomorrow I will be reviewing the new mapper flow, then this can be merged 👍

varun2948:
Sounds great i hope the alembic and other stuff doesnt break create project api's

spwoodcock:
I like this a lot better - more intuitive - it looks great!

Only three things to note:
- The descriptions don't appear from the drop down menu:
image

- The legend is present in two places:
image

- The map button icons are not all centred (this will be addressed in the map component overhaul though).
image

These are minor issues, so I will merge & we can continue refining in other PRs 🎉

varun2948:
@spwoodcock yes Sam, we had a look onto that and have address all the issue u have mentioned and will create a PR or should come on another PR that we have issued

PR #873, opened 2023-10-03
fix: alembic revision issue

nrjadkry:


nrjadkry:
@spwoodcock 
do not merge it yet.
I am making few more changes. And then we can merge it.

spwoodcock:
They don't need to be migrated - they are handled entirely by Postgis between upgrades (i.e. are just baked into the Postgis images we use for the DB).

spwoodcock:
You already added the spatial_refs table to excluded from alembic. It may be good to add tiger and topology too, but it probably shouldn't matter 👍

nrjadkry:
@spwoodcock I ma facing this exact same issue with alembic migration.
https://github.com/sqlalchemy/alembic/discussions/1282

SQLAlchemy==1.4.49 is ok, but not SQLAlchemy==2.0.19

spwoodcock:
Can you use the solution provided by the maintainer with include_schema?

https://github.com/sqlalchemy/alembic/discussions/1282#discussioncomment-6529470

nrjadkry:
How much does this affect us if we downgrade sqlalchemy.
It should not matter much and we will be removing it completely later.

spwoodcock:
Hard to tell, but we would a minimum lose out on some performance (and perhaps have to downgrade geoalchrmy too).

I would be more in favour of just removing alembic (sorry you did the work, twice 😬).

Migrations aren't as important as a well operating connection to the DB.

If this issue isn't easily solvable, then I suggest we have a raw SQL schema file, with sequential SQL migration files in a directory.

nrjadkry:
Let's use raw sql just for migration instead.
It is more complicated than it seems in alembic.

PR #881, opened 2023-10-06
Filter data extract with osm-rawdata (allow custom form upload)

### LINKED ISSUE #509: Custom data extracts alongside custom XLSForm
osm-fieldwork uses osm-rawdata to generate data extracts.
The `data_models/*.yaml` files define the SQL query that will be used.

Right now a custom XLSForm needs to be mapped to one of the existing data models (categories).

This works fine for a standard XLSForm, however, if we upload a custom XLSForm we need to also be able to specify a custom SQL query to get the data extract we want.

To facilitate this we need to be able to upload a `.yaml` file defining the SQL query alongside out custom XLSForm.

To do this we need:
- Backend: to pass the yaml config file to PostgresClient in osm-rawdata.
- Frontend: UI component to generate the yaml for a project (better UX).

The generated yaml is stored in the `project.form_config_file` table, then passed to osm-rawdata when generating the extract.

Sujanadh:
This pull request extracts the data from osm using osm-rawdata package and the output are passed into filter_data in osm_fieldwork in order to filter the data according to the xlsform. 

I have another pull request with the same functionality in #877 which was from my fork. I am closing that pull request as all the changes in that pull request are in this one.

closes #509


nrjadkry:
This requires new release of osm-fieldwork. Config files are updated in the osm-fieldwork. Older config files wont work with these changes. @spwoodcock

Sujanadh:
@nrjadkry ok i added the package.

spwoodcock:
## Dependencies

For future ref, the lock file also needs to be regenerated for builds to work:

```bash
cd src/backend
pdm lock
```

It's done automatically if you add the package with: `pdm add osm-rawdata`.

Or install after updating manually: `pdm install`.

## osm-fieldwork

I'll make the new osm-fieldwork release tomorrow @nrjadkry and check / merge this PR.

We also need to update the make_data_extract refs in FMTM at the same time.

## Additional notes

@Sujanadh on first glance it looks like we remove the UNDERPASS_API_URL config from the code. Can this be added back in anywhere, so it's easily configurable?

PR #891, opened 2023-10-10
Replace make_data_extract PostgresClient with osm-rawdata

### LINKED ISSUE #848: Push osm-fieldwork latest release and update on FMTM
- Also requires updating the code for get_data_extract within FMTM for it to work.
- It should be a simple task to removing code, then calling the get_data_extract directly from osm-fieldwork.

spwoodcock:
Fixes #848


nrjadkry (approved):
Thanks @spwoodcock
Now that we are using osm-rawdata, we should pass "underpass" to the PostgresClient. 
@Sujanadh might have missed it.

PR #908, opened 2023-10-12
Add Minio to backend compose stack & CI

### LINKED ISSUE #892: Use distributed storage instead of container storage: S3

- Currently we download, process, and serve files directly from the FastAPI container.
- This is an acceptable strategy if running a single container, but it does mean that container storage can fill up.
- We have no way to cache submissions https://github.com/hotosm/fmtm/issues/713
- We want to serve cloud-optimised formats if possible https://github.com/hotosm/fmtm/issues/381, then fallback to offline (downloaded) files.
- We need to easily download generated files: https://github.com/hotosm/fmtm/issues/890 https://github.com/hotosm/fmtm/issues/889

- Distributed storage via S3.
- I propose we use [Garage](https://github.com/deuxfleurs-org/garage) as a lightweight alternative to S3 as part of our software stack.
  - For development and self-hosted installs of FMTM we would use Garage.
  - For production deploys on cloud infrastructure, we would use managed S3 buckets (AWS).
  - This needs to be configurable.
- Then whenever we process a file, it should be uploaded to S3 for later retrieval (by the frontend, another container, etc).

**Additional context**

Doing this would make it much easier to deploy with something like Kubernetes.

spwoodcock:
- Decided on Minio over Garage - it's has better assurances of maintained, and is time tested.
- Add Minio to docker-compose stack in development (+staging).
- Minio is not required in production - we use managed S3 buckets instead.


robsavoye:
Will this then only work with S3 ? I still need to be able to run everything locally.

spwoodcock:
You will have to run with the docker compose stack going forward, or run your own Minio instance (which is easy) and connect to that.

I will add docs for both 👍 

By locally you mean without docker right?
As this will run locally without issues (minio runs locally).
In production we will use managed S3 on AWS.

PR #909, opened 2023-10-13
Download submissions in  a single CSV filr

nrjadkry:
I have created an api endpoint which downloads all the submissions in a single csv file.

Endpoint: `projects/{project_id}/download_csv/`

This will download a zip file containing submissions csv and the media folder for all the medias uploaded in the form.

I have moved these scripts from scripts directory to the project directory and arranged them.

### DIFF
--- a/.github/workflows/pr_label.yml
+++ b/.github/workflows/pr_label.yml
@@ -1,7 +1,7 @@
 name: PR Label
 
 on:
-  pull_request:
+  pull_request_target:
 
 jobs:
   pr-label:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,7 @@ Right now, we are in the process of building the prototype. We warmly welcome yo
 Create pull requests (PRs) for changes that you think are needed. We would really appreciate your help!
 
 Skills with the following would be beneficial:
+
 - Python
 - FastAPI
 - Javascript
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,15 @@ Right now, we are in the process of building the prototype. We warmly welcome yo
 
 Create pull requests (PRs) for changes that you think are needed. We would really appreciate your help!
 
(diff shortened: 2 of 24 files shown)


spwoodcock:
Sorry for the lack of input here. Initially I didn't understand the requirements fully.

This sort of functionality for interacting with ODKCentral ideally belongs in osm-fieldwork, to be called from FMTM. I think Rob left a message saying this somewhere.

If you see an easy way to do that, then go ahead.

Otherwise, I will review fully next week and see what we can do 👍

robsavoye:
I'm not really sure we want to use any of the code "as is" from the scripts directory. Any interaction with ODK Central should be in osm-fieldwork/OdkCentral.py, and the FMTM backend should just use that,

robsavoye:
I don't think we need expand_geopoints(). This is already handled by osm-fieldwork/json2osm.py, which produces a GeoJson file. osm-fieldwork/CSVDump.py also does the same thing, but with the CSV file from Central. We should use existing code as much as possible to avoid duplication.

project_submissions_unzipped() should also be in osm-fieldwork, not project_crud.py. There is already support in osm-fieldwork for downloading all the data and media, which should be used instead.

robsavoye:
The JSON format is preferred over the CSV format as the CSV format can loose the location data when using data extracts. The JSON format doesn't appear to have the same issue.

nrjadkry:
Thanks @robsavoye, @spwoodcock 
I have received submissions from osm-fieldwork and processed them in FMTM. 
You guys are right. We should have this in osm-fieldwork, since this can me made modular.
I will make a new PR in osm-fieldwork.

robsavoye:
When you migrate this to osm-fieldwork, it can't be just a cut & paste thing. The code will need to be refactored to use functionality already in osm-fieldwork. Otherwise we wind up with code duplication.

nrjadkry:
Since we will be doing this osm-fieldwork, I am closing thus Pull Request.

PR #913, opened 2023-10-16
Addressing of project using Nominatim.

### LINKED ISSUE #833: Use project address from AOI using nominatim to show on project cards.
FireShot Capture 328 - Field Mapping Tasking Manager - fmtm naxa com np
P.S.: currently the description is there  but needs to have address of the project.

Sujanadh:
fixes #833 
This new addition generates location of project using centroid of respective AOI. I tried to generate the location for every existing project and return location as a summary of project. But it took long time to load. 
So i added feature to save location while creating project.
Instead of using a Python module, we're using the Nominatim URL for location retrieval. This change allows us to get project locations faster because the Nominatim URL is quick and responsive.

### DIFF
--- a/src/backend/app/projects/project_crud.py
+++ b/src/backend/app/projects/project_crud.py
@@ -453,15 +453,7 @@ def update_multi_polygon_project_boundary(
         result = db.execute(query)
         data = result.fetchone()
 
-        db_project.outline = data[0]
-        centroid = (wkt.loads(data[0])).centroid.wkt
-        db_project.centroid = centroid
-        geometry = wkt.loads(centroid)
-        longitude, latitude = geometry.x, geometry.y
-        address = get_address_from_lat_lon(latitude, longitude)
-        if address is None:
-            address = ""
-        db_project.location_str = address
+        update_project_location_info(db_project, data[0])
 
         db.commit()
         db.refresh(db_project)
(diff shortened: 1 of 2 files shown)


robsavoye:
Nominatim addresses are ok for navigation, but they aren't very accurate so we don't want to use them for the addr:housenumber tag. What is the purpose ?

Sujanadh:
it's true that Nominatim addresses may not always be as accurate as other sources, as far my knowledge in our use case, we are primarily concerned with obtaining a general project location at the center of the area of interest (AOI). 

The goal is to provide users with a representative location for their projects without significant delay.

robsavoye:
I'm not quite sure I understand. Is the address (the centroid of the project AOI) then mostly for mapper navigation ?

Sujanadh:
yes it is mainly used for mapper navigation. It serves as a reference point for mappers and users to easily locate and work on the project within their application.

spwoodcock:
Looks like a nice idea to me - it shouldn't add much overhead in project creation. 

Currently the user can specify the project location in the title or description, but this just adds a nice location string for where the project is located.

My only comment is that exactly the same code is duplicated 3 times in this PR (isn't DRY / is harder to maintain).

Could the code to add the address string to the project be put in a function in projects_crud, then calls 3 times instead?

robsavoye:
For a field mapping project, I'm not sure if everyone would go to the project center. Usually there is a location with sufficient parking or public transportation, usually not the center of the AOI. Also the centroid of the project could be under water, etc... So maybe the project manager should specify the "staging location" where everyone gets their morning briefing, etc... This is more of a project manager issue than a software one.

When I'm out field mapping but without FMTM, I commonly do make a GPX file of areas I want to map, a trailhead, parking lot, small town, etc... The GPX file I load into OsmAnd (or OrganicMaps) to navigate to that location. There's even a script in osm-fieldwork that adds OsmAnd styling and icons to the locations so they look pretty. I use it all the time when out ground-truthing.

nrjadkry:
@robsavoye Nominatim is used not for the mapper, but for this projects cards section.
image

For the mapper, project centroid wont be that much of a help, We have other features supporting mappers. Nominatim is used just for this projects list card in the dashboard, and not for the mappers. 
FYI: @Sujanadh

PR #918, opened 2023-10-17
feature/New UI for create project 

varun2948:
1. All Steps have been revised with changes in most of the functioning of the Create project Pages
here's a preview of the step 
image

You can check all the steps on the create project and tweak around a bit 
Probable bugs could be there with this PR letting @spwoodcock know that we will now gradually keep improving this feature with the requirement and bug fixes here and there. Sorry for this huge PR


varun2948:
@spwoodcock No, you can merge it now. I completed most of the functionality so little tweaks will happe on other PR, after this looks good on development server

varun2948:
@spwoodcock as i noticed now frontend test case have failed something to do with my package install or anything you know ?

spwoodcock:
Fixed 👍 
The pnpm-lock.yaml just needs to be rebuilt and pushed after adding a package.
`pnpm install` will do it.

spwoodcock:
Yes you should be able to

```
cd src/frontend
pnpm install
```

Are you getting an error?

You shouldn't have to install via docker.
That is done during the docker build of the frontend.
To rebuild the frontend image for running:
`docker compose build ui`

varun2948:
@spwoodcock Yeah not only me most of us were getting errors related to any package installed on the frontend side but when we do a `--no-build-cache build and docker compose down and docker compose up it goes away ` will create an issue based on this if again something like this comes up and put a real detail on the issue.

spwoodcock:
Ah yes, I see the issue - in the docker-compose.yml:

```yaml
    volumes:
      - ./src/frontend:/app
      - /app/node_modules/
```

The second volume prevents the local node_modules being mounted into the container.
This is nice, as we don't always have to pnpm install locally and docker build too.

The issue is that it prevents the node_modules updating when just running `docker compose up`.
We also have to run `docker compose down` first.

I think this could be handled better - thanks for raising 👍 (I'll look into it)

varun2948:
ah thats why it always acted weird installing new packages it didn't use local node_modules so had to do inside the docker container.

spwoodcock:
One extra thing.

Copying the pnpm-lock.yml seems to be have been missed in the frontend dockerfiles after the pnpm upgrade.

So adding these may help (I will push now).
The solution for the above (renewing node_modules after adding a dependency / rebuilding), is to use the flag: `docker compose up -d --renew-anon-volumes`, which will recrease the node_modules dir for you.

Issue #911, opened 2023-10-13
Error handling for AOI in projected coordinate system


susmina94:
**Describe the bug**
I uploaded an AOI which was in projected coordinate system. There was no issue with upload however, when I tried to split the task, I could not generate tasks and the api showed pending status only.

**To Reproduce**

1. Go to Create Project
2. Click on Upload AOI
3. Use any splitting algorithm
4. The tasks aren't generated
5. The api status is pending

**Expected behavior**
If the projected coordinate system isn't supported in the platform, it should've shown error while the user uploaded the file stating "The platform only supports file in geographic coordinate system" 

**Screenshots**

![Image](https://github.com/hotosm/fmtm/assets/108750444/c9703abb-c6e0-4e6d-8fa5-8329009f040b)

spwoodcock:
Good catch, thanks!

Geojson can have a optional `crs` field to specify the coordinate system, so even better would be if we could parse this and covert the geom to web mercator. 

Failing that, we give a not supported error, as you say 👍

robsavoye:
We want to only use an SRID of 4326, which is very common. Or we can add code to transform it to 4326.

spwoodcock:
EPSG 4326 as default for sure 👍

We should also support 3857, as this is a web map after all.

Support for anything other than that via the geojson crs key is just a bonus.

So two things we should consider to facilitate this:

- Allow loading in file formats other the geojson (often with better crs support inbuilt, e.g. geopackage).
- If a crs other than 4326 is identified, attempt to convert. Otherwise give an error.

robsavoye:
I think for a project AOI, at least when I make one with QGIS, JOSM, or ogr2ogr, it usually defaults to 4326. It would be easy to transform the crs, but I'm curious what software produced an AOI with a different crs ?

spwoodcock:
You are right that GeoJSON default is generally EPSG 4326.

This is not necessarily an optimisation we need if we are only looking at GeoJSON.
But GeoJSON is a format for the web, and many GIS users work with other formats, such as geopackage, shapefiles, you name it.

Although we don't support e.g. geopackage upload now, we need to be able to handle different CRS if the user uploads one other than EPSG 4326. At the bare minimum, we need error handling to tell them their CRS is not supported.

PR #919, opened 2023-10-17
Inform user if GeoJSON CRS is not EPSG:4326

Sujanadh:
fixes: #911 
This pull request aims to standardize the Coordinate Reference Systems (CRS) in our GeoJSON data by converting them to the  EPSG:4326 standard.
pyproj is used to change the coordinate reference system.


spwoodcock:
GeoJSON can have a `crs` property, but rarely is it ever set.

What happens if we pass a GeoJSON without this property?
We should fallback to parsing as EPSG 4326.

robsavoye:
I have had no problems using GeoJson files with no crs as AOIs, and it appears 4326 is the default.

nrjadkry (commented):
@spwoodcock  it looks like @Sujanadh  have used pyproj.
It it okay if we used this package?

spwoodcock:
@robsavoye `crs` isn't a mandatory key for GeoJSON, and it's rarely used. There is no default per se, just software that parse geojson will generally assume it's in EPSG4326.

See https://github.com/hotosm/fmtm/issues/911#issuecomment-1767551816 too.

robsavoye:
I'm curious what program generated a GeoJson files with a non 4326 crs ?

spwoodcock:
@nrjadkry good point - we will need to install system package `libproj`, which isn't ideal.

IMO we should shelve this PR for now, until we support formats other than GeoJSON.
Determining the CRS of a GeoJSON, if the user doesn't provide a `crs` key is difficult.
If the user uploads e.g. a geopackage, then it's much easier to handle crs.

robsavoye:
I think for a project AOI, we should stick to GeoJson. A shapefile is usually not just a polygon, it's more like a FeatureCollection. It's possible to transform not using libproj, but I think for now we can default to GeoJson, and if it's another format, they can use ogr2ogr to transform the projection.

spwoodcock:
Actually, ignore my previous comment. We currently install `libproj` anyway, probably for shapely.
Adding pyproj is fine.

To make this workable for now, @Sujanadh could we please:
- Update the code to skip the CRS check if the `crs` key is not present in the GeoJSON.
- If the `crs` key is present. then transform to EPSG 4326.
- If the GeoJSON can't be transformed, return an error message to inform the user.

nrjadkry:
Is it okay, if we throw an exception if it is in other format than 4326, and asks users to have tht geojson in 4326?

spwoodcock:
Yes, that also works.
If the `crs` key is present, and anything other than 4326, then throw an error and inform the user.

spwoodcock:
@robsavoye this isn't a priority, but GeoJSON isn't necessarily the most common format for GIS users. Others, even KML, would be useful. We can wait to see if any user feedback suggests this though & forget about it for now.

The nice thing is we can probably just tell users to go to geojson.io to generate one easily.

Saying 'just use ogr2ogr2' isn't an option for many users.

robsavoye:
What does TM use for an AOI ? I think just GeoJson or drawing the polygon by hand.

spwoodcock:
@nrjadkry

I was going to suggest that we update the error handling for the `preview_tasks` endpoint to account for the scenario in #911.

But I think a better solution is to fix the crs parsing, so this shouldn't happen hopefully.

In addition to checking for invalid values for the `crs` key, we should also parse the coordinates and check they are inside bounds: -180.0 -90.0 180.0 90.0 (EPSG 4326).
If they are not, then also throw an error.

spwoodcock:
@robsavoye good point, having the 'draw by hand' feature working solves this issue.

nrjadkry:
> @nrjadkry
> 
> I was going to suggest that we update the error handling for the `preview_tasks` endpoint to account for the scenario in #911.
> 
> But I think a better solution is to fix the crs parsing, so this shouldn't happen hopefully.
> 
> In addition to checking for invalid values for the `crs` key, we should also parse the coordinates and check they are inside bounds: -180.0 -90.0 180.0 90.0 (EPSG 4326). If they are not, then also throw an error.

Okay @spwoodcock , we will make these changes.
1. We will check if the crs key is present. if crs is other than 4326, throw an error.
2. Check the bounds for the coordinates,  if they are not in -180.0 -90.0 180.0 90.0 (EPSG 4326), we throw an error.

Sujanadh:
This new changes removes the conversion function and simply handles the geojson and return error message if other than WGS 84/EPSG:4326.

spwoodcock (approved):

PR #924, opened 2023-10-19
700 warning message before project creation action is cancelled

varun2948:


spwoodcock:
Very nice 🙌 What's the reference to 700 about though?

spwoodcock (approved):

varun2948:
I am also confused about the number 700 :smile:  as I tried making a direct branch from the issue it gave me 700 as the branch name prefix.

PR #959, opened 2023-11-02
Added Pagination info in the project summary

### LINKED ISSUE #944: Redesign pagination & project summaries endpoint
## Original Issue

**Describe the bug**
Since we have the infinite scroll in projects page, the text displaying the range doesn't make sense. 

**To Reproduce**

1. Go to 'Explore Projects"

**Expected behavior**
Add the text "Total Projects : N" , where N = project count instead of showing n of n projects

**Screenshots**

Image

## New Issue

- Pagination of the backend is partially addressed by: #225
- This is an approach where we can only limit the number of results, not actual pagination (?page=1, ?page=2, etc).
- We need true pagination as we grow, or the API call to fetch projects will either return too much data, or we will be unable to return data past ?limit=100.

## Proposed Solution

**Backend**

- Add pagination to the `summaries` endpoint so we can include ?page=1 syntax.
- Only return approx 13-20 project - assess what is best.
- Move the results list under the key `results`.
- Create one other top level key in the JSON called `pagination`, where pagination details are included (number of pages, total number of projects, etc).

**Frontend**

- Add `Showing x of x projects` to the home page, based of `pagination.total` key.
- Access the project summary content via the `results` key instead.

Sujanadh:
fixes #944 

This pull requests adds new pagination info in project summary.
Screenshot from 2023-11-02 14-31-13

### DIFF
--- a/src/backend/app/projects/project_routes.py
+++ b/src/backend/app/projects/project_routes.py
@@ -109,12 +109,12 @@ def get_task(lat: float, long: float, user_id: int = None):
     return "Coming..."
 
 
-@router.get("/summaries", response_model=List[project_schemas.ProjectSummary])
+@router.get("/summaries", response_model=project_schemas.PaginatedProjectSummaries)
 async def read_project_summaries(
     user_id: int = None,
     hashtags: str = None,
-    skip: int = 0,
-    limit: int = 100,
+    page: int = Query(1, ge=1),  # Default to page 1, must be greater than or equal to 1
+    results_per_page: int = Query(13, le=100),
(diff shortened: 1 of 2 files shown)


spwoodcock:
Thanks @Sujanadh 

@NSUWAL123 this PR will move the project summary list under a `results` key.

We would need frontend updates for this to work (and to also handle the pagination).

Could you both coordinate on this PR, so that the backend and frontend changes are made together?
(or if easier, make the frontend edits in another PR & we can merge them at the same time)

NSUWAL123:
okay @spwoodcock

PR #970, opened 2023-11-06
Qr code in task list

Sujanadh:
changed schema to add qr code in response.


spwoodcock:
It may be best to add the qr code to TaskOut instead of changing the response model to Task.

We have a fair bit of duplication in the schemas, but generally `<schema>Out` is used for a response, which in this case it is.

What do you think?

Sujanadh:
okay @spwoodcock since this out schema is used for reponse only, i would do so.

PR #973, opened 2023-11-06
Fix tasks api replacement

NSUWAL123:
In this PR, the project_tasks data is fetched from **tasks/task-list?project_id=id** which was previously fetched from **projects/:id**.


spwoodcock:
A lot of code is commented out in the PR - could it be deleted instead?

If we need to get the old code it will still be in the git history 👍

NSUWAL123:
Okay, I'll just delete it and push again.

PR #976, opened 2023-11-07
task id on task polygon and feature extracts

### LINKED ISSUE #964: Add task ids on task_polygon file and task id, feature id on data extract file
I wanted to use the splits generated from FMTM for my personal project. It is really great that I can download them but unfortunately there are no attributes related to the file (For eg: It would've been really helpful if we could fetch the task ids for each polygons)

Also, the same goes with data extract geojson file. 
I would like to pre fill the attributes for each feature id with the task id to which they are associated with.
As of now, there is a field "FID" but I don't think they are unique feature ids because there are null values in few of the rows.

I could manually do this using any GIS software but would've been very convenient if FMTM provided it by default.
Would save a lot more time for further processing.

Sujanadh:
fixes #964 
added task_id on the task polygon and feature extracts file.

### DIFF
--- a/src/backend/app/projects/project_crud.py
+++ b/src/backend/app/projects/project_crud.py
@@ -1723,7 +1723,7 @@ async def get_project_features_geojson(db: Session, project_id: int):
 
     for feature in features["features"]:
         if (feat_id := feature["id"]) in task_feature_mapping:
-            feature["properties"]["task_id"] = task_id_mapping[feat_id]
+            feature["properties"]["task_id"] = task_feature_mapping[feat_id]
 
     return features
 
--- a/src/backend/app/projects/project_crud.py
+++ b/src/backend/app/projects/project_crud.py
@@ -1673,7 +1673,7 @@ def get_task_geometry(db: Session, project_id: int):
     Returns:
         str: A geojson of the task boundaries
     """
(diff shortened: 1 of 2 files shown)


spwoodcock:
@Sujanadh can you address the looping issue when you get a chance please?

I'll merge as soon as you do 👍

susmina94:
@spwoodcock 
**Problem**

We have a few users using FMTM splitting algorithm to create just the splitted polygons and download the splits and data extracts for each of them.

But it seemed that there are no linkages between the splitted polygons and the data extracts that contained within them.
The attributes for the splits and data extract only had a unique id column.

They had to manually add columns to the data extract to assign the ids of split polygons to each data extract 

**Requirement**

They requested if we could generate task ids and also map the task ids to each feature in the data extract that lies under the specific task.

spwoodcock:
@susmina94 thanks for the context =)

I have no issue with the rationale for the PR, just the implementation is inefficient.
I will make the changes I proposed in the code review👍

Sujanadh:
@spwoodcock i have been working on other issue , i will get to back to this as soon as i complete other one.

spwoodcock:
Oh, sorry @Sujanadh - I just made the update 😄

Issue #912, opened 2023-10-14
Termination of background tasks when user cancels the project setup activity


susmina94:
**Describe the bug**
I created a project, had some issues with splitting the task (which was the platform not supporting the file in projected coordinate system), I couldn't proceed ahead so I just refreshed the tab and started creating new projects one after another. 

After sometime, the platform didn't respond. 
When I asked the devs, they mentioned the previous incomplete project setups kept the background tasks running even though I cancelled it or refreshed the tab.

The reason being all the workers were occupied with the previous project background tasks.
They had to restart the system manually and then it was back to its working state.

**Expected behavior**
The first behavior would be :

1. If the user cancels the project setup, it should stop all the background tasks related to it so that the workers are not loaded unnecessarily
2. If there is any error in input, validations and proper error messages should be displayed for the users to try again.

**Screenshots**
![image](https://github.com/hotosm/fmtm/assets/108750444/01fc1cf5-ca55-43b3-b7b7-a429b56990b9)

spwoodcock:
There are two issues to consider here:

### Event Loop Blocking

- Running as async function within the background task will run it in the asyncio event loop.
- If there are no more workers to handle the request (set to 4), then FastAPI will hang.

More context here: https://stackoverflow.com/questions/67599119/fastapi-asynchronous-background-tasks-blocks-other-requests

### Cancelling Jobs

- We track background jobs in our DB (background_tasks table).
- Their status is updated when complete.
- BackgroundTasks cannot be easily stopped though. 
- Ideally we need a way to collect running background tasks associated with an endpoint (tagged with project ID), then cancel existing jobs if the user cancels, goes back, refreshes, etc.

### Possible Solution

We discussed using celery before & agreed it's too heavyweight. 

I agree - scalability is also a pain, so using the database to manage tasks is a better solution IMO.

[apscheduler](https://github.com/agronholm/apscheduler) looks like a good tool for our use case, to replace the simple BackgroundTasks.

It allows for tasks to be cancelled, paused, resumed. It could probably be implemented ourselves, but this is simple.

It can be set up with FastAPI asyncio loop:

```python
from apscheduler.schedulers.asyncio import AsyncIOScheduler

sched = AsyncIOScheduler()
sched.start()
```

Then jobs can be spawned from called endpoints with an ID associated (note no async).

```python
def job():
    sched.add_job(download_tiles, "date", run_date=datetime.now(), args=[], id=<unique_id>
[...]

PR #977, opened 2023-11-07
added apscheduler

nrjadkry:


nrjadkry:
@spwoodcock 
I just realized that date trigger adds a job in the appscheduler_jobs table it creates in the db and then removes it immediately.
This might be a small problem if we want to use this table for our usecase.

spwoodcock:
Do we need to trigger on date? This is for scheduling things in the future.

I think our use case is just to be able to return the job id, so we can queue, cancel, or pause the job.

nrjadkry:
@spwoodcock 
There are 3 different triggers in apscheduler.
1. Interval trigger: Schedules a job to run at fixed intervals.
2. Cron trigger
    Schedules a job based on a cron-like expression.
    Options: year, month, day, week, day_of_week, hour, minute, second: Specify the cron expression.
3. Date trigger.
    Schedules a job to run once at a specific date and time.

Date trigger is the best one for our use case, but it creates the job in the db and then removes it immediately after the job is started.

Are there any other triggers that I have missed?

spwoodcock:
See #912

We decided not to proceed with this.

PR #978, opened 2023-11-07
uuid in read project and total_task in generate-log

Sujanadh:
added uuid in read project and total task count in generate log


spwoodcock:
What is the uuid used for?

It's not saved to the database, so it will be different every time you call the API - is that intended?

spwoodcock:
@Sujanadh just wondering if I should merge this? What is the use case for the uuid?

nrjadkry:
@spwoodcock , we can merge this now. It should work fine now since we won't be using apscheduler

PR #982, opened 2023-11-08
task wise mbtile generation

### LINKED ISSUE #890: Allow downloading of mbtiles per task
Problem:
The mbtiles are too large and takes up more memory space if I download the entire mbtile of the project area. 

Solution:
I want to download only the mbtiles of assigned task 
This will save my memory space and easy to load.

Sujanadh:
fixes #890 
this changes allows to generate the mbtiles of individual task using their task_id.  if task_id is not provided it will generate the mbtiles of entire project boundary.

### DIFF
--- a/src/backend/app/projects/project_crud.py
+++ b/src/backend/app/projects/project_crud.py
@@ -2357,10 +2357,10 @@ def get_project_tiles(
     try:
         db.add(tile_path_instance)
         db.commit()
-        
+
         get_id = task_id or project_id
         entity = "tasks" if task_id else "projects"
-        
+
         # Project/task Outline
         log.debug(f"Getting bbox : {get_id}")
         query = text(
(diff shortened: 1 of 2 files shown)


spwoodcock:
For this to work we would also need to update the frontend.

But please also see the comments in the linked issue & how Rob proposed we solve this.
We still want to generate the tiles for the entire project, then subdivide.
I proposed a way to do this with project level PMTiles, then dividing into task level MBTiles on request.

I will probably make a PR for this in the coming weeks.

Sujanadh:
so the tile of entire project must be generated not the task wise? and allowing to download tile of required task from the generated tile of entire project?

spwoodcock:
Yes exactly.

osm-fieldwork does support outputting PMTiles, so we should do that.
The subdivision from PMTiles into MBTiles won't be simple though & will require delving into the format internals.

If you want to have a go, feel free, otherwise I will be working on this soon.

spwoodcock:
Rob already gave his feedback for how to solve this: https://github.com/hotosm/fmtm/issues/890#issuecomment-1758590525

But he can always weigh in again.

Issue #961, opened 2023-11-03
Upload AOI doesn't support projected coordinate system however, appropriate error messages and validation is needed.


susmina94:
**Problem Statement:** 

I uploaded an AOI that is in projected coordinate system (44N), it accepted the file and I proceed ahead with project setup. But at the end of the project creation, I received the following error.

The problem occurred because the file was in projected coordinate system.

**Expected behavior**
Validation of the AOI should've been done when the user uploads the file. Should show an error message stating "Invalid spatial reference system. Please only import WGS84 (EPSG: 4326)".

**Screenshots**
![Image](https://github.com/hotosm/fmtm/assets/108750444/ba4b486e-1b95-42ad-8997-5b3ec579e8c3)

spwoodcock:
If it's easy to get the coordinate system from the file in a consistent way (it's often hard to tell without the user specifying), then we can display the error message as you say: "Invalid spatial reference system. Please only import WGS84 (EPSG: 4326)."

But if it's more tricky to get the coordinate system, then we could fallback to a more generic: "AOI is invalid. Please check the coordinates and ensure EPSG:4326 is used."

susmina94:
![image](https://github.com/hotosm/fmtm/assets/108750444/448ded3f-5e92-4193-a34e-5b4e4e08b406)

The validation has been done in backend in the last step.
However, the validation should've been done right at the time when the user uploads the file.
@nrjadkry dai

NSUWAL123:
@susmina94 didi, the validation has been done in the upload area step itself but its not reflecting on the development server. 
@spwoodcock the PR for validating the coordinate system has already been merged but its not working here. Though it works completely fine on my local host.
![image](https://github.com/hotosm/fmtm/assets/81785002/76704c64-2eac-4aef-b5d7-ccc14aa84ae2)

susmina94:
Tested on https://dev.fmtm.hotosm.org/ and is working fine.

PR #987, opened 2023-11-20
fix (createNewProject): uploadArea - validation added to upload file,…

NSUWAL123:
This PR solves issue #961 where error is shown if the uploaded files projection system is not WGS84.
image


src/frontend/src/components/createnewproject/UploadArea.tsx
```diff
@@ -101,6 +102,56 @@ const UploadArea = ({ flag, geojsonFile, setGeojsonFile }) => {
     dispatch(CreateProjectActions.SetTotalAreaSelection(null));
   };
 
+  useEffect(() => {
+    function isWGS84GeoJSONBasedOnCoordinates(geojson) {
```
varun2948, line 106: @NSUWAL123 can you keep this function Seperately to a utils folder functions file not here as it can be used in multiple places

varun2948 (approved):

Issue #986, opened 2023-11-14
Add task_split_type to project when task splitting complete


spwoodcock:
**Describe the bug**

- We have the database field `task_split_type` for projects.
- Three options for splitting: square, manual, or task splitting algorithm (fmtm-splitter).
- The `task_split_type` is not updated in the database after the splitting is complete.
- We have a field `splitting_algorithm` that is used in the frontend that should be renamed.

**To Reproduce**

- Create a project.
- Check the request body.
- Check the db.

**Expected behavior**

- The `task_split_type` field is updated.

## Backend Tasks

- Create an enum for `task_split_type`.
- Ensure the field is updated on project creation.

## Frontend Tasks

- Rename field `splitting_algorithm` to `task_split_type` throughout the frontend code.
- Check this is a suitable replacement in all components (`splitting_algorithm` may be used to track the component state / user selection, but the field `task_split_type` field must be included in the project creation POST request - using one of the enum options defined above).

## Additional considerations

Even if we have the `task_split_type`, we do not store the `dimension` for split on square, or the `no_of_buildings` per task for the splitting algorithm. These params would be required to reproduce a project (i.e. via an export/import).

spwoodcock:
The param is passed by the frontend, just not saved:

![image](https://github.com/hotosm/fmtm/assets/78538841/31888886-c5b1-428b-892e-a2b97c33e75c)

spwoodcock:
Backend part done, thanks @Sujanadh.

@NSUWAL123 or @varun2948 could you please update the frontend to use the param `task_split_type` instead of `splitting_algorithm`.

The param is an integer (enum on the backend) instead of a string.
RadioButton doesn't support type **number**, so you will have to `parseInt` somewhere.

It would be great if you could also post the params:
- `task_split_dimension` - the dimension used for divide on square
- `task_num_buildings` - the number of buildings selected for splitting algorithm.

NSUWAL123:
@spwoodcock, before merging please review the params for project creation.

PR #990, opened 2023-11-20
added task_split_type to project when task splitting complete

Sujanadh:
This pull request addresses the issue #986  the task_split_type field not being saved in the database. 

changes:
- Added an enum field for task_split_type.
- Implemented the necessary changes to ensure the task_split_type is saved in the database.


spwoodcock (commented):
This is great. The only thing is that the enum should probably map to a number value for the database.

Check the other enums in the file for reference 😃

But to do this the field would have to be changed to an integer, which would require a migration 🫤

I'll write a migration file for this PR that also includes some other db updates we need (to save making many separate ones), if you could address the enum 👍

Sujanadh:
so we are using int enum for memory efficiency. ok, @spwoodcock  i will make it int enum if you could make migration file.

Sujanadh:
@spwoodcock i have created migration file too please do check it once.

PR #1001, opened 2023-11-22
Refactor bytesio .seek(0) & .read() methods with .getvalue()

spwoodcock:
Minor PR to refactor:

```python
obj = BytesIO()
obj.seek(0)
obj.read()
```
with

```python
obj = BytesIO()
obj.getvalue()
```

The `.getvalue` method reads the entire content of the buffer, so no seek required.

@nrjadkry @Sujanadh useful to know 👍

### DIFF
--- a/src/backend/app/central/central_schemas.py
+++ b/src/backend/app/central/central_schemas.py
@@ -17,7 +17,6 @@
 #
 from enum import Enum
 
-from loguru import logger as log
 from pydantic import BaseModel
 
 
--- a/src/backend/app/main.py
+++ b/src/backend/app/main.py
@@ -19,8 +19,8 @@
 
 import logging
 import sys
-from typing import Optional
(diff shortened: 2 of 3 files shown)


spwoodcock:
😂 I was actually wrong here!

While it's true for BytesIO objects (that UploadFile inherits from), the FastAPI UploadFile class is actually an async wrapper, that's doesn't have all the same methods!

So while `BytesIO.getvalue()` is the Python way, for FastAPI it should be `await UploadFile.read()`.
The UploadFile.seek() method is not required unless you read the file in twice.

Reverted: https://github.com/hotosm/fmtm/commit/b2ba19149225ba33f6c38f89c7d2efe486bf980a

spwoodcock:
image

MOBILE-VIEW

spwoodcock:
image

spwoodcock:
image

PR #1008, opened 2023-11-24
Upload organisation logo in S3.

nrjadkry:
## Changes made
1. Added `S3_BUCKET_NAME_BASEMAPS` and `S3_BUCKET_NAME_OVERLAYS` in the config file
2. Upload organisation logo in the `fmtm-data` bucket in the s3.

I have used `S3_DOWNLOAD_ROOT` to get the image path. @spwoodcock is it okay?
## Todo
1. We might need to migrate existing images of organisation logo which are in the container to the s3 and update the url  path in db.
2. Use this url to display the logo in the frontend.

### DIFF
--- a/src/backend/app/config.py
+++ b/src/backend/app/config.py
@@ -113,8 +113,6 @@ class Settings(BaseSettings):
     S3_ACCESS_KEY: Optional[str] = ""
     S3_SECRET_KEY: Optional[str] = ""
     S3_BUCKET_NAME: str = "fmtm-data"
-    S3_BUCKET_NAME_BASEMAPS: str = "basemaps"
-    S3_BUCKET_NAME_OVERLAYS: str = "overlays"
     S3_DOWNLOAD_ROOT: Optional[str] = None
 
     @field_validator("S3_DOWNLOAD_ROOT", mode="before")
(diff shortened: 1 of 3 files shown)


spwoodcock:
Looks good, but I intentionally removed those two env vars from the setup. I consolidated into a single bucket defined by `S3_BUCKET_NAME`.

To have all the data in the same bucket, we need a consistent directory structure.

What do you think about:

`ROOT/{ORG_ID}/{PROJECT_ID}`? 

So logos would go in
`ROOT/{ORG_ID}/logo.png`

(if we consistently named it logo.png we could even remove the database record for the file)

For the PR I am working on for project export it would be:
`ROOT/{ORG_ID}/{PROJECT_ID}/export.zip`

nrjadkry:
Sounds good, we can store data in this format `ROOT/{ORG_ID}/{PROJECT_ID}` . 
But, we should be extra careful  in maintaining this consitency. 
I will remove those two env var from config file.

spwoodcock:
Yes using `S3_DOWNLOAD_ROOT` for the download url is perfect 👍 

I took the opportunity to fix the linting errors on the file for pre-commit.
Also updated the error handling, with organization rollback on exception.

The logo always goes to `/{org_id}/logo.png` for every organization, so we consistently know where the logo is.

spwoodcock (approved):

Issue #1002, opened 2023-11-22
Add a template for custom data extract upload


susmina94:
**Describe the bug**
I chose the default xlsform provided by FMTM and uploaded a custom data extract.
When I tried to open the form via ODKCollect, it showed me an error as shown in the screenshot below:

The error states that the ODK is searching for the title column in the data extract. 
The syntax occurs if the form includes a calculation element with the SelectFromMap feature where the id and geometry of the selected feature is calculated in the form.

In absence of the title column in the attribute, we will see this error message while we try to open the form in ODKCollect.

**To Reproduce**

1. Go to Create Project
3. Click on 'Select Default form'
4. Click to upload custom data extract
5. Submit to create a project
6. Go to ODKCollect
7. Scan the QR Code
8. Select the form to fill up
9. You will see the error

**Solution**
When the user chooses to upload a custom XLS form, we provide then with a sample template of the XLS form.

Similarly, we shall provide a sample template of the data extract while the user tries to upload a custom file.
The template shall include all the necessary attributes and columns as per the need.
I have attached a sample file that we can use as the template for the data extract.

**Screenshots**
![Screenshot_2023-11-22-14-43-40-412_org odk collect android](https://github.com/hotosm/fmtm/assets/108750444/aed6fd2f-fb94-45a2-a72a-2e070cd1b948)

Issue #934, opened 2023-10-25
Use data extracts hosted in raw-data-api


spwoodcock:
## Issue

- Currently we keep track of data extracts within the FMTM database.
- We request a data extract from raw-data-api, then input the resulting geojson data into our db.
- When we make a request for a subset of the extract (i.e. a task area), it comes from the db.
- If we need to update the data extract, we request/download again & then update the db.

## Solution

All of the above can be handled by raw-data-api:

- Request a data extract to be generated in flatgeobuf with bind_zip=False (i.e. not zipped).
- Normally the data only lasts for 90 days, but if we use a filename of `hotosm-fmtm-project-{project_id}`, the data will be persisted.
- If we request another extract for the same file, the file will be updated in place.

Essentially we remove the burden of handling data extracts ourselves, and outsource it to raw-data-api.

Tasking manager is going to do something similar.

The resulting flatgeobuf has a spatial index inbuilt and can be accessed via RANGE request, so we can subset the data for a specific task area, the same way we do now.

To do this, we pass these params via POST to the `/snapshot` endpoint:

```json
{
    "outputType": "fgb",
    "fileName": "hotosm-fmtm-project-{project_id}",
    "bindZip": false,
    "geometry": {...},
    "filters": {...}
}
```

#### Environment

- We should also probably consider what environment we are doing this on: main, staging, development:

**main**: the filename can be `hotosm-fmtm-project-{project_id}` (will exist forever)
**staging**: the filename can be `hotosm-fmtm-staging-project-{project_id}` (will exist for undefined period, perhaps 180 days?)
**development**: the filename can be `fmtm-project-{project_id}` (will expire in 90 days with no `hotosm` prefix)

The final S3 path will be something like:

```
S3_URL/fmtm/{project_id}/hotosm-fmtm-project-{project_id}.fgb
```

#### Redundant Tables

- We can completely remove the database tables and models for managing data extracts once this refactor is complete.

- Currently we have DB tables DbOsmLines (ways) and DbBuildings.

- Do we need a flatgeobuf for each, or can we just filter in the frontend by a key? In the final parsed geojson we will have the property `building=true|false`.

robsavoye:
[...]

PR #1011, opened 2023-11-27
Upload template files to s3

Sujanadh:
fixes: #1002 
Created an api to upload template files. Template file names are made static so that every time new template is uploaded it will replace the existing file .


spwoodcock (commented):
The template path should be nested under the org and the project.

I would say the path should be:
`f'{org_id}/{project_id}/form.xls`

As for the data extracts, they won't be stored in our S3 bucket.
Please see #934
A link to the extract is stored in our DB, but the extract file itself will be stored externally in flatgeobuf format.

nrjadkry:
This is just a template for the users wanting to upload a custom data extracts.
Custom data extracts requires `title` field in the properties and few more of them. If the user misses those fields, odk collect throws an error. I have kept an empty title field for those features without the title field from the code. But, providing the user with the template and allow them to view the format they require might be better.

nrjadkry:
Yes, you are right. API is actually not required here. We can do this just through a simple python script.
I wanted to confirm you with the templates folder structure in S3.

@Sujanadh , It might be better to use python script rather than these unauthenticated apis.

spwoodcock:
The easiest place to put it for now is probably:

`backend/app/static` (make the dir)

Then we can serve any static files we have from an endpoint.

The endpoint could be e.g. `/template` and it returns a File response of the template file.
The frontend can then use it too.

Sujanadh:
so i should just change the endpoint and return file response from `static` dir ?

Sujanadh:
@spwoodcock @nrjadkry

spwoodcock (commented):
Nice, looks good 👍

Issue #915, opened 2023-10-16
The map interface in project dashboard requires orientation tool


susmina94:
**Is your feature request related to a problem? Please describe.**
I want to identify the map orientation in project dashboard. 
Also, if I click to view my GPS location, I am not sure where and in which pathway I am facing towards.

**Describe the solution you'd like**
The map orientation could be identified by simply adding on a north arrow icon.
For the GPS location, we can have an icon showing the orientation with a tooltip for eg: 

![Image](https://github.com/hotosm/fmtm/assets/108750444/91d05119-9843-4fbc-a584-9cde5406e235)

varun2948:
@spwoodcock i am having issue implementing orientation with arrow on the geolocation its spinning like crazy when keeping my phone stable too any idea on this i kinda think openlayers is not supporting accurate gps orientation

varun2948:
https://github.com/hotosm/fmtm/assets/37866666/126a33c8-317e-4321-932e-73870d04ecff

spwoodcock:
I wouldn't use OpenLayers for mobile geolocation - it's a very basic implementation and don't interface with mobile APIs.

Use Capacitor instead: https://capacitorjs.com/docs/apis/geolocation
I have implemented this before: https://gitlab.com/mastweb/frontend/-/blob/main/src/pages/NewSitePage.vue?ref_type=heads

But to save battery, I made it so the user has to click to refresh their location.
It is possible to put a watch on geolocation and have it live update: https://capacitorjs.com/docs/apis/geolocation#watchposition

The advantage of this approach is that it can use the native mobile API if we make FMTM installable via appstore (as PWA). Otherwise it falls back to navigator.geolocation in all browsers.

spwoodcock:
Never used this one, but to capture the orientation I think you also need: https://capacitorjs.com/docs/apis/motion

NSUWAL123:
@spwoodcock, using the capacitor partially solves the problem but in terms of UX wise is not great.  The orientation arrow shows the correct orientation but flickers most of the time. Here I have used `Geolocation` for location and `Motion` for orientation.
https://github.com/hotosm/fmtm/assets/81785002/debc8bde-ae8f-42ba-af80-95c3c4af8739

spwoodcock:
Hmm that's not great.

I actually think that using Capacitor directly from the web would have the same result as the OpenLayers implementation (it falls back to the web navigator implementation if the mobile APIs can't be accessed).

[...]

PR #1016, opened 2023-11-29
Feat capacitor geolocation

NSUWAL123:
- Test #915 
This PR integrates `geolocation` & `motion` from `capacitor`. Also the current geolocation fetch from the navigator is replaced with capacitor. There are more enhancements to be done for device orientation.


spwoodcock (approved):

NSUWAL123:
@spwoodcock the `package-lock.json` is included on the `gitignore` file due to which it's not being pushed.

spwoodcock:
It's not the package-lock.json we need.
We are using pnpm for the project, so it's `pnpm-lock.yaml` 👍

PR #1026, opened 2023-12-05
Fix vectorlayer style

NSUWAL123:
This PR replaces the string `taskStatus` with enums for getTaskStatusStyle, updateStatus, and displayStatus.


spwoodcock:
Good catch! Thanks @varun2948 and @NSUWAL123 🎉

PR #1031, opened 2023-12-06
Returned task status as string

Sujanadh:
Previously, int value was retruned in the response which created conflicts in the old projects. With these changes , it accepts integer value of enum and saves it in db. But while returning it passes "name" of the corresponding "value" resolving probable conflicts.


spwoodcock:
Will this conflict with the frontend fix by @NSUWAL123 and @varun2948 ?

NSUWAL123:
@spwoodcock, now task status string will be used to show layers and data in frontend as well. The task status enum will only be used while updating(post) status. As soon you merge this PR I will work on it and will update the frontend part.

spwoodcock:
Thanks guys 😄

PR #1035, opened 2023-12-07
Feat cache submissions

### LINKED ISSUE #713: Cache submissions
There are a number of times FMTM downloads the submissions. For better performance, they should probably be cached.

nrjadkry:
Now that we have decided to add submission cache in s3, I have created an endpoint that adds the submission json in the s3 bucket. 
Another api to download the submission from cache.
The path in the s3 bucket is like this: `/{organisation_id}/{project_id}/submission.json`

This Pull Request tries to 
- Fix #713

### DIFF
--- a/src/backend/app/submission/submission_crud.py
+++ b/src/backend/app/submission/submission_crud.py
@@ -456,10 +456,14 @@ def update_submission_in_s3(
         file_obj = BytesIO(json.dumps(submission).encode())
 
         import sozipfile.sozipfile as zipfile
-        with zipfile.ZipFile(file_obj, 'w', 
-                            compression=zipfile.ZIP_DEFLATED,
-                            chunk_size=zipfile.SOZIP_DEFAULT_CHUNK_SIZE) as myzip:
-            myzip.writestr('submission.json', json.dumps(submission))
+
+        with zipfile.ZipFile(
+            file_obj,
+            "w",
+            compression=zipfile.ZIP_DEFLATED,
+            chunk_size=zipfile.SOZIP_DEFAULT_CHUNK_SIZE,
+        ) as myzip:
+            myzip.writestr("submission.json", json.dumps(submission))
 
         add_obj_to_bucket(
             settings.S3_BUCKET_NAME,
--- a/src/backend/app/submission/submission_crud.py
+++ b/src/backend/app/submission/submission_crud.py
@@ -473,18 +473,11 @@ def update_submission_in_s3(
         )
 
         # Check if the file already exists in s3
-        s3_path = f"/{project.organisation_id}/{project_id}/submission.zip"
+        metadata_s3_path = f"/{project.organisation_id}/{project_id}/submissions.meta.json"
         try:
-            file = get_obj_from_bucket(settings.S3_BUCKET_NAME, s3_path)
-
-            with zipfile.ZipFile(file, "r") as myzip:
-                metadata_index = myzip.getinfo("metadata.json")
-                with myzip.open(metadata_index) as file:
-                    metadata_json = file.read()
-
             # Get the last submission date from the metadata
-            zip_file_last_submission = (json.loads(metadata_json))["last_submission"]
-
+            file = get_obj_from_bucket(settings.S3_BUCKET_NAME, metadata_s3_path)
+            zip_file_last_submission = (json.loads(file.getvalue()))["last_submission"]
             if last_submission <= zip_file_last_submission:
                 # Update background task status to COMPLETED
                 update_bg_task_sync = async_to_sync(
@@ -514,13 +507,20 @@ def update_submission_in_s3(
             chunk_size=zipfile.SOZIP_DEFAULT_CHUNK_SIZE,
         ) as myzip:
             myzip.writestr("submissions.json", json.dumps(submissions))
-            myzip.writestr("metadata.json", json.dumps(metadata))
 
         # Add zipfile to the s3 bucket
         add_obj_to_bucket(
             settings.S3_BUCKET_NAME,
             submissions_zip,
             f"/{project.organisation_id}/{project_id}/submission.zip",
+        )
+
+        # Upload metadata to s3
+        metadata_obj = BytesIO(json.dumps(metadata).encode())
+        add_obj_to_bucket(
+            settings.S3_BUCKET_NAME,
+            metadata_obj,
+            f"/{project.organisation_id}/{project_id}/submissions.meta.json",
         )
 
         # Update background task status to COMPLETED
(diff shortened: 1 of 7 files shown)


spwoodcock:
Will look into it 👍

We could recreate the submissions zip on a schedule, based on the last activity in the project.

If a project is edited in the last 24hrs, then we could recreate the submissions say every hour.

It's not a perfect solution though, as a user may want to download their submissions immediately after mapping. 

So if the file exists in S3, how do we determine if we need to recreate it or download the existing? It's a tricky problem, needs more thought.

spwoodcock:
Both good ideas, thanks @nrjadkry !

If we make the S3 file cloud optimised (a sozipfile), then it would enable option 2 (without downloading the whole file and checking), which would have the best UX.

I wonder how we can best determine the submission count from the submission files.
Perhaps we add a metadata file into the zip we generate.

Option 1 is a good fallback, displaying to the user:
- Time of last submission
- Time of last submission export / file generation
- Button to download
- Button to regenerate submission file (if it's outdated and then need it immediately).

nrjadkry:
ODK Provide the submission count as well the last submission date for a form from this api
`/v1/projects?forms=true`

We just need to process extra for the submission count and date for the whole project.
We have already implemented this in fmtm as well 

submission_count

Shall we continue with the option 1. What do you say?

spwoodcock:
Yes I think so!

The only last thing to consider is making the data available in a cloud optimised format.
I haven't looked too much into the submission code yet, so can't say what format would be best.
You created it as a geojson, but maybe we could create a flatgeobuf instead?

I am currently implementing this for the data extract too, so to makes sense to use the same formats (although geoparquet may be more suited to this data).

I will test adding this tonight 🤞 

The advantage would be that we can read metadata, without reading the entire file.
I assume the submission JSON could be many 10's of MB, up to 100's for large projects.
If we can add the submission count to the flatgeobuf **headers**, then we can access that data very easily.

spwoodcock:
I just rebased to solve conflicts.

The implementation of the zip creation is great 👍 

Three things to note:
- There is already a function `submission_crud.download_submission_for_project`. Do we need to compare what the difference is between using `get_all_submissions` and decide on only one?
- As part of the S3 zipfile upload, I think we should also add a file called `metadata.json` in the zip root. This file can contain a key like `submissionCount`, so we can use this info to compare with the submission count available on ODK Central. If the submission count is up to date, then we redirect to the S3 zip. If the count is not up to date, then we regenerate the zip.
- The logic does not need it's own endpoint `update_submission_cache`. The flow I described above could be implemented all within the `download` endpoint.

nrjadkry:
Thanks @spwoodcock 
1. `submission_crud.download_submission_for_project` gives the csv submission for each tasks and zipped them in a single file. `get_all_submissions` gives submission in a json format. This is the reason why I used `get_all_submissions`.  We might need to store csv submissions as well in a cache.
2. We might need `metadata.json` . I will add this file too in s3. What do you think should be in the metadata apart from submission count, last submission date.
3.  What do you think will be the workflow to download submission if we add `update_submission_cache` function inside of the `download` endpoint? If the submission count is same, it won't have problem, since we can download directly from the cache, but if the submission count is different, updating the cache will take place in background task.

src/backend/app/submission/submission_crud.py
```diff
@@ -436,6 +443,99 @@ def extract_files(zip_file_path):
...
+        file_obj = BytesIO(json.dumps(submission).encode())
+
+        # Metadata
+        # ODK Credentials
+        odk_credentials = project_schemas.ODKCentral(
+            odk_central_url=project.odk_central_url,
+            odk_central_user=project.odk_central_user,
+            odk_central_password=project.odk_central_password,
+        )
+        from sqlalchemy.sql import text
```
spwoodcock, line 466: Try not to put imports directly in functions.
They should always be at the top of the file, except some specific scenarios with optional dependencies.

spwoodcock:
1. Thanks for clarifying - I renamed:
get_all_submissions --> get_all_submissions_json
download_submission_for_project --> gather_all_submission_csvs

2. After considering, do you think we actually need the submission count? Can we just extract the last timestamp and use that? I updated to do that, but if you think it's wrong, then feel free to revert the commit: `git pull && git reset --soft HEAD~1`

3. I actually did something similar in the unmerged import/export endpoints. I'm sure this could be modified to do similar:

```python
@router.get("/{project_id}/export")
async def export_project(
    project_id: int,
    background_tasks: BackgroundTasks,
    task_id: Optional[str] = None,
    with_odk: bool = False,
):
    """Export a project to a zip, for import into FMTM.

    Args:
        task_id(str): Query the background task to see if complete.
        with_odk(bool): Export the ODK associated ODK Central project.
    """
    # Return existing export if complete
    if task_id:
        # Get the backgrund task status
        task_status, task_message = await project_crud.get_background_task_status(
            task_id, db
        )

        if task_status != 4:
            return project_schemas.BackgroundTaskStatus(
                status=task_status.name,
                message=task_message or "",
                # progress=some_func_to_get_progress,
            )

        bucket_root = f"{settings.S3_DOWNLOAD_ROOT}/{settings.S3_BUCKET_NAME}"
        return RedirectResponse(f"{bucket_root}/{project_id}/export.zip")

    # logic from update_submission_in_s3 here
```

> Note: important to use settings.S3_DOWNLOAD_ROOT is may be different from S3_ENDPOINT between dev/prod.

spwoodcock:
Just one minor thing - the whole idea with sozipfile is that the file is 'cloud optimised'.
You can read a subset of the data without reading in the whole file - perfect for reading only the metadata.json.

Example (in theory this should work):
```python
url = "s3 url for zipfile"
with zipfile.ZipFile(url, mode="r") as myzip:
	metadata_index = myzip.getinfo("metadata.json")
	with myzip.open(metadata_index) as file:
		metadata_json = file.read()
```

nrjadkry:
```
url = "s3 url for zipfile"
with zipfile.ZipFile(url, mode="r") as myzip:
    ....
```

This can't find the file
I have used get_object_from_s3 function instead, but I think this function will return all the files from this zipped file.

spwoodcock:
Ah ok, that makes sense - I think to read a sozip via RANGE you need to use GDAL vsicurl drivers.

Reading the entire zip is fine for now - we can think of alternatives later on 👍

spwoodcock:
Sorry to mess you around @nrjadkry - just a thought:
- If we need to download the entire zip to view the metadata, then it's not that useful.
- Perhaps we store the submission in the zip, but put the metadata outside it for now?

Something like this?
s3_root/org_id/project_id/submissions.zip
s3_root/org_id/project_id/submissions.meta.json

To make the endpoint fast, we need to download and check the metadata file first, do the timestamp comparison, then either redirect or generate the zip.

spwoodcock (approved):

PR #1042, opened 2023-12-12
delete projects api made authenticated

nrjadkry:
Delete project API is public as of now. I have made this endpoint authenticated. 
Now, we need to pass the access-token as well. This adds a small level of security to this endpoint.
 
## Todo
We will have to check the user roles and make this endpoint  accessible only by the admin user.


spwoodcock (approved):
Probably a good idea for gradual introduction 👍

PR #1044, opened 2023-12-14
hotfix: calculate max last_submission only if valid_datetimes present

nrjadkry:


spwoodcock:
You might have noticed the pre-commit.ci commits that are being added to your PRs.

To reduce them, you can install the pre-commit hooks locally to help with linting and formatting code. In the repo:

```bash
# global package install (can use PDM instead)
pip install pre-commit
pre-commit install
```

Issue #889, opened 2023-10-10
Allow downloading data extract from FMTM and support upload of the same data extract back to FMTM


susmina94:
Problem:
I cannot view the data extract generated by FMTM. I would like to see if there is any standard format for the data extract so that I can custom upload or edit the extract provided by FMTM.

Solution:
1. I want to download the data extract generated by FMTM.
2.  I want to update and modify the data and re-upload the data extract back to FMTM before generating QR code

spwoodcock:
I hope not much work has been done implementing these endpoints 🙏 
I think it's mostly been superseded by https://github.com/hotosm/fmtm/issues/934

So to address your questions:

### 1) Download the data extract generated by FMTM

- Once https://github.com/hotosm/fmtm/issues/934 is implemented then this should be trivial.
- The data extract will be available in flatgeobuf format on the raw-data-api S3 bucket (with a URL that won't change.
- We can add a link to the S3 URL in our database, which is returned when the user requests an extract download.
- A user clicks download and the frontend just redirects to the returned URL to download.
- Note: we could possibly also specify to download only a  **subset** of the data (i.e. with a bbox specified).

### 2 Update and modify the data and re-upload

- If the data extract isn't exactly how we would like, we can go back and tweak the params and regenerate  the extract.
- If it's still not good enough using the filters on raw-data-api, then we can give the user the option to upload a custom data extract (we need this functionality anyway for some users).
  - We should have a small endpoint that accepts this new extract, then uploads it to **our** S3 bucket, also inserting the URL into the database.
  - The uploaded data extract would have to be parsed and ideally converted to flatgeobuf for efficient access.

PR #1047, opened 2023-12-16
Use flatgeobuf data extracts

### LINKED ISSUE #381: Replace geojson features with flatgeobuf
@robsavoye We are thinking about switching from geojson to vectortile so that it can handle large size of map layers be it buildings or the project or task boundaries. What do you think about this?

**EDIT** see discussions below. We should use vector tiles for baselayers, but the boundary overlays should be a vector format such as flatgeobuf.

### LINKED ISSUE #934: Use data extracts hosted in raw-data-api

spwoodcock:
Fixes #934 and addresses part of #381 and #889

### Updates

Custom data extract:
- If user uploads a flatgeobuf, this is deserialised as geojson for display on the map. The file is uploaded to our S3.
- If user uploads geojson, this is converted to flatgeobuf on the backend and file uploaded to our S3.
- In either case, a URL to the `.fgb` is stored in `public.projects.data_extract_type` (abusing this param for now, but it should be migrated to `public.projects.data_extract_url` eventually).

> Custom data extracts are stored in the FMTM S3 bucket and managed by us.

OSM data extract:
- When the user continues from the 'Data Extract' page to the 'Split Tasks' page, the endpoint is called to generate a new data extract in the background.
  - This same endpoint can also accept a `project_id` param to return an already existing data extract URL.
- The data extract is generated in flatgeobuf format and stored in the S3 bucket that is part of `raw-data-api`. The URL to this extract is stored in our database.
- The data extract is converted to GeoJSON file during the project creation and passed in as a param to the task splitting `/task_split` endpoint (eventually the splitting algorithm may accept flatgeobuf instead).

> OSM data extracts are stored in the raw-data-api S3 bucket and managed by that service.

### Checks

- I have given this quite a few tests with different scenarios, but please thoroughly test this code @nrjadkry and @varun2948 before we merge.
- The frontend implementation is a bit messy and could probably do with cleaning up.
- The backend tests need to also be updated accordingly - should be pretty simple.

### Future updates

As a consequence of this big change, we need to update a few things.

Frontend:
- Improve the error handling and user feedback during this flow.
- We need to block the user generating splits using the 'task splitting algorithm' until the data extract generation completes.
- This data extract should be loaded in OpenLayers directly as flatgeobuf format throughout the frontend. [There are helpers for this](https://github.com/flatgeobuf/flatgeobuf/blob/master/examples/openlayers/filtered.html). We can even filter the data returned based on the viewport, or only return the extract data for a given task area.
- References to 'buildings' and 'lines' uploaded extracts should be consolidated into one reference (dataExtractFile), as the type of geometries can be determined on the backend instead. 
  - On the project creation Data Extract page we can also remove the radio to select the type of extract. The user can just upload an extract and we handle the rest.
  - The extracts are stored externally and no longer in our database, so this simplifies code on both backend and frontend.

Backend:
- To do this quickly, I ended up reusing an endpoint that uses `raw-data-api` directly.
- We have a partial implementation in place to do this via `osm-rawdata` instead: project_crud.get_data_extract_from_osm_rawdata.
- We need to update osm-rawdata to support the `bind_zip=False` param, then use the package instead.

### Note

We also need to consider regeneration of the extract, as it lasts 90 days.

This should be enough time for most projects, however if the user calls the `get_data_extract` endpoint and the extract is missing (deleted) a new file should be generated (already implemented). 

We just need to make sure this endpoint is called on the project detail page, to return the flatgeobuf URL and display it on the map.

### DIFF
--- a/src/backend/app/db/postgis_utils.py
+++ b/src/backend/app/db/postgis_utils.py
@@ -66,7 +66,7 @@ def get_centroid(geometry: Geometry, properties: str = {}, id: int = None):
     return {}
 
 
-def geojson_to_flatgeobuf(db: Session, geojson: FeatureCollection) -> bytes:
+async def geojson_to_flatgeobuf(db: Session, geojson: FeatureCollection) -> bytes:
     """From a given FeatureCollection, return a memory flatgeobuf obj.
 
     Args:
@@ -101,7 +101,13 @@ def geojson_to_flatgeobuf(db: Session, geojson: FeatureCollection) -> bytes:
     # Run the SQL
     result = db.execute(text(sql))
     # Get a memoryview object, then extract to Bytes
-    flatgeobuf = result.fetchone()[0].tobytes()
+    flatgeobuf = result.fetchone()[0]
+
     # Cleanup table
     db.execute(text("DROP TABLE IF EXISTS public.temp_features CASCADE;"))
-    return flatgeobuf
+
+    if flatgeobuf:
+        return flatgeobuf.tobytes()
+
+    # Nothing returned (either no features passed, or failed)
+    return None
--- a/src/backend/app/db/postgis_utils.py
+++ b/src/backend/app/db/postgis_utils.py
@@ -66,8 +66,16 @@ def get_centroid(geometry: Geometry, properties: str = {}, id: int = None):
     return {}
 
 
-def geojson_to_flatgeobuf(db: Session, geojson: FeatureCollection):
-    """From a given FeatureCollection, return a memory flatgeobuf obj."""
+def geojson_to_flatgeobuf(db: Session, geojson: FeatureCollection) -> bytes:
+    """From a given FeatureCollection, return a memory flatgeobuf obj.
+
+    Args:
+        db (Session): SQLAlchemy db session.
+        geojson (FeatureCollection): a geojson.FeatureCollection object.
+
+    Returns:
+        flatgeobuf (bytes): a Python bytes representation of a flatgeobuf file.
+    """
     sql = f"""
         DROP TABLE IF EXISTS public.temp_features CASCADE;
 
--- a/src/backend/app/organization/organization_crud.py
+++ b/src/backend/app/organization/organization_crud.py
@@ -127,7 +127,8 @@ async def create_organization(
         db.refresh(db_organization)
 
         # Update the logo field in the database with the correct path
-        db_organization.logo = await upload_logo_to_s3(db_organization, logo)
+        if logo:
+            db_organization.logo = await upload_logo_to_s3(db_organization, logo)
         db.commit()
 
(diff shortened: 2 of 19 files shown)


nrjadkry:
@spwoodcock I have pushed minor fixes here.
I have tested creating project by drawing an aoi. 
Other than that, I think this is good to go.

nrjadkry:
@spwoodcock , I can see that you have used get_data_extracts api for downloading the data extracts. But, I think we need to use osm-rawdata in this too since we will need to extract different data according to the category. Is there any reason you have used raw-data-api query directly?

spwoodcock:
I won't have time to fix the broken test or update to use osm-rawdata before the new year, so this will get merged in Jan (unless you are able to work on those @nrjadkry ) 👍

spwoodcock:
Note: modify to address #1056

PR #1051, opened 2023-12-19
Endpoint for submission page, return submission count by date

Sujanadh:
The API fetches submissions from S3, utilizing the `project_id` to construct the relevant paths. Submissions are then processed, and counts are determined based on the specified time range `days`.


spwoodcock (commented):
See comment

Sujanadh:
Okay. But one thing to consider is that submissions might not be updated in s3. In that case , what should we do? @spwoodcock

spwoodcock:
Handle it in the same way as the data extract download.

If project id is passed, check for existing submission. 

If one doesn't exist then create it. 

If one exists then we also check to see if its up to date for submissions.

PR #1053, opened 2023-12-20
Add image to user profile

Sujanadh:
# Updates

1. New DbUser field `profile_img` is created to store logged in user 's profile image.
2. Updated auth_routes so as to save it in the database. Here, i have added portion of code to save image in existing users(already logged-in users) too which might be removed later.
3. Created a function `update_task_history` to populate `task_history` with `username`, `status`(extracted from the action_text to keep track of task status history) and `profile_img` .

### DIFF
--- a/src/backend/app/auth/auth_routes.py
+++ b/src/backend/app/auth/auth_routes.py
@@ -23,10 +23,10 @@ from fastapi.responses import JSONResponse
 from loguru import logger as log
 from sqlalchemy.orm import Session
 
-from ..db import database
-from ..db.db_models import DbUser
-from ..users import user_crud
-from .osm import AuthUser, init_osm_auth, login_required
+from app.db import database
+from app.db.db_models import DbUser
+from app.users import user_crud
+from app.auth.osm import AuthUser, init_osm_auth, login_required
 
 router = APIRouter(
     prefix="/auth",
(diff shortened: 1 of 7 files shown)


spwoodcock (changes requested):
Do we need to store the profile picture in our db when it's stored in OSM anyway? Was it something requested by someone?

When the frontend makes the call to the /me endpoint the profile picture URL is returned & should be stored in the frontend state.

spwoodcock:
Also, a backend test seems to be broken by this change

Sujanadh:
Yes, you are right. Frontend makes the call to the `/me` endpoint to get the profile picture of current logged-in user only. But, to keep the track of history of user who updated the submission before we need to save their profile picture.

spwoodcock:
Got you - thanks for clarifying. 

I assume you are working on a submission dashboard with profile pictures included 👍

The picture is an extra piece of user data we store, if we need to declare a data protection policy in the future.

spwoodcock:
Could you add a revert migration please & I will merge 😄

Sujanadh:
Sure.

Sujanadh:
Previously i had merged feat-submission-page branch to this branch, to use same code to check. I mistakely pushed changes along with the merged contents of branch feat-submission-page. That's why i reset the changes and pushed the migration change. :smiley:

spwoodcock (approved):

spwoodcock:
Thanks @Sujanadh , the revert looks good, but if you see the folder structure for the migrations it would be good to be consistent with the others. There is a revert folder for migration reversions (we need these for users that accidentally update, but may need to go back a version)

Sujanadh:
Thanks  :+1: @spwoodcock  for the guidance.

PR #1061, opened 2023-12-26
Check coordinate system of project AOI

Sujanadh:
# Updates

1. Previously, the coordinate system of the uploaded AOI was verified on the frontend. However, it is now preferable to perform this check on the backend. Consequently, a new API has been created specifically for validating the coordinate system of the uploaded GeoJSON.
2. Subsequently, it is recommended to invoke this API immediately after uploading the GeoJSON.

### DIFF
--- a/src/backend/app/projects/project_routes.py
+++ b/src/backend/app/projects/project_routes.py
@@ -1256,8 +1256,9 @@ async def check_coordinate_system(boundary: UploadFile = File(...)):
     boundary = json.loads(content)
 
     # Validating Coordinate Reference System
-    if check_crs(boundary):
+    try:
+        check_crs(boundary)
         return {"message": "Valid boundary file"}
-    else:
+    except Exception as e:
         return {"message":"Unsupported coordinate system, it is recommended to use a "
         "GeoJSON file in WGS84(EPSG 4326) standard."}
--- a/src/backend/app/projects/project_routes.py
+++ b/src/backend/app/projects/project_routes.py
@@ -1234,3 +1234,30 @@ async def project_dashboard(
     """
 
     return await project_crud.get_dashboard_detail(project_id, db)
+
+
+@router.post("/check_crs")
+async def check_coordinate_system(boundary: UploadFile = File(...)):
+    """Check coordinate system of uploaded file.
+
+    Args:
+        boundary (UploadFile): File containing the boundary.
+
+    Returns:
+        dict: Dictionary containing the result of the check.
+    """
+
+    boundary_type = boundary.content_type.split("/")
+    if len(boundary_type) != 2 or boundary_type[0] != "application" or boundary_type[1] != "geo+json":
+        raise HTTPException(
+            status_code=400, detail="Invalid boundary file. Must be GeoJSON."
+        )
+    content = await boundary.read()
+    boundary = json.loads(content)
+
+    # Validating Coordinate Reference System
+    if check_crs(boundary):
+        return {"message": "Valid boundary file"}
+    else:
+        return {"message":"Unsupported coordinate system, it is recommended to use a "
+        "GeoJSON file in WGS84(EPSG 4326) standard."}


spwoodcock:
Do we need a new endpoint for this? The project creation already makes quite a lot of calls, adding load to the server.

We added CRS checking before as a low priority item. GeoJSON in CRS other than EPSG:4326 is very uncommon.

And how come CRS checking is preferred on the backend? I would actually argue that checking via OpenLayers and proj on the frontend is preferable. We don't even need reprojection, just a simple check. 

@susmina94 do you have thoughts?

Sujanadh:
I agree. Have you tried checking via openlayers or proj in frontend? @NSUWAL123

NSUWAL123:
Currently, only polygon is handled via frontend. I will try to handle multipolygon also.

Sujanadh:
For now, geojson was checked using manual procedure that is whether the coordinate of polygon lies between -90 to +90 or -180 to +180. Checking like this will probably slows down. That's why i created endpoint to avoid that in frontend.

spwoodcock:
If we are making an api call anyway, then it's ok to check crs in the backend (as we do currently when we upload geojson files).

I would be surprised if the frontend struggles to calculate if a polygon is inside bounds though. If the polygon has thousands of vertices to check it would make sense. But the backend wouldn't be much quicker unless the calculation is vectorised.

It it's an issue and the user experience is too bad then we should reconsider doing the crs checks.

Sujanadh:
okay, lets handle it on the frontend side. @NSUWAL123  will do that, he will check every possible type of EPSG: 4326 for both polygon and multipolygon which might be as follows:
1. `"urn:ogc:def:crs:OGC:1.3:CRS84"`,
2. `"urn:ogc:def:crs:EPSG::4326"`,
Or same like before, checking coordinates (lat between -90 to 90 and lon between -180 to 180) but if we are taking it seriously then, it only wont be enough since other crs like `NAD 83`, `EPSG: 3395` and so on also have similar range for lat lon.

And there is not point doing check_crs in every function where geojson is uploaded. So, will refactor those.

spwoodcock:
Thanks for the feedback @Sujanadh 👍

I agree it's difficult to accurately assess the CRS without writing lots of code.

For cases when the user has immediate feedback / the geom is displayed on the map, viewing that the geom is incorrectly placed is validation enough.

We only really need CRS checking where passing an uploaded geojson to a backend function might break it / make it behave unexpectedly.

A geojson with the crs key isn't overly common, but checking the coord bounds is a good rough check for EPSG:4326 (although not perfect like you say).

spwoodcock:
I'm wondering if OpenLayers has some functions already written to handle this too

varun2948:
@spwoodcock openlayers support most of the geojson types natively so right now backend needs to do some handling with the provided geojson right ?

spwoodcock:
OpenLayers does support multiple coordinate systems, but by default can only load EPSG:3857 and EPSG:4326. For others, the proj code needs to be loaded, e.g.

```js
import proj4 from 'proj4';
import {register} from 'ol/proj/proj4.js';
proj4.defs('EPSG:21781',
  '+proj=somerc +lat_0=46.95240555555556 +lon_0=7.439583333333333 +k_0=1 ' +
  '+x_0=600000 +y_0=200000 +ellps=bessel ' +
  '+towgs84=660.077,13.551,369.344,2.484,1.783,2.939,5.66 +units=m +no_defs');
register(proj4);
const swissProjection = getProjection('EPSG:21781');
```

So you really need a good idea of what coordinate systems will be provided, to have support for them.
It's difficult to cover all possible coordinate systems may provide, and honestly it quite a niche thing to require in our application (more for scientific / research uses).

As of now we just need EPSG:4326 support.

So the idea here is just a simple check to see if the coordinates are in bounds: -180.0 -90.0
180.0 90.0.

I think the frontend could handle that right?
(and provide an error if the geojson is not in the bounds)

varun2948:
@spwoodcock yeah sure we can do that, Let's do that for now instead of requesting backend for the check.

spwoodcock:
Closing, handled on the frontend: https://github.com/hotosm/fmtm/pull/1112

PR #1071, opened 2023-12-27
Separate task history end point

Sujanadh:
# Updates

I tried updating task history in task-list. It was working fine. But to apply sorting and filtering operations for history list, it will be quite hectic since it is inside list of nested json. Using separate endpoint for only task history seems good idea and makes easy to apply filter.
This PR address this feature.
And yes, there are so many schemas with similar name, I will restructure those and use only required schemas.

### DIFF
--- a/src/backend/app/tasks/tasks_schemas.py
+++ b/src/backend/app/tasks/tasks_schemas.py
@@ -22,7 +22,7 @@ from typing import Any, List, Optional
 
 from geojson_pydantic import Feature as GeojsonFeature
 from loguru import logger as log
-from pydantic import BaseModel, ConfigDict, Field, ValidationInfo
+from pydantic import BaseModel, ConfigDict, Field, ValidationInfo, computed_field
 from pydantic.functional_serializers import field_serializer
 from pydantic.functional_validators import field_validator
 
@@ -96,7 +96,9 @@ class Task(BaseModel):
 
     @field_validator("outline_centroid", mode="before")
     @classmethod
-    def get_centroid_from_outline(cls, value: Any, info: ValidationInfo) -> str:
+    def get_centroid_from_outline(
+        cls, value: Any, info: ValidationInfo
+    ) -> Optional[str]:
         """Get outline_centroid from Shapely geom."""
         if outline := info.data.get("outline"):
             properties = {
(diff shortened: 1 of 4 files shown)


Sujanadh:
## Response 
```
[
  {
    "task_id": 605,
    "action_text": "Status changed from LOCKED_FOR_MAPPING to MAPPED by: Sujan Adhikari",
    "action_date": "2023-12-20T07:44:48.339376",
    "status": "MAPPED",
    "username": "Sujan Adhikari",
    "profile_img": "https://www.openstreetmap.org/rails/active_storage/representations/redirect/eyJfcmFpbHMiOnsiZGF0YSI6MzI4NTE3MTksInB1ciI6ImJsb2JfaWQifX0=--6b73d47fac28854f227f3d0ecd56f2f6554c8709/eyJfcmFpbHMiOnsiZGF0YSI6eyJmb3JtYXQiOiJwbmciLCJyZXNpemVfdG9fbGltaXQiOlsxMDAsMTAwXX0sInB1ciI6InZhcmlhdGlvbiJ9fQ==--473f0190a9e741bb16565db85fe650d7b0a9ee69/sujan.webp"
  },
  {
    "task_id": 605,
    "action_text": "Status changed from READY to LOCKED_FOR_MAPPING by: Sujan Adhikari",
    "action_date": "2023-12-20T07:44:23.825255",
    "status": "LOCKED_FOR_MAPPING",
    "username": "Sujan Adhikari",
    "profile_img": "https://www.openstreetmap.org/rails/active_storage/representations/redirect/eyJfcmFpbHMiOnsiZGF0YSI6MzI4NTE3MTksInB1ciI6ImJsb2JfaWQifX0=--6b73d47fac28854f227f3d0ecd56f2f6554c8709/eyJfcmFpbHMiOnsiZGF0YSI6eyJmb3JtYXQiOiJwbmciLCJyZXNpemVfdG9fbGltaXQiOlsxMDAsMTAwXX0sInB1ciI6InZhcmlhdGlvbiJ9fQ==--473f0190a9e741bb16565db85fe650d7b0a9ee69/sujan.webp"
  }
]
```

spwoodcock:
Thanks for the example response, very useful 😄

What immediately stands out is that the profile image URL can be quite long, but is replicated in each item.

Not a big deal if the history is only a few items. But if there are 100s of entries, then it's a bit inefficient to send all that extra data in the json payload. I wonder if there is a way to optimise.

src/backend/app/tasks/tasks_schemas.py
```diff
@@ -135,3 +137,46 @@ class ReadTask(Task):
...
+    )
+
+    # Excluded
+    user: Any = Field(exclude=True)
+
+    task_id: int
+    action_text: str
+    action_date: datetime
+
+    @computed_field
```
spwoodcock, line 156: I moved all the logic to the pydantic model.

Note we can use `@computed_field` decorators to generated computed values.
We can also use this to generate values based on database relationships.
To do this we need the model_config `from_attributes=True`:
```python
    model_config = ConfigDict(
        from_attributes=True,
    )
```

spwoodcock (approved):
See my comment on usage of `@computed_fields`.

I'm not sure what this endpoint is needed for, but I will merge it & it can be updated as needed.

Sujanadh:
i couldn't get back to it as i got busy on other things :smiley:

spwoodcock:
No worries! The code is merged & we can built on it at another point if needed 👍

PR #1076, opened 2024-01-02
Serialization for task_status and created project dashboard fields

Sujanadh:
# Updates
Previously, the `created` field in the `ProjectDashboard` model was of type `datetime` and `task_status` field in the `TaskBase` model was of type enum class `TaskStatus`, causing a validation issue. 
This pull request resolves this issue.

### DIFF
--- a/src/backend/app/projects/project_schemas.py
+++ b/src/backend/app/projects/project_schemas.py
@@ -158,7 +158,7 @@ class ProjectDashboard(BaseModel):
     total_tasks: int
     total_submission: int
     total_contributors: int
-    created: datetime
+    created: str
     last_active: Optional[str] = None
 
     @field_validator("created", mode="before")
--- a/src/backend/app/projects/project_schemas.py
+++ b/src/backend/app/projects/project_schemas.py
@@ -161,7 +161,7 @@ class ProjectDashboard(BaseModel):
     total_contributors: Optional[int] = None
     last_active: Optional[str] = None
 
-    @field_validator("created", mode="before")
+    @field_validator("created", mode="after")
     def get_created(cls, value, values):
         date = value.strftime("%d %b %Y")
         return date
--- a/src/backend/app/projects/project_schemas.py
+++ b/src/backend/app/projects/project_schemas.py
@@ -18,11 +18,12 @@
 
 import uuid
 from datetime import datetime
-from dateutil import parser
 from typing import List, Optional
 
+from dateutil import parser
 from geojson_pydantic import Feature as GeojsonFeature
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, ValidationInfo
+from pydantic.functional_serializers import field_serializer
 
 from app.db import db_models
 from app.models.enums import ProjectPriority, ProjectStatus, TaskSplitType
@@ -141,6 +142,7 @@ class ProjectBase(BaseModel):
 class ProjectOut(ProjectBase):
     project_uuid: uuid.UUID = uuid.uuid4()
 
+
 class ReadProject(ProjectBase):
     project_uuid: uuid.UUID = uuid.uuid4()
     location_str: str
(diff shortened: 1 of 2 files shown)


spwoodcock:
The enum is used for validation to make sure the provided value is one of the available options.

I think we should fix the validation issue another way.

The frontend should be passing an integer - is this the case?

For the datetime, the provided string could be cast to datetime within a field_validator.

Sujanadh:
Yes frontend passes an integer. What happened here is, from the database `TaskStatus.READY` enum type is retrieved and in order to return its name i processed the value in the function `get_enum_name` which returns the name of respective value. Since no other name than from TaskStatus will be returned so there wont be issue of getting different value other than from the enum.

Why i changed it to str is because TaskStatus expects int value something like 0,1,2  to be returned from the function but i changed the int to its name , thats why it caused validation error.

I think i could do something like this,
`task_status: Union[TaskStatus, str]`

spwoodcock:
Thanks for the context 😃

You could solve as you suggested.

Or you could have a field validator do the conversion, putting the `get_enum_name` function content in the validator instead.
Validators can accept data in different forms but make sure the value type in the model is always the same.

Sujanadh:
For created field, it is expected to be returned as
`"created": "28 Dec 2023"`
if `Datetime` field type is used for `created`, it will raise error as `"created": "28 Dec 2023"` is not valid datetime for the field type `Datetime`.

spwoodcock:
Looks like pydantic already has handling for string values --> datetime validation: https://docs.pydantic.dev/2.0/usage/types/datetime/#validation-of-datetime-types

The datetime is in the wrong format from the frontend.
`28 Dec 2023` should be `2023-12-28` and validation *should* work.

spwoodcock:
Actually I misunderstood this - we are looking at a GET endpoint where we need to pass the datetime to the frontend, not a POST where we need to parse a value provided by the frontend.

In this case we can use `@field_serializer` (used for output) instead of `field_validator` (used for input).

However, I am wondering what the purpose of the custom serialization is, when a datetime field is already serialized to string automatically by Pydantic:
image

Is it possible for the frontend to parse the string in format `2023-12-15T17:55:51.032322` and display as required?

spwoodcock:
I made a minor update to use `field_serializer`.

I think that this should be removed

```
    @field_serializer("created")
    def get_created(self, value: datetime, info: ValidationInfo):
        date = value.strftime("%d %b %Y")
        return date
```

And the datetime string parsed on the frontend correctly.

But if is absolutely necessary the frontend receives a string in format `28 Dec 2023`, then this can be merged as it is.

What is the error your were receiving originally?

spwoodcock:
With the following code I am getting errors on the project detail page (http://fmtm.localhost:7050/newproject_details/MTAwMDAx):

image

image

Removing the validators fixes it. I removed them for now.

The datetime should be serialized as a a datetime string, and the TaskStatus serialized as an integer (the frontend then converts it to the required string value using the enum definitions in the frontend).

Sujanadh:
The issue was similar like mentioned above. Date like `28 dec 2023` was requested by the frontend as per the new UI desgin. 
For the TaskStatus if frontend uses int value and get the respective status name then we can do as you have suggested.

spwoodcock:
Dates should always be passed as datetime strings, then cast as Date objects in the frontend.

Values like 2021-01-21 are unambiguous and consistent. They can be read and then formatted as required.

Casting a string representation is more difficult.

Example:

```js
const backendDatetime = '2021-02-23T15:01:51.444422';
const dateObject = new Date(backendDatetime);
const formattedDate = new Intl.DateTimeFormat('en-US', { day: 'numeric', month: 'short', year: 'numeric' }).format(dateObject);

console.log(formattedDate); // Output: 23 Feb 2021
```

@NSUWAL123 
@Sujanadh

Sujanadh:
okay :+1:

PR #1080, opened 2024-01-04
Pre-generated mbtile files for each task area in a project

### LINKED ISSUE #890: Allow downloading of mbtiles per task

nrjadkry:
Fixes #890 

1. I have uploaded the mbtiles we create in s3 bucket in this path. `/orgid/projectid/basemap.mbtiles`
2. Generated task wise mbtiles from the same file path so that it uses the cache from the tiles generated when creating the mbtiles for project. and uploaded the resulting mbtile into s3 bucket.

@spwoodcock has suggested to use pmtiles rather than mbtiles when generating the project wise mbtiles. I was wondering if these pmtiles generated works for odk collect? And if not, do we need project wise mbtiles too.


robsavoye:
I don't believe pmtiles will work for ODK Collect.

spwoodcock:
No they won't.

The idea is:
- PMTiles for project area to efficiently load everything on the web UI (when online).
- Pre-generated MBTile archives for each task area, used when mapping (e.g. in ODK Collect).

nrjadkry:
I have revised the output format for task-wise tiles to use MBTiles. Would it be okay to initiate this endpoint during the project generation process? The endpoint already generates tiles in the background. What do you think @spwoodcock?

src/backend/app/projects/project_crud.py
```diff
@@ -2211,6 +2211,76 @@ def get_project_tiles(
         )
         log.info(f"Basemap created for project ID {project_id}: {outfile}")
 
+
+        get_project_sync = async_to_sync(get_project)
+        project = get_project_sync(db, project_id)
+
+        from app.s3 import add_file_to_bucket
```
spwoodcock, line 2218: Try not to import inline like this, it goes against the PEP8 style guide. Always import at the top of the file 👍

spwoodcock:
Code is good, but this is a larger PR than the initial code suggests and requires modifying the backend and frontend together.

The code as it is will break the functionality of generating individual tile archives.

I will push some changes to keep the existing functionality, while also running in the background on project creation as mentioned.

nrjadkry:
@spwoodcock @varun2948 @NSUWAL123 where are we with making changes in the frontend for this PR.

spwoodcock:
I'll look at fixing conflicts and working on a basic frontend implementation in the coming days 👍

spwoodcock:
Closing for now as this is too far out of sync.

We can return to it in future and use this as inspiration.

PR #1091, opened 2024-01-09
Cookie based authentication

spwoodcock:
## What type of PR is this? (check all applicable)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Issue

As discussed in https://github.com/hotosm/fmtm/discussions/1083

## Describe this PR

- We had authentication in place to return the OSM user details.
- However the **authentication** details did not persist, so implementing **authorisation** would require reacquiring **authentication** every time.
- This PR adds support for storing the auth token (OSM JWT token) in a secure cookie.
- Cookie details: samesite:lax (for CSRF), httpOnly (for XSS), secure (for MIM).

## Screenshots

image

## Implications

- For testing the api using auth, we no longer have to pass the access_token manually.
- If you login once via the localhost frontend, then a cookie will be saved in your browser.
- This cookie will be automatically extracted when testing endpoints on the docs page: http://api.fmtm.localhost:7050/docs making development a lot smoother.
fyi @nrjadkry @Sujanadh 

## Checklist before requesting a review

- 📖 Read the FMTM Contributing Guide: <https://github.com/hotosm/fmtm/blob/main/CONTRIBUTING.md>
- 📖 Read the FMTM Code of Conduct: <https://github.com/hotosm/fmtm/blob/main/CODE_OF_CONDUCT.md>
- 👷‍♀️ Create small PRs. In most cases, this will be possible.
- ✅ Provide tests for your changes.
- 📝 Use descriptive commit messages.
- 📗 Update any related documentation and include any relevant screenshots.

## [optional] What gif best describes this PR or how it makes you feel?

giphy

PR #1094, opened 2024-01-10
Add basic user role support to backend

### LINKED ISSUE #243: Implement support for roles in the backend
Currently the database and backend has no support for roles, so this will need to be added. The initial roles would be these:

Based on the experience in Dominica, I suspect we'll need the following roles:

- **Admin** (all permissions for everything on the server)
- **Group (Organisation) Admin** (all permissions within a group, so they can create, edit, process, or delete projects for their own agency, but not see or mess with projects by another group; so for example the Red Cross admin can do everything with the Red Cross' projects but not see or mess with Save the Children's projects)
- **Field Admin** (cannot create projects, but can unlock tasks unilaterally—we saw the need for this during the Dominica mapping when mappers locked tasks, didn't map them, and no one other than the mapper who locked them could force-unlock them to access the QR codes)
- **Field Mapper** (select tasks and mark them complete)
- **Validator** (select tasks, download the submissions, mark them validated or rejected)

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Issue

Fixes #243

## Describe this PR

This PR anticipates the merge of #1091 ✅ as it relies on the auth.
Also anticipates merge of #1096 ✅.

As described in discussion #1083 we are in the process of figuring out user roles.

I have implemented a basic configuration as such (this can change as needed):

```python
class UserRole(IntEnum, Enum):
    """Available roles assigned to a user site-wide in FMTM.

    Can be used for global user permissions:
        - READ_ONLY = write access blocked (i.e. banned)
        - MAPPER = default for all
        - ADMIN = super admin with access to everything
    """

    READ_ONLY = -1
    MAPPER = 0
    ADMIN = 1


class ProjectRole(IntEnum, Enum):
    """Available roles assigned to a user for a specific project.

    Invitation is required for a MAPPER to join a project.
    All roles must be assigned by someone higher in the hierarchy:
        - MAPPER = default for all (invitation required)
        - VALIDATOR = can validate the mappers output
        - FIELD_MANAGER = can invite mappers and organise people
        - ASSOCIATE_PROJECT_MANAGER = helps the project manager, cannot delete project
        - PROJECT_MANAGER = has all permissions to manage a project, including delete
        - ORGANIZATION_ADMIN = has project manager permissions for all projects in org
    """

    MAPPER = 0
    VALIDATOR = 1
    FIELD_MANAGER = 2
    ASSOCIATE_PROJECT_MANAGER = 3
    PROJECT_MANAGER = 4
    ORGANIZATION_ADMIN = 5
```

Above I have divided roles into two types.

- UserRole is one of three types and is used for site-wide permissions.
- ProjectRole is related to project specific permissions.

## Next Steps

- Implement the remaining auth levels.
- Add to endpoints as required.
  - UserRole probably won't be used much, except for granting site-wide admin access or blocking users.
  - ProjectRole endpoints will require the project_id to passed as a param (as part of the dependency injection).

Examples:

```python
@router.get("/superadmin/")
async def superadmin(
    user: AuthUser = Depends(super_admin),
):
    return user
```

```python
@router.get("/validator/")
async def validator(
    user: AuthUser = Depends(validator),
):
    return user
```

giphy

### DIFF
--- a/src/backend/app/auth/roles.py
+++ b/src/backend/app/auth/roles.py
@@ -28,8 +28,9 @@ from sqlalchemy.orm import Session
 
 from app.auth.osm import AuthUser, login_required
 from app.db.database import get_db
-from app.db.db_models import DbUser, DbUserRoles
-from app.models.enums import ProjectRole, UserRole
+from app.db.db_models import DbProject, DbUser, DbUserRoles
+from app.models.enums import HTTPStatus, ProjectRole, UserRole
+from app.projects.project_deps import get_project_by_id
 
 
 async def get_uid(user_data: AuthUser) -> int:
@@ -38,22 +39,67 @@ async def get_uid(user_data: AuthUser) -> int:
         return user_id
     else:
         log.error(f"Failed to get user id from auth object: {user_data}")
-        raise HTTPException(status_code=401, detail="Auth failed. No user id present")
+        raise HTTPException(
+            status_code=HTTPStatus.UNAUTHORIZED,
+            detail="Auth failed. No user id present",
+        )
 
 
 async def super_admin(
     db: Session = Depends(get_db),
     user_data: AuthUser = Depends(login_required),
 ) -> AuthUser:
+    """Super admin role, with access to all endpoints."""
     user_id = await get_uid(user_data)
 
     match = db.query(DbUser).filter_by(id=user_id, role=UserRole.ADMIN).first()
 
     if not match:
         log.error(f"User ID {user_id} requested an admin endpoint, but is not admin")
-        raise HTTPException(status_code=403, detail="User must be an administrator")
+        raise HTTPException(
+            status_code=HTTPStatus.FORBIDDEN, detail="User must be an administrator"
+        )
 
     return user_data
+
+
+async def org_admin(
+    project: DbProject = Depends(get_project_by_id),
+    org_id: int = None,
+    db: Session = Depends(get_db),
+    user_data: AuthUser = Depends(login_required),
+) -> AuthUser:
+    """Organization admin with full permission for projects in an organization."""
+    user_id = await get_uid(user_data)
+
+    org_admin = (
+        db.query(DbUserRoles)
+        .filter_by(user_id=user_id, role=ProjectRole.ORGANIZATION_ADMIN)
+        .first()
+    )
+
+    if not org_admin:
+        log.error(f"User ID {user_id} is not an admin for any organization")
+        raise HTTPException(
+            status_code=HTTPStatus.FORBIDDEN,
+            detail="User must be an organization admin",
+        )
+
+    matched_project = db.query(DbProject).filter_by(id=org_admin.project_id).first()
+    matched_org_id = matched_project.organisation_id
+
+    if (
+        org_id
+        and matched_org_id == org_id
+        or project
+        and matched_org_id == project.organisation_id
+    ):
+        return user_data
+
+    log.error(f"User ID {user_id} is not an organization admin for id {org_id}")
+    raise HTTPException(
+        status_code=HTTPStatus.FORBIDDEN, detail="User is not an organization admin"
+    )
 
 
 async def validator(
@@ -61,6 +107,7 @@ async def validator(
     db: Session = Depends(get_db),
     user_data: AuthUser = Depends(login_required),
 ) -> AuthUser:
+    """A validator for a specific project."""
     user_id = await get_uid(user_data)
 
     match = (
@@ -69,7 +116,9 @@ async def validator(
 
     if not match:
         log.error(f"User ID {user_id} has no access to project ID {project_id}")
-        raise HTTPException(status_code=403, detail="User has no access to project")
+        raise HTTPException(
+            status_code=HTTPStatus.FORBIDDEN, detail="User has no access to project"
+        )
 
     if match.role.value < ProjectRole.VALIDATOR.value:
         log.error(
@@ -77,7 +126,8 @@ async def validator(
             f"for project ID {project_id}"
         )
         raise HTTPException(
-            status_code=403, detail="User is not a validator for this project"
+            status_code=HTTPStatus.FORBIDDEN,
+            detail="User is not a validator for this project",
         )
 
     return user_data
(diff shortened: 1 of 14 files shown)


nrjadkry:
1. I think we need to `login_required` function for invalid access-token. As of now, it throws an error, if invalid access-token is passed.
2. You have removed organisations from user role table. We might need to have a user having an organisation level clearance, who will have all the access to every projects within that organisations. My thought is we need to save that data in the db too. What do you think?

spwoodcock:
1. Excellent find - I'll make a minor commit to handle that.
2. My thinking was that a project can only be in one organization, so the information is redundant in `user_roles`, as it is already present in the `project` table. But you are completely right, for the organization_admin role, we actually need the organization_id present in the `user_roles` table.

Either `project_id` is used or `organization_id`, so we could do one of:
- convert `project_id` --> `id` and use the field for the project or org id, depending on what the role is.
- add `organization_id` back in, and use one or the other.

Then in `roles.py` we we make `org_admin` we would still be receiving a project_id from and endpoint, so would first need filter DbProject to get the org_id, then filter DbUserRole to check if they are org admin for that org.

nrjadkry:
I think we need to add `organization_id` back in.
My guess is only super admin can create an organisation and then Organization admin role can be assigned to the users, probably by super admin only. 
It seems that, in the beginning, the organization may not have any projects. Then, only assigned organisation admin can create a project within that organisation.
 In this case, backtracking from project might not be feasible.

spwoodcock:
It's a bit of a pain to need one person (admin) to approve all organisations. 

Not a problem for organisation specific instances, but for the HOT maintained one it's extra work and may put off users if it's a slow process (we also need a mechanism for org admins to request creation too).

Could we possibly allow anyone to create a new organosation, then that person immediately becomes org admin (as the first user in the org). It's not a perfect solution, so perhaps there are other ideas too.

spwoodcock:
Perhaps part of the solution is to have an organization called `other` that is generated by default (along with the super admin being automatically generated on first start).

Then an user that wants to make a project, but does not have an organization, can get started and create one in the `other` org (the project could perhaps be moved if they request a organization be made for them too).

spwoodcock:
> 1. Excellent find - I'll make a minor commit to handle that.
> 
>     2. My thinking was that a project can only be in one organization, so the information is redundant in `user_roles`, as it is already present in the `project` table. But you are completely right, for the organization_admin role, we actually need the organization_id present in the `user_roles` table.
> 
> 
> Either `project_id` is used or `organization_id`, so we could do one of:
> 
>     * convert `project_id` --> `id` and use the field for the project or org id, depending on what the role is.
> 
>     * add `organization_id` back in, and use one or the other.
> 
> 
> Then in `roles.py` we we make `org_admin` we would still be receiving a project_id from and endpoint, so would first need filter DbProject to get the org_id, then filter DbUserRole to check if they are org admin for that org.

Thinking about this more just now. As we pass the project_id as a param, and never the org_id, we need to do a filter on DbProjects anyway.

I don't think it's necessary to add organization_id back into the model, unless an organization exists without a project (I guess that is feasible, so we need to consider that case too).

I just pushed code for org_admin that works on testing. Let me know what you think 😄

To test
```python
from app.auth.roles import org_admin
@router.get("/org_admin/")
async def org_admin(
    user: AuthUser = Depends(org_admin),
):
    return user
```

src/backend/app/auth/roles.py
```diff
@@ -0,0 +1,133 @@
...
+
+    match = db.query(DbUser).filter_by(id=user_id, role=UserRole.ADMIN).first()
+
+    if not match:
+        log.error(f"User ID {user_id} requested an admin endpoint, but is not admin")
+        raise HTTPException(
+            status_code=HTTPStatus.FORBIDDEN, detail="User must be an administrator"
+        )
+
+    return user_data
```
nrjadkry, line 63: Hi @spwoodcock 
What is the purpose of this function? 
Does this check if the user is super_admin or not? If that is the case, we might need to do something like 
`user.role == UserRole.ADMIN` after we get the user from db.
How does this work?

nrjadkry:
Can a user be a project admin in one project and be just a mapper in other. Likewise, an organisation admin in one organisation, project admin in the project of some other organisation and just a mapper for another?

If this can be true, we might need to think differently?

src/backend/app/auth/roles.py
spwoodcock: Doesn't this line do that? https://github.com/hotosm/fmtm/blob/35eaa0d1b9fab5ee769f9e751384f348a8c59421/src/backend/app/auth/roles.py#L55

I definitely could have got something wrong here though - thanks for the sanity check!

spwoodcock:
Yes they can definitely be all of the above.

Above I have divided roles into two types:

**UserRole**:
- One of three types and is used for site-wide permissions.
  - ADMIN (super admin)
  - MAPPER (most users)
  - READ_ONLY (write access blocked)
- This is used for `user.role` in the database.

**ProjectRole**:
- Related to project specific permissions.
- Stored in `user_roles` link table, with the `project_id` and `user_id` being a composite primary key.
  - This means the user can multiple roles across projects (multiple entries in the table).
  - BUT, can only have _one role within a project_.
- This is where we define the roles like org admin, project admin, validator, etc.

src/backend/app/auth/roles.py
nrjadkry: @spwoodcock my understanding is that this line filters data from user role with id=user_id and role=admin.
But, we need to do this.
```
user_role = db.query(DbUser).filter_by(id=user_id,).first().role

if user_role == UserRole.ADMIN:
    return True
else:
    return False
```

spwoodcock: They are basically the same thing.

The line further down does a check `if match`.
- If no user with the user id and role `admin` exists, then return a HTTPException.
- If a user exists and has the role `admin`, then continue to return the user details.

Does that make sense?

spwoodcock:
Should we merge this, so we can get the linting fixes in, then continue in another PR?

nrjadkry:
@spwoodcock  let's merge this. If we encounter any issues or lag, then we can obviously fix them in another PR.

PR #1141, opened 2024-01-26
hotfix: retrieve id and username directly instead of using .get

Sujanadh:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Issue
`AttributeError: 'AuthUser' object has no attribute 'get'`

## Describe this PR

In the roles.py file, there was an issue where the `get` method was incorrectly used on an 'AuthUser' object. This resulted in an AttributeError since 'AuthUser' does not have a 'get' attribute. The problem was fixed by accessing the 'id' and 'username' attribute directly. 

### DIFF
--- a/src/backend/app/auth/roles.py
+++ b/src/backend/app/auth/roles.py
@@ -38,7 +38,7 @@ from app.projects.project_deps import get_project_by_id
 
 async def get_uid(user_data: AuthUser) -> int:
     """Extract user id from returned OSM user."""
-    if user_id := user_data.get("id"):
+    if user_id := user_data.id:
         return user_id
     else:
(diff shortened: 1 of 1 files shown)


spwoodcock:
Good catch, thank you!

With DEBUG=True I was returning a user details dict, but Pyright picked up that it should be an AuthUser object in non debug versions, so I changed it (and forgot to update to remove .get)!

Type checking is handy 😄

PR #1146, opened 2024-01-30
fix: API interceptor used in getOrganizationRequest

varun2948:

- [ ] 🍕 Feature
- [X ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Describe this PR

- Fixing the cookie not going into the response header.


spwoodcock:
Did this fix the error?

It's strange to me, as the axios interceptor in `App.jsx` should add `withCredentials = true` to all axios calls automatically, so the cookie should be passing correctly.

The fix in this PR will actually add `baseURL: import.meta.env.VITE_API_URL` to the axios config, so the final URL called will probably be:
`$(import.meta.env.VITE_API_URL)/$(import.meta.env.VITE_API_URL)/organisations`
putting the `VITE_API_URL` in front twice.

varun2948:
yeah it fixed the error sam, 
I think the interceptor only works when we use the axios interceptor like i did in organization service i have been using that way for a long time and when we dont use the interceptor config and use axios directly from the new import it is making a new axios instance where the interceptor isn't working.

spwoodcock:
Adding an interceptor to axios globally should add the interceptor every time axios is imported in other modules:

```js
import axios from 'axios';

axios.interceptors.request.use(
  (config) => {
    config.withCredentials = true;
    return config;
  },
  (error) =>
    Promise.reject(error),
);

// Another file
import axios from 'axios';
axios.get(...) // Includes interceptor
```

You can also add an interceptor to an axios instance:

```js
import axios from 'axios';

export API instance = axios.create()
API.interceptors.request.use(
  (config) => {
    config.withCredentials = true;
    return config;
  },
  (error) =>
    Promise.reject(error),
);

// Another file
import axios from 'axios';
import { API } from 'some/file.js'
API.get(...) // Includes interceptor
```

The second approach is preferred, so I'm not saying the change in this PR is an issue.
But as a quick fix I added a global interceptor to all axios calls, and `withCredentials` should have been added to every request.
(I had tested it and thought it worked)
If the global interceptor doesn't work, then we can remove it & need to refactor to use the `API` axios instance everywhere.

varun2948:
The global interceptor should be working the way you did it, and one the first load it is working, it is showing the cookie getting passed to the URL but when we refresh the same page the cookie doesn't go through the request. That's some weird thing going on will maybe look once.

spwoodcock:
Ah that makes more sense!

I think it's to do with where I put the code in App.jsx - on refresh it must not trigger again 😅

PR #1171, opened 2024-02-06
Endpoints for task comments

### LINKED ISSUE #241: Add support for comments
On the page a mapper uses to get their QR code, we should add support for them to add comments.

varun2948:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Issue

- Fixes #241 

## Describe this PR
Integrated a task_comment with migration in backend with field included
id
task_id
project_id
comment_text
commented_by
created_at

Can you check this PR for task_comment @spwoodcock ?


varun2948:
yeah @spwoodcock i had worked even before the comment of rob and yours, so i will change this to the way TM is handling.

spwoodcock:
So sorry for the extra work 🙏 
I'm sure most of the code can be adapted, just without the migration.

varun2948:
no worries sam, will have it changed and yeah the code will work based on that change with little refactor

varun2948:
Hmm, yeah we can do that but one query is about the task action is, are we supposed to go with comment being enable only if the task gets locked for mapping or any other status, i mean aren't we supposed to give comment based on any status going on like i should be able to comment on the task even before it gets released for mapping, locked for mapping, locked for validation ?

spwoodcock:
Task history contains just a list of records for what happened in the task.

My thoughts were it is probably something like:

```
Task locked by user abc
Task marked complete by user abc
User abc left comment 'I tried my best with the footprints, but there were a few missing buildings in this area...'
Task marked as validated by user cba
User cba left comment 'This area will be remapped in a second round in Q4 2025'
```

So the history is just in date order and is a list of events associated with task.
I'm not certain, but I think if a user locks as task, it shouldn't prevent someone commenting.

varun2948:
@spwoodcock i have worked on this to add comment with new endpoint but do we need a delete functionality on comment aswell ? only by the one who commented it or any super_admin or higher role person could delete it ? do we need this ?

spwoodcock:
Maybe at some point we will, but for now it's ok to leave it 👍 

We can discuss in the future - either we:
- Delete the actual data.
- Create an additional task history entry for the comment deletion.

varun2948:
@spwoodcock can you review this if its good or any changes need to be done ?

spwoodcock:
Will do tomorrow!

There are some linting errors to fix before merge by the way 😉
https://results.pre-commit.ci/run/github/548162511/1707661028.mVAxA96-Twuz44rUxhNEcQ

varun2948:
Fixed linting issue

src/backend/app/db/db_models.py
```diff
@@ -350,6 +350,32 @@ class DbTaskHistory(Base):
     )
 
 
+class TaskComment(Base):
```
spwoodcock, line 353: Is this still needed?

src/backend/app/tasks/tasks_routes.py
```diff
@@ -32,6 +32,8 @@
 from app.tasks import tasks_crud, tasks_schemas
 from app.users import user_schemas
 
+from ..auth.osm import AuthUser, login_required
```
spwoodcock, line 35: Please try to always use absolute imports `app.auth.osm`

spwoodcock (approved):
Looks very nice!

I will rebase (to fix the failing test), remove the db model, and mark ready.
Merge when you are happy with it👍

nrjadkry (approved):

spwoodcock:
I just made some small updates:
- Fixed the linting.
- Improved the SQL readability with indenting & by removing the CTE from the insert query.
- Used query params to insert vars, instead of f-strings. This is important to prevent possible SQL injection risk.

spwoodcock:
Also, ignore the failing backend test for this. I need to fix it on development for cases when the database is empty.

spwoodcock:
Oh no it wasn't actually to do with the SQL formatting. 

It was some line length errors and redefined variables.

To run the linting locally and check, run:

```
pip install pre-commit
pre-commit run --all-files
```

varun2948:
i did that too. Does this command auto adjust the the sql query ?

spwoodcock:
No it can't format multiline strings unfortunately. 

It formats all other code though, sorts imports, lints code etc.

varun2948:
so we have to manually adjust the sql query right ?

spwoodcock:
Yes exactly - any multiline strings need to be manually formatted

varun2948:
Sam, I just heard that Tm task history is huge on tm as it has alots of stuff related with tasks and due to lots of row it is taking alot of time to query data out of it. This is fine right the way we are replicating same as TM for the task history. Just a query…

spwoodcock:
Yeah I can imagine!

It's probably not the most efficient or simple strategy implemented (using history entries for everything, including comments).

If we imagine that we have 100 projects, then 100 tasks per project, then 100 history records per task, this equals 1,000,000 task history entries quite quickly.

I'm not opposed to separating comments out to a different table: id, task id, user id, comment.

I guess I don't really have clarity on if we want to keep the structure of TM as much as possible for compatibility (what I have tried to do until now), or if we should optimise the FMTM schema to what we think is best.

Any thoughts @robsavoye ?

robsavoye:
the task_history table is huge, and I don't think FMTM will have the same usage patterns. For TM, mappers interact with the website, but for FMTM, they're using Collect on their phone. TM has a project_chat table which I'm integrating into tm-admin, which might be more appropriate. FMTM will have some task_history itself, it's entirely possible an FMTM task will have bad data, so force TM to invalidate the task and have it mapped and validated. Then FMTM would have some changes when the task is field mapped again.

Since we're working on code sharing between FMTM and TM, we should make a unified plan.  What type of chatting, commenting, user messaging, etc... do we need to support ? It's going to be different user stories, but the DB schema and access API should be the same. We'd implement this in tm-admin to use the new schema and the fully async code that uses it to avoid future performance problems.

PR #1179, opened 2024-02-07
Fix update submissions in s3

### LINKED ISSUE #1170: Number of forms submitted not visible in form submission graphs
**Describe the bug**
I submitted few forms. But i am unable to see my submissions on form submission graph of submission pages

**To Reproduce**

1. Go to project details page
2. Click on "view infographics"
3. Scroll down to 'Submission form graph'

**Expected behavior**
graph showing number of form submitted per day as in designs. 

**Screenshots**
image

Sujanadh:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Issue

- Fixes #1170 

## Describe this PR

This PR solves the issue of submissions not being update to s3. while checking last submission there needs to be object created in s3 fmtm-data bucket. But if there are not any submissions in s3 the ValueError was raised from s3 which didn't allow to add obj to s3 further. So instead of raising error 'None' is returned.

### DIFF
--- a/src/backend/app/s3.py
+++ b/src/backend/app/s3.py
@@ -3,7 +3,7 @@
 import json
 import sys
 from io import BytesIO
-from typing import Any
+from typing import Any, Optional
 
 from loguru import logger as log
 from minio import Minio
@@ -93,7 +93,7 @@ def get_file_from_bucket(bucket_name: str, s3_path: str, file_path: str):
     client.fget_object(bucket_name, s3_path, file_path)
 
 
-def get_obj_from_bucket(bucket_name: str, s3_path: str) -> BytesIO:
+def get_obj_from_bucket(bucket_name: str, s3_path: str) -> Optional[BytesIO]:
     """Download an S3 object from a bucket and return it as a BytesIO object.
 
     Args:
@@ -112,8 +112,8 @@ def get_obj_from_bucket(bucket_name: str, s3_path: str) -> BytesIO:
     try:
         response = client.get_object(bucket_name, s3_path)
         return BytesIO(response.read())
-    except Exception as e:
-        raise ValueError(str(e)) from e
+    except Exception:
+        return None
     finally:
         if response:
             response.close()
--- a/src/backend/app/s3.py
+++ b/src/backend/app/s3.py
@@ -112,8 +112,9 @@ def get_obj_from_bucket(bucket_name: str, s3_path: str) -> Optional[BytesIO]:
     try:
         response = client.get_object(bucket_name, s3_path)
         return BytesIO(response.read())
-    except Exception:
-        return None
+    except Exception as e:
+        log.warning(f"Failed attempted download from S3 path: {s3_path}")
+        raise ValueError(str(e)) from e
     finally:
         if response:
             response.close()
--- a/src/backend/app/s3.py
+++ b/src/backend/app/s3.py
@@ -3,7 +3,7 @@
 import json
 import sys
 from io import BytesIO
-from typing import Any, Optional
+from typing import Any
 
 from loguru import logger as log
 from minio import Minio
@@ -93,7 +93,7 @@ def get_file_from_bucket(bucket_name: str, s3_path: str, file_path: str):
     client.fget_object(bucket_name, s3_path, file_path)
 
 
-def get_obj_from_bucket(bucket_name: str, s3_path: str) -> Optional[BytesIO]:
+def get_obj_from_bucket(bucket_name: str, s3_path: str) -> BytesIO:
     """Download an S3 object from a bucket and return it as a BytesIO object.
 
     Args:
--- a/src/backend/app/submissions/submission_crud.py
+++ b/src/backend/app/submissions/submission_crud.py
@@ -357,22 +357,19 @@ def update_submission_in_s3(
         metadata_s3_path = f"/{s3_project_path}/submissions.meta.json"
         try:
             # Get the last submission date from the metadata
-            file = get_obj_from_bucket(settings.S3_BUCKET_NAME, metadata_s3_path)
-            if file:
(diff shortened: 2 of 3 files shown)


src/backend/app/s3.py
```diff
@@ -112,8 +112,8 @@ def get_obj_from_bucket(bucket_name: str, s3_path: str) -> BytesIO:
     try:
         response = client.get_object(bucket_name, s3_path)
         return BytesIO(response.read())
-    except Exception as e:
-        raise ValueError(str(e)) from e
+    except Exception:
```
spwoodcock, line 115: Try not to use bare Except with `from e`, otherwise linters will show an error

src/backend/app/s3.py
```diff
@@ -112,8 +112,8 @@ def get_obj_from_bucket(bucket_name: str, s3_path: str) -> BytesIO:
...
+        return None
```
spwoodcock, line 116: I think the exception should remain here, as other things may rely on the information if download fails:
```python
log.warning(f"Failed attempted download from S3 path: {s3_path}")
raise ValueError(str(e)) from e
```

Then you an put the exception catch higher up in the submission logic:
```python
try:
    data = await get_obj_from_bucket(xxx, xxx)
except ValueError as e:
    log.debug(f"Submission data does not exist in S3 for project {project_id}")
    data = None
```

Perhaps we even want to halt execution and raise an error to the user instead (depending on the context)?

```python
try:
    data = await get_obj_from_bucket(xxx, xxx)
except ValueError as e:
    msg = f"Submission data does not exist in S3 for project {project_id}"
    log.debug(msg)
    raise HTTPException(status_code=HTTPStatus.NOT_FOUND, detail=msg) from e
```

src/backend/app/submissions/submission_crud.py
```diff
@@ -358,14 +358,17 @@ def update_submission_in_s3(
...
             file = get_obj_from_bucket(settings.S3_BUCKET_NAME, metadata_s3_path)
-            zip_file_last_submission = (json.loads(file.getvalue()))["last_submission"]
-            if last_submission <= zip_file_last_submission:
-                # Update background task status to COMPLETED
-                update_bg_task_sync = async_to_sync(
-                    project_crud.update_background_task_status_in_database
-                )
-                update_bg_task_sync(db, background_task_id, 4)  # 4 is COMPLETED
-                return
+            if file:
```
spwoodcock, line 361: For this kind of check, I always find it good to do a pattern like:

```python
if not file:
    return # note this should probably return something, even if None
# continue logic here
```

it prevents the extra indenting

spwoodcock (changes requested):
See comments

src/backend/app/s3.py
Sujanadh: usually we dont have submissions in s3, so most of the time there wont be any objects related to that project. In that case we need to update submissions in s3. if we raise error from this point , it wont update submissions in s3 at very first place.

spwoodcock: Ah that makes sense. In this case, the error handling would be:

```python
try:
    data = await get_obj_from_bucket(xxx, xxx)
except ValueError as e:
    msg = f"Submission data does not exist in S3 for project {project_id}. Continuing to creation"
    log.debug(msg)
```

Sujanadh: I found the main issue is elsewhere, i think some changes needs to be done in osm_fieldwork, i will let you know after i completely track the issue and debug it.

Sujanadh: the issue was in downloadThread ,
previously ,
```python
subs = form.listSubmissions(project_id, task, filters)
if type(subs) == dict:
    log.error(f"{subs['message']}, {subs['code']} ")
    continue
# log.debug(f"There are {len(subs)} submissions for {task}")
if len(subs) > 0:
    data += subs
```
here subs is always dict so it obstructs data to be appended.
i am thinking of solution like this which solves this issue.
```python
subs = form.listSubmissions(project_id, task, filters)
if subs == {}:
    log.error(f"{subs['message']}, {subs['code']} ")
    continue
# log.debug(f"There are {len(subs)} submissions for {task}")
if len(subs["value"]) > 0:
    data += subs["value"]
```
Since we want to append only the values of key "value" of submissions and upload to s3.

Sujanadh: > Ah that makes sense. In this case, the error handling would be something like:
> 
> ```python
> try:
>     data = await get_obj_from_bucket(xxx, xxx)
> except ValueError as e:
>     log.debug(f"Submission data does not exist in S3 for project ({project_id}). Data will be generated.")
> ```

it will make error message static , i dont think we should do this, since other errors may also be raised from `get_obj_from_bucket(xxx, xxx)`. i think we should leave it as it is.

src/backend/app/s3.py
```diff
@@ -93,7 +93,7 @@ def get_file_from_bucket(bucket_name: str, s3_path: str, file_path: str):
     client.fget_object(bucket_name, s3_path, file_path)
 
 
-def get_obj_from_bucket(bucket_name: str, s3_path: str) -> BytesIO:
+def get_obj_from_bucket(bucket_name: str, s3_path: str) -> Optional[BytesIO]:
```
spwoodcock, line 96: Sorry to be pedantic, but now the error handling was reverted, this function always returns `BytesIO`, so `Optional` is not necessary.

spwoodcock (changes requested):
Will merge after small edit 👍

spwoodcock (approved):

PR #1191, opened 2024-02-12
Filter on submission table

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Describe this PR

This PR implements filter on project-wise submissions and task-wise submissions where filters are applied directly on json for project wise submission and filter param is passed to odk central to filter task-wise submissions. For now username/submitterName (submitted_my) filter is not working, need advice on this @spwoodcock .

## Screenshots

N/A


nrjadkry:
It looks like at the moment, we're limited to filtering against the four parameters mentioned in the documentation. here https://docs.getodk.org/central-api-odata-endpoints/#data-document

Unfortunately, this means we can't directly filter submission data using form metadata like usernames. We'll have to put a pause on filtering data against users directly from ODK for now while we explore alternative solutions.

spwoodcock:
As we have the submission json in the S3 bucket, can't we just filter that?

Sujanadh:
yes we are filtering submissions by project directly on json but for task wise submission we have to rely on odk central since we dont have task wise submissions in s3

spwoodcock:
But if the project submissions json is regularly updated & includes the task id field, then can't we just filter the project submission by both task and user?

Sujanadh:
If task_id is included in the submission, then it won't be a problem. We have to test the performance of the filter since submissions can be large.

spwoodcock:
I haven't managed to test properly yet, but I used sozip to hopefully make the data read accessible.

On reflection, I think a better format for this kind of data would have been GeoParquet - specially for tabular data.
If we had a GeoParquet file of submissions, it would be very efficient to filter data without downloading it all.

We need to do a bit more research on usage, or other ideas. But for now, we can use the json as it is, then optimise later 😄

Sujanadh:
For now, we can merge it and work on it; there is always room for optimization. :smiley:

spwoodcock:
Sorry, I must have missed this. 

Remember that as long as somebody reviews the code and it's well documented, then other people can merge too (we don't want everything to be waiting on me!)

PR #1197, opened 2024-02-13
Add community_type for organisations, add unapproved org list endpoint

### LINKED ISSUE #1189: organisation dropdown isn't visible after refreshing create project page
**Describe the bug**
the organisation dropdown shows the list of all the organisations in initial step of project creation. However, once the user refresh create project page, the dropdown shows no organisations to be choosed.

**To Reproduce**

1. Go to 'create project step 1.
2. Refresh the browser
3. Click on organisation name dropdown

**Expected behavior**
the list of organisations should be visible on dropdown.

**Screenshots**
https://vimeo.com/912136553/770aa8c21f?share=copy

Sujanadh:

- [X] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Issue

Fixes #1189

## Describe this PR

- Implemented login_required for organization permissions
- added API for listing unapproved organizations
- added community_type field in organizations
- made org_id mandatory for project creation
- removed org_id from update project since it is not required while updating.

## Screenshots

N/A


src/backend/app/projects/project_schemas.py
```diff
@@ -133,7 +133,7 @@ class ProjectIn(BaseModel):
     """Upload new project."""
 
     project_info: ProjectInfo
-    organisation_id: Optional[int] = None
+    organisation_id: int
```
spwoodcock, line 136: I removed this intentionally, as it doesn't actually do anything.
If you check the logic for create_project, the org id is taken from the url route param, not from the json body 👍

src/backend/tests/test_projects_routes.py
```diff
@@ -62,6 +62,7 @@ async def test_create_project(client, admin_user, organisation):
         },
         "xform_title": "buildings",
         "hashtags": ["#FMTM"],
+        "organisation_id": organisation.id,
```
spwoodcock, line 65: same for this

src/frontend/src/components/createnewproject/SplitTasks.tsx
```diff
@@ -119,6 +119,7 @@ const SplitTasks = ({ flag, geojsonFile, setGeojsonFile, customLineUpload, custo
       hashtags: arrayHashtag,
       data_extract_type: projectDetails.data_extract_type,
       data_extract_url: projectDetails.data_extract_url,
+      organisation_id: projectDetails.organisation_id,
```
spwoodcock, line 122: and this

src/backend/app/organisations/organisation_routes.py
```diff
@@ -49,6 +49,15 @@ async def get_organisations(
     return await organisation_crud.get_organisations(db, current_user)
 
 
+@router.get("/unapproved/", response_model=list[organisation_schemas.OrganisationOut])
```
spwoodcock, line 52: This already existed before on the "/" endpoint with a flag `approved=True|False`.
I removed that option, as I didn't think we needed to display unapproved orgs via the API.

It's a super admin only thing, so the only possible application is if we have an admin dashboard on the frontend (which doesn't really make sense imo).

Can you think of a use case why we need this code?

src/backend/app/db/db_models.py
```diff
@@ -168,6 +169,14 @@ class DbOrganisation(Base):
     odk_central_user = cast(str, Column(String))
     odk_central_password = cast(str, Column(String))
 
+    email = cast(str, Column(String))
```
spwoodcock, line 172: Why would the organisation need an email? Is this not something that would be set on the DbUser model?

src/backend/app/organisations/organisation_routes.py
nrjadkry: How do the super admin approve the organisation requests ?
The idea is to display a list of pending organisations, only visible to the super admin. And, then accepts or rejects them.

src/backend/app/db/db_models.py
nrjadkry: Does the organisation email differ from the user email? It not, we can retrieve from the user, else we need it.

src/backend/app/organisations/organisation_routes.py
spwoodcock: But unless we have an admin dashboard to call the API, this isn't a very user friendly way to approve organisations (having to regularly check the unapproved list).

Ideally the admin receives an email asking for approval, then accepts either by clicking a button (which calls an endpoint), via the approval endpoint, or in the database.

spwoodcock: If you have been discussing and plan to have an admin dashboard for organisation approval, then this makes sense 👍

src/backend/app/db/db_models.py
spwoodcock: The organisation needs an admin, so the org admin user will ideally have something in the `email_address` field.

I don't think we need the email field on organisations.

src/backend/app/organisations/organisation_routes.py
nrjadkry: Yes, the idea is to have a simple dashboard for super admin and the organisation admins, where they can assign new users as the admin, approve/reject the organisations, like those features based on the user roles.

src/backend/app/projects/project_schemas.py
Sujanadh: I understand, but projects were created even without an organization ID. I don't know how, which is why it was made mandatory.

src/backend/app/organisations/organisation_routes.py
spwoodcock: Makes sense, thanks for clarifying =)

src/backend/app/projects/project_schemas.py
spwoodcock: We would have to specify org_id in both the url and the data json, potentially causing conflicts if different values are passed.

This line should add the org_id to the project: https://github.com/hotosm/fmtm/blob/bd6c74379dd167a787b1cab050eab4f65bc9cb92/src/backend/app/projects/project_routes.py#L267

When were you testing? Perhaps I fixed it since you last tested (I have been heavily refactoring the project creation, sorry if it causes issues 🙏 )

Sujanadh: The day before yesterday.

spwoodcock: Please test the create_project endpoint and report if errors 👍

spwoodcock:
This PR is needed to fix https://github.com/hotosm/fmtm/issues/1189, so I will update slightly & merge 👍

PR #1205, opened 2024-02-13
Mandatory odk credentials if no organisation default

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Describe this PR
1. Default #FMTM tag add for `Tags field`

2. @spwoodcock, currently there's a loophole where ODK details are optional while creating an organization as well as creating a project. As of now a user can left the ODK fields empty while creating a project which shouldn't be the case if the user haven't filled ODK credentials on organization creation. So, now all the ODK fields are made required as of before until works on backend are done for checking if ODK credentials are already filled or not. Currently error is shown at the last step if no ODK credentials are provided.

**Propose Solution:** 
First check if the ODK credentials are already filled during organization creation and accordingly show ODK fields to be required if the ODK credentials are not provided and not show required if ODK credentials are provided during organization creation.

@manjitapandey @Sujanadh

## Screenshots
image


spwoodcock:
The `#FMTM` tag is already added by default in the backend

spwoodcock:
We should discuss this issue.
For the public beta we can't have these fields mandatory (users may not even know what ODK Central is).
We need an alternative solution, something like:
- On the project creation page the organsations are listed in the dropdown).
- The user selects an organisation.
- Based on the selected org, we set a flag if the org has credentials set or not (we may need to modify the backend to return the central url if there is one set).
- If no url set, then make the field mandatory.
- If url is set, let the user specify if the want default creds or will provide.

NSUWAL123:
Yes, @spwoodcock, we also came to the same conclusion as yours.

NSUWAL123:
> nclude the odk_central_url in the organisation json, if one exists.
> I just fully read the page, sorry @NSUWAL123 you proposed basically the same thing

Sam I have removed default #FMTM tag

varun2948 (approved):

spwoodcock (changes requested):
As comments above - ODK creds must be optional if the organisation has them set.

NSUWAL123:
@spwoodcock, now if ODK credentials has been included while organization creation then there is an checkbox in project creation: 
if organization is associated with ODK cred (option where defualt or custom ODK credentials can be filled):
image
if no ODK cred is linked while org creation (user mandatorily needs to fill ODK credentials): 
image

In organization creation: 
Toggle checkbox accordingly to fill or not fill ODK cred: 
image
image

varun2948 (approved):

spwoodcock (approved):

PR #1225, opened 2024-02-16
Limit the no.of threads according to CPU core

nrjadkry:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Describe this PR
I have limited the number of threads running when generating QR Codes, by looking into the cpu cores. Sometimes, fmtm backend freezes when creating a project. This might be a potential solution. What do you think @spwoodcock .
I have kept 2 cpu cores free.

### DIFF
--- a/src/backend/app/projects/project_crud.py
+++ b/src/backend/app/projects/project_crud.py
@@ -32,6 +32,7 @@ import requests
 import shapely.wkb as wkblib
 import sozipfile.sozipfile as zipfile
 from asgiref.sync import async_to_sync
+from cpuinfo import get_cpu_info
 from fastapi import File, HTTPException, Response, UploadFile
 from fastapi.concurrency import run_in_threadpool
 from fmtm_splitter.splitter import split_by_sql, split_by_square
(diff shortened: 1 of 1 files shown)


spwoodcock:
ThreadpoolExexutor doesn't use cores, so this won't work as expected.

Did you see any change?

The uvicorn is set to only use one worker, so it makes sense that while project generation is happening the server cannot respond, as that one worker is busy.

But, we have multiple replicas of the container, so the idea is to spawn multiple FastAPI instances that run in parallel. For dev we default to 2 replicas, so I imagine only two projects can be created at one time. 

In production we increase the replica number.
Have you tried testing with this?

nrjadkry:
I have not test this thoroughly. I thought threadpoolexecutor can be limited to the max workers. Thanks for this info.

Dev server sometimes freezes when we create project and whenever background tasks are running. 
If we have more replicas of container in the production, then I guess we won't have any problem.

PR #1232, opened 2024-02-20
Add user in organisation

### LINKED ISSUE #1229: Make User an Organization Admin When Requested Organisation is Approved
The user requesting an organisation is still a mapper, even though the organisation is approved.

On approval, the user creating organisation should be an organisation admin.
When creating an organization, user details should be stored in the organisation table. Upon approval of the organization, the user who created it should be designated as an organisation manager.

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Issue

- Fixes #1229 

## Describe this PR

- The user is added to the organization table as a foreign key to save user information after the organization is created.
- Add the respective user as an organization_manager while the organization is approved.

## Screenshots

N/A


spwoodcock:
With this approach an organisation can only have one user?

Sujanadh:
> With this approach an organisation can only have one user?

Only one user is associated in creation of organisation, but it can have mutliple users as organisation managers.

spwoodcock:
Do we need a record of who created the organisation for something in specific?

The org creator is automatically the first org manager, so I would think that's enough info.

Are you looking for a way to get find which users are associated with an organisation?

Sujanadh:
After approval of the organization, the user needs to be an organisation admin by default. For this purpose, we need to save users in the organization table. I don't think the user is automatically org_manager even after creating organisation. We need to add the user as an org_manager manually.

spwoodcock:
Sorry for all the questions - maybe I'm missing something!

Why do we need to add the user to the organisations table, when we already have the organisation_managers table for this purpose?
Can't we just add them to that?

Sujanadh:
yes you are right if we are making user an org_manager just after creating organisation even though it is not approved. My concern was since the organisation won't be approved till the super user approves it,  the user will get privilege of org_admin.
For that to not happen we need to make the user admin only after approval thats why we need user in org to know which user created that organisation. what do you think?

spwoodcock:
Ah you raise a very good point!

Options I can think of:
- Create the organisation_managers entry by default, but add an extra check for `approved=True` in the SQL for `check_access` (probably not the best, as it complicates the SQL there / adds another table).
- Store the uid of the creating user somewhere, as you suggest.
- Pass the uid of the creating user as a param to the /approve/ endpoint.

The uid would be included in the email/message to the super admin, and in the button for them to click with the URL param included.

However, if the email failed to send, we would have no record of which user requested org creation 🤔

spwoodcock:
I didn't really want to create an extra db column and migration for this purpose, but I think it's useful to have the trail of information for who requested org creation.

The solution you suggested is probably the best one 👍 
(I may just update the field user_id --> created_by)

src/backend/app/organisations/organisation_crud.py
```diff
@@ -324,10 +330,27 @@ async def approve_organisation(db, organisation):
...
+        log.info(f"Approving organisation ID {organisation.id}")
+        user = await user_exists_in_db(organisation.user_id, db)
+
+        if organisation and user:
+            organisation.managers.append(user)
+            organisation.approved = True
+            db.commit()
+            return Response(
+                status_code=HTTPStatus.OK,
+                detail="Organisation is approved successfully",
```
spwoodcock, line 343: The `detail` variable does not exist on the `Response` object. You are probably thinking of HTTPError.
If you wish to display a message, probably best to return a JSONResponse.

spwoodcock: In this case I would use a `response_model` though and return the modified organisation JSON.

Sujanadh:
> I didn't really want to create an extra db column and migration for this purpose, but I think it's useful to have the trail of information for who requested org creation.
> 
> The solution you suggested is probably the best one 👍
> 
> I will just update this PR with two small changes:
> 
> * Change field name user_id --> created_by.
> * Remove the foreign key constraint, to make the column more flexible. It should just be an int that stores the requesting user id, but we should never need to access the user details via this route. If the user deletes their profile etc in the future, this won't cause any conflicts.
> 
> Do you think this is sensible?

yeah you are right. I agree with that

PR #1334, opened 2024-03-07
Implemented raw sql replacing sqlalchemy in me endpoint

### LINKED ISSUE #1333: Implement user roles in me endpoint
Currently `me` endpoint returns only roles which are Mapper, Read only and Admin. It should also return project_roles, associated project and organisation if created (org_admin).

- add project_roles, project_id, org_id in the endpoint
- Implement raw sql to avoid multiple db calls

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [X] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert
## Related Issue

- Fixes #1333 

## Describe this PR

- This PR adds user roles along with project_roles, associated project, and organisation if created by the user, which means if the user is the org_admin.
- It implements raw SQL in order to check and update the user or create the user.
- Tried to use the CASE statement, but it didn't work as expected since it only allows to return boolean-type results (true, false) but not for update purposes, if I am not wrong.

One problem I encountered while performing the insert action was that I needed to mention all the not null fields even though they have default values, unlike those handled by Alchemy automatically since we do not have the default values set at the database level.

## Screenshots

Different Responses:
Screenshot from 2024-03-07 23-20-48
Screenshot from 2024-03-07 23-22-09

### DIFF
--- a/src/backend/app/auth/auth_routes.py
+++ b/src/backend/app/auth/auth_routes.py
@@ -28,6 +28,7 @@ from sqlalchemy.orm import Session
 from app.auth.osm import AuthUser, init_osm_auth, login_required
 from app.config import settings
 from app.db import database
+from app.models.enums import UserRole
 
 router = APIRouter(
     prefix="/auth",
@@ -132,21 +133,7 @@ async def get_or_create_user(
     try:
         update_sql = text(
             """
-        DO
-        $$
-        BEGIN
-            IF EXISTS (SELECT 1 FROM users WHERE id = :user_id) THEN
-                UPDATE users
-                SET profile_img = :profile_img
-                WHERE id = :user_id;
-            ELSIF EXISTS (SELECT 1 FROM users WHERE username = :username) THEN
-                -- Username already exists, raise an error
-                RAISE EXCEPTION
-                    '
-                    User with this username % already exists
-                    ', :username;
-            ELSE
-                INSERT INTO users (
+            INSERT INTO users (
                     id, username, profile_img, role, mapping_level,
                     is_email_verified, is_expert, tasks_mapped, tasks_validated,
                     tasks_invalidated, date_registered, last_validation_date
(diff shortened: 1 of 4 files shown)


Sujanadh:
I will fix the test cases related to the user.

src/backend/app/auth/auth_routes.py
```diff
@@ -127,45 +127,104 @@ async def logout():
...
+                WHERE id = :user_id;
+            ELSIF EXISTS (SELECT 1 FROM users WHERE username = :username) THEN
+                -- Username already exists, raise an error
+                RAISE EXCEPTION
+                    '
+                    User with this username % already exists
+                    ', :username;
+            ELSE
+                INSERT INTO users (
+                    id, username, profile_img, role, mapping_level,
```
spwoodcock, line 150: This could possibly be simplified by the suggestion Ping made, with an 'upsert' query.

Something like

```
INSERT INTO table_name (column1, column2, ...)
VALUES (value1, value2, ...)
ON CONFLICT (id, username)
DO UPDATE SET column1 = value1, column2 = value2, ...;
```

To make this work do we need a migration to ensure the username field has `UNIQUE` specified? Both changes combined would simplify this query a lot 👍

spwoodcock (changes requested):

src/backend/app/auth/auth_routes.py
Sujanadh: Ok, Sure. By the way, the username field was already a unique field.

Sujanadh:
updated the sql using `CONFLICT ON`

src/backend/app/auth/auth_routes.py
```diff
@@ -127,45 +128,89 @@ async def logout():
...
-    db.commit()
-
-    return db_user
-
-
-@router.get("/me/", response_model=AuthUser)
+        get_sql = text(
+            """
+            SELECT users.*,
+                COALESCE(user_roles.project_id) as project_id,
```
spwoodcock, line 167: Coalesce could be removed here, as it's for merging two values together.

It's useful to set the default value to mapper below, but not needed for the user role or org manager parts 👍

src/backend/app/auth/auth_routes.py
```diff
@@ -127,45 +128,89 @@ async def logout():
...
+            LEFT JOIN user_roles ON users.id = user_roles.user_id
+            LEFT JOIN organisation_managers on users.id = organisation_managers.user_id
+            WHERE users.id = :user_id;
+            """
+        )
+        result = db.execute(
+            get_sql,
+            {"user_id": user_data.id},
+        )
+        db_user = result.fetchall()
```
spwoodcock, line 180: .first() will return a single value, so that you don't have to index with [0] below. Then you also can avoid needing the for loop too 👍

spwoodcock (commented):
Looks great!

Just a few minor points and it can be merged

src/backend/app/auth/auth_routes.py
Sujanadh: Yes, you are right. Great! 
Thanks for the feedback!!

spwoodcock (approved):

Issue #1365, opened 2024-03-19
page crash while uploading AOI with no feature during project creation


manjitapandey:
**Describe the bug**

**To Reproduce**

1. Go to 'create project'
2. go to step 2 
3. upload geojson file having no features 

**Expected behavior**
the proper error message should be displayed prompting users to reupload correct boundary file. but the page shouldn't crash.

**Screenshots**
![image](https://github.com/hotosm/fmtm/assets/97273021/ed9c6c0e-775b-471a-a1d9-e76f32870adf)

**Additional context**

[presentation.zip](https://github.com/hotosm/fmtm/files/14650001/presentation.zip)

PR #1367, opened 2024-03-20
fix uploadArea: geojson with no features leading page crash error solve

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Fix: #1365

## Describe this PR
This PR solves the page crash problem whenever a geojson with empty features is uploaded.

## Screenshots
image

## Review Guide

### DIFF
--- a/src/frontend/src/components/MapComponent/OpenLayersComponent/Layers/VectorLayer.js
+++ b/src/frontend/src/components/MapComponent/OpenLayersComponent/Layers/VectorLayer.js
@@ -18,8 +18,8 @@ import MultiPoint from 'ol/geom/MultiPoint.js';
 import { buffer } from 'ol/extent';
 import { bbox as OLBbox } from 'ol/loadingstrategy';
 import { geojson as FGBGeoJson } from 'flatgeobuf';
-
 import { isValidUrl } from '@/utilfunctions/urlChecker';
+import { featureCollectionHasFeatures } from '@/utilfunctions/commonUtils';
 
 const selectElement = 'singleselect';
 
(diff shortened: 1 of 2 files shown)


NSUWAL123:
@varun2948 , @spwoodcock 
Although this PR solves the page crash problem. But there may be hundreds of cases where the geojson can be invalid which can lead to page crash.
for example: if a user uploads this geojson;
{"type": "FeatureCollection"}

Since the geojson is invalid it leads to page crash. Now how to handle each of the cases in a good way to avoid page crashes? Wrapping every function and code to a try-catch block in vectorLayer file or what?

Can you give some suggestions. Also, geojson-validator doesn't work in this case.

spwoodcock:
Thanks for the detailed feedback!

This PR may not work well, as geojson is not necessarily a FeatureCollection.

We need to support Polygon, MultiPolygon,  (maybe) GeometryCollection and FeatureCollection.

Like you say, is there a way to simply try catch the loading of the geojson File via OpenLayers?

If the loading works, but the crs is invalid, keep the existing message.

If loading fails though, then display a different message like: GeoJSON file is invalid

spwoodcock:
I think you may need to find where the geojson data is included as a source for the vectorlayer and find the place where the try catch must be.

If the error isn't triggered until the layer is rendered, there may be other ways to catch this issue.

For example using .getSource().getGeometry() and if there is no geometry return an error. 

The error should probably be bubbled up to the project creation page and a more informative toast error displayed from there

PR #1380, opened 2024-03-22
Separate endpoint to check the validity of login cookies.

### LINKED ISSUE #1289: Frontend login state is sometimes incorrect
**Describe the bug**

- Calling `/me` sets the users login cookie.
- It also sets the frontend store login details.

Two scenarios could put these out of sync:
- The cookie expires, but the frontend state has not been cleared (i.e. because the logout endpoint was not called).
- The user was logged in, but the server had an update. The cookie is invalidated, but the login state persists (uncertain about this one, as in theory the cookie value should always be the same).

This would mean the user is not blocked on the frontend to e.g. create a project, but when they call the API they will get an error because they are not logged in.

**To Reproduce**

- Log in via FMTM, delete your cookie manually, attempt project creation.

or

- Log in via FMTM on dev, wait for a server update (quite regular), attempt project creation after.

**Expected behavior**

- We need a quick and simple way to check the login state for a user when they refresh or enter the FMTM website.
- If the loginSlice store contains user details, but the user is in fact no logged in, then the values in loginSlice should be cleared.
- In theory any endpoint with a simple `login_required` on the backend would work for this, but to reduce server load it would be best to create a new minimal endpoint: `/auth/introspect`.
- This endpoint should be called by the frontend at certain points, either:
  - At first page load.
  - Or on specific functions (harder to maintain, but lower load). E.g. at the start of project creation.

@manjitapandey

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Fixes #1289 

## Describe this PR

This PR adds new endpoint that checks for login cookie validity. Returns True if valid and False for the invalid cookie.

### DIFF
--- a/src/backend/app/auth/auth_routes.py
+++ b/src/backend/app/auth/auth_routes.py
@@ -220,7 +220,7 @@ async def my_data(
     return await get_or_create_user(db, user_data)
 
 
-@router.get("/introspect")
+@router.get("/introspect", response_model=AuthUser)
 async def check_login(
     db: Session = Depends(database.get_db),
     user_data: AuthUser = Depends(login_required),
@@ -229,4 +229,4 @@ async def check_login(
 
     Returns True if authenticated, False otherwise.
     """
-    return bool(user_data)
+    return user_data
--- a/src/backend/app/auth/auth_routes.py
+++ b/src/backend/app/auth/auth_routes.py
@@ -218,3 +218,15 @@ async def my_data(
         user_data(dict): The dict of user data.
     """
     return await get_or_create_user(db, user_data)
+
+
+@router.get("/introspect")
+async def check_login(
+    db: Session = Depends(database.get_db),
+    user_data: AuthUser = Depends(login_required),
+):
+    """Verifies the validity of login cookies.
+
+    Returns True if authenticated, False otherwise.
+    """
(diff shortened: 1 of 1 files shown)


src/backend/app/auth/auth_routes.py
```diff
@@ -218,3 +218,15 @@ async def my_data(
...
+
+@router.get("/introspect")
+async def check_login(
+    db: Session = Depends(database.get_db),
+    user_data: AuthUser = Depends(login_required),
+):
+    """Verifies the validity of login cookies.
+
+    Returns True if authenticated, False otherwise.
+    """
```
spwoodcock, line 231: Good stuff 😄

Thinking about it, the endpoint may as well return the actual user data json, as the frontend could then update the profile picture if it changed since last login 👍

spwoodcock (changes requested):

src/backend/app/auth/auth_routes.py
Sujanadh: okay

spwoodcock (approved):

Issue #1393, opened 2024-03-27
start mapping button is not functional on mobile website and PWA


manjitapandey:
**Describe the bug**
I am unable to click on start mapping of any task on mobile. However it works on laptop. 

**To Reproduce**

1. Go to 'project page'
2. Click on 'any task'
3. Click on "start mapping"

**Expected behavior**
The status of task should be changed to locked for mapping. 

**Screenshots**

**Smartphone (please complete the following information):**

- Device: [Oppo F19 pro]
- Browser [Chrome]

spwoodcock:
Hmm, is this dev stage or prod?

manjitapandey:
It was replicable on all servers. Seems like this issue is there when users aren't logged in. 
The solution could be either allowing user to lock task without logging in or redirecting them to log in page.

spwoodcock:
For now can we allow locking and unlocking without log in? @Sujanadh

Sujanadh:
> For now can we allow locking and unlocking without log in? @Sujanadh

okay but how do we track who updated the status?

spwoodcock:
Good question!

We either don't provide the info of who updated & allow unauthenticated access, or provide a basic API for a simple type of auth.

We talked about email based auth being an option, but it would be very nice if we don't even need an email address!

What do you think about us having a temporary login token option, valid for one week?

The token would have restricted access to only view projects and update task status. 

The user could provide their name in an input field, and the API will generate and return a token in a cookie. 
This token will behave like the OSM cookie, but only be valid for one week

Open to other ideas too!

Sujanadh:
It sounds good to me

Sujanadh:
what about using jwt ? or are we saving token in database for basic authentication?

Sujanadh:
saving user info wont be possible unless we create a separate table for that. if it is temporary approach then we don't have to think much on saving user data. Using svcfmtm might be good as you mentioned.

spwoodcock:
I think it's the way to go for now then 👍

We can work out how to link to private projects later. 

One idea:
- The temp tokens can be granted by the project admin, where they enter an email address.
- Then a link is sent to the user.
- When they click a link in the email it sets a temp cookie via the API.

manjitapandey:
I am unable to test this since. the project area is not visible in mobile.

https://github.com/hotosm/fmtm/assets/97273021/611fb74c-c53e-40d9-b200-00a3e065c7f5

spwoodcock:
This should be testable now.

PR #1410, opened 2024-04-02
Temporary Authentication login

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue #1393 

## Describe this PR

This PR implements JWT authentication with a validity of 7 days. It includes a new endpoint, `/temp/login/` which creates a token and sets a cookie, and a wrapper function to verify the token and return static user info, which is `svcfmtm`.

## Review Guide


Sujanadh:
- I have used PyJWT package
- AUTH_SECRET_KEY with 64 bytes is randomly generated and is used in .env file
- Still need to update the dependency to update-task-status endpoint and related project  summary endpoints
- PyJWT package and pdm lock file are not pushed yet
- open to suggestions and improvements

spwoodcock:
This is pretty neat 😁

But there is no need for us to generate our own tokens, as this can be handled by OpenStreetMap.

I will send you the OSM auth token for svcfmtm and we can simply verify that using login_required.

The token can be set inside a cookie in the same way we do currently too

Sujanadh:
ohh ok i thought we are not using osm auth. I totally forgot svcfmtm is already an osm user ,not just made up user.

Sujanadh:
i have updated the temp login to use osm auth tokens.
what are the other endpoints that we are giving temporary access?

spwoodcock:
For sure we need to update the task status (lock/unlock) - that's the main one 👍

Sujanadh:
I have added login_required in update_task_status , is there anything to add?

src/backend/app/config.py
```diff
@@ -116,6 +116,9 @@ def assemble_db_connection(cls, v: Optional[str], info: ValidationInfo) -> Any:
     OSM_URL: HttpUrlStr = "https://www.openstreetmap.org"
     OSM_SCOPE: str = "read_prefs"
     OSM_LOGIN_REDIRECT_URI: str = "http://127.0.0.1:7051/osmauth/"
+    OSM_SVC_TOKEN: str = """ZXlKcFpDSTZNakF6T0RZeU1Ua3NJblZ6WlhKdVlXMWx
```
spwoodcock, line 119: This shouldn't be hardcoded, but configurable.

In the Pydantic, specifying `Optional[str]` makes the field nullable. Specifying `Optional[str] = None` makes the field optional (what we want).

The value is then injected via the `.env` file.
Don't worry about updating the example dotenv etc, I can do that after.

src/backend/app/auth/temp_auth.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    access_token = settings.OSM_SVC_TOKEN
+
+    response = JSONResponse(content={"access_token": access_token}, status_code=200)
+    cookie_name = settings.FMTM_DOMAIN.replace(".", "_")
+    response.set_cookie(
+        key=cookie_name,
+        value=access_token,
+        max_age=604800,
+        expires=604800,  # expiry set to 7 days,
```
spwoodcock, line 56: Very nice - we just need the frontend implementation now 👍

src/backend/app/main.py
```diff
@@ -30,7 +30,7 @@
 from osm_fieldwork.xlsforms import xlsforms_path
 
 from app.__version__ import __version__
-from app.auth import auth_routes
+from app.auth import auth_routes, temp_auth
```
spwoodcock, line 33: We probably don't need a new router for the single endpoint - should it go under the auth router?

src/backend/app/auth/temp_auth.py
```diff
@@ -0,0 +1,63 @@
...
+router = APIRouter(
+    prefix="/temp",
+    tags=["temp-auth"],
+    responses={404: {"description": "Not found"}},
+)
+
+
+@router.get("/login/")
+async def temp_login(
+    request: Request,
```
spwoodcock, line 34: One last thing: I think we should take an extra param `email` here. 

Thus implementation will work for now. 

When we have an email service included, maybe we ads the extra step to require verification by clicking a button in the users email inbox.

spwoodcock (changes requested):

spwoodcock:
> I have added login_required in update_task_status , is there anything to add?

There is also the GET project endpoint that we should probably put behind login again.

Sujanadh:
> > I have added login_required in update_task_status , is there anything to add?
> 
> There is also the GET project endpoint that we should probably put behind login again.

There is no dependency check in `GET` endpoints of project view; even non-logged-in users are allowed to view projects, right?
Are we now adding login_required to them?

src/backend/app/auth/temp_auth.py
Sujanadh: just an extra param for now?

spwoodcock:
> > > I have added login_required in update_task_status , is there anything to add?
> > 
> > 
> > There is also the GET project endpoint that we should probably put behind login again.
> 
> There is no dependency check in `GET` endpoints of project view; even non-logged-in users are allowed to view projects, right? Are we now adding login_required to them?

In the past we had projects behind a ProtectedRoute in the frontend, requiring login.
I removed it temporarily for field testing, but we probably need it back in, to cover the case of private projects (we can't have users able to access everything by default).

We can add the ProtectedRoute back into the frontend, meaning we probably also want `login_required` on 
https://github.com/hotosm/fmtm/blob/a50540a83a6f194092e052a57cb55edc6849c53c/src/backend/app/projects/project_routes.py#L218

The `project_dashboard` endpoint is also called on the project detail page.
https://github.com/hotosm/fmtm/blob/a50540a83a6f194092e052a57cb55edc6849c53c/src/backend/app/projects/project_routes.py#L1182
This should have `login_required` too, but also should not be called when the user goes to the project details, only when they go to the dashboard (requires frontend change)

Sujanadh:
makes sense :+1:

Sujanadh:
One question. Since private and public access are checked in mapper role but not in login_required, and for the temporary login also we use svcfmtm which is mapper by default. Again, in the future new login from any source will be assigned role `mapper` so better to use `mapper` dependency instead of login_required what do you think?

spwoodcock:
You are completely right!

We need to give this more thought!

We need temp users to be able to access projects, so they have to be assigned mapper role.
As they all use the svcfmtm user, this works fine for public projects. 
We just need to make sure the user svcfmtm exists and has MAPPER role.

I actually think that by default svcfmtm is ADMIN... we need to separate those roles and use two different usernames.

For private projects, we probably need a mechanism for project admins to generate an invite link. This could create a specific user tied to a project with MAPPER permission, so a user must have an invite like to access it.

Open to ideas!

Sujanadh:
Can't we change the role of svcfmtm to mapper? 
Yes, I agree with the idea of sending invites to users via email for the private project since we don't have ui to assign users to the project.

spwoodcock:
After discussion, we decided to:
- Remove ADMIN permission from default svcfmtm user account.
- Create a second user `betauser` that is manager for the FMTM Public Beta org.
  - This user is not an OSM user and is only local in the database.
  - It cannot be logged into.
- Use `svcfmtm` for the temp auth in this PR.

spwoodcock:
I just pushed updates:
- The user `svcfmtm` will have it's role set to 'MAPPER' at startup.
- The user `localadmin` is created at startup and is now the public beta org admin.

spwoodcock:
How are you using localadmin?

Login shouldn't work, except for during local development where login is skipped and localadmin is returned.

Can you check your db? The user should have been created at server startup with id 0 (the field should accept 0 no problem)

Sujanadh:
yes i checked db, it has record of localadmin. but while getting uid during check access in roles.py it doesn't extract id from the user_data though the data have id : 0. instead it returned above error.

spwoodcock:
Nice, if changing to 1 fixed the error, then all good!

spwoodcock (approved):

Issue #1326, opened 2024-03-05
Task IDs should be UUID or similar


spwoodcock:
**Describe the bug**

- Task IDs are currently just an incrementing integer.
- If we didn't use the IDs in code (and used the `project_task_prefix`, this wouldn't be a problem).
- But as things are we will end up with 1000's of tasks and a very large number.

**Expected behavior**

- Field `tasks.id` should be a uuid or similar, so that the task number doesn't matter.

manjitapandey:
The task breakdown:

- [ ] needs change on task ID 
- [ ] Needs updating on QR since  QR is related to task ID
- [ ] needs change on ODK 

Is any more components need to be modified? @spwoodcock

spwoodcock:
There's no need to update the QRCode component, as that only contains the `odk_token` from the database.

However:
- Previously we were creating the xFormId for referencing in ODK from the project name and task id.
- If we change the task id to be the internal task id to a project, this may break out references to ODK.
- As a result, it might be wise to complete https://github.com/hotosm/fmtm/issues/1418 prior to this task.

manjitapandey:
Shifting this issue to the top then. @Sujanadh can you look both of these issue parallelly??

PR #1470, opened 2024-04-22
Project task index as an user facing task_id

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [X] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- issue #1326 

## Describe this PR

- uses project task index to use it as a task id starting from 0 for each individual project
- uses uuid as a XFormId(odk_form_id)
- saves odk_form_id in the database table xforms along with category and project_id
there is still an issue uploading entities in the odk central
need some changes in osm fieldwork for proper log, i.e. xform name is taken from xformid, now it has been changed to uuid ,now need clear distinction between xform name and xformid
## Screenshots

N/A

## Review Guide

### DIFF
--- a/src/backend/app/central/central_crud.py
+++ b/src/backend/app/central/central_crud.py
@@ -20,6 +20,7 @@
 import csv
 import json
 import os
+import uuid
 from io import BytesIO, StringIO
 from typing import Optional, Union
 from xml.etree.ElementTree import Element, SubElement
@@ -211,8 +212,8 @@ def create_odk_xform(
             status_code=500, detail={"message": "Connection failed to odk central"}
         ) from e
 
-    form_name = xform.createForm(odk_id, xform_data, publish=True)
-    if not form_name:
+    xform_id = xform.createForm(odk_id, xform_data, publish=True)
+    if not xform_id:
         namespaces = {
             "h": "http://www.w3.org/1999/xhtml",
             "odk": "http://www.opendatakit.org/xforms",
@@ -229,8 +230,7 @@ def create_odk_xform(
         raise HTTPException(
             status_code=HTTPStatus.UNPROCESSABLE_ENTITY, detail=msg
         ) from None
-
-    return form_name
+    return xform_id
 
 
 def delete_odk_xform(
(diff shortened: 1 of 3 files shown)


spwoodcock:
Apologies Sujan, I have also been working on XForm related code and have created conflicts for you!

Code look good - we should discuss how best to integrate this into the frontend, as it may require updating a significant number of API endpoints for the frontend to pass the project_id+task_numer combo (instead of the `task_id`) 👍

Sujanadh:
Only the user-facing id will be that project-specific id like that in QR pop up. But for the endpoints, we can still use the same id that we are using till now. Using both project_id and task_index won't be that good. And except id, we don't have other unique fields.

Sujanadh:
So, we can limit the use of both project_id and task index only in submissions endpoints to interact with odk central.

spwoodcock (approved):

spwoodcock:
You are definitely right - the `project_id` and `task_index` combo isn't a good solution!

Looks like the task details from the API return both the id and index, so the best solution is probably to have both in place:
image

We use the `task.id` for referencing via the API, and use the index for user facing things.
Nice update to the data extract task_id by the way 😄 
I will update the frontend to show the index, then merge this PR.

spwoodcock:
From a brief look I need to update:

- Submission page:
    image
    
    image

- Project detail page:
    
    image

spwoodcock:
All done - could you review this works ok @Sujanadh please? 🙏

Sujanadh:
fixed the entity not being uploaded problem

Sujanadh:
> All done - could you review this works ok @Sujanadh please? 🙏

looks good so far

Sujanadh:
One thing I noticed quickly was the order of task completion.
Screenshot from 2024-04-24 19-21-18
they are not in ascending order.

spwoodcock:
Nice work - thanks! 🎉

PR #1473, opened 2024-04-24
Replace popup auth flow with redirect flow

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR

- Replaces popup based OSM login with a redirect flow (as popup is quite clunky).
- The redirect returns to `/osmauth` which handles getting the cookie from the backend and then redirecting again to the home page.
- During local development there is a hardcoded additional redirect that is required due to https://github.com/openstreetmap/openstreetmap-website/pull/4287
  - We accept the redirect from OSM to `http://127.0.0.1:7051` and then pass through the URL params (containing the authorisation code) to `http://fmtm.localhost:7050` for a seamless developer experience.
  - In production this additional redirect is not required.
  - If OSM accept the PR to use `localhost` in the OAuth redirect_uri, then we can remove this additional redirect, in addition to removing the port binding for 7051 on the frontend container.
- I also refactored / simplified the loginSlice store, and changed a few variable names for clarity.

@manjitapandey 

## Screenshots

image

- The popup based approach was slick using window.postMessage to avoid the restrictions imposed by OSM, but this is a nicer user experience IMO.

## Review Guide

- Login with both OSM and temp auth to test.


src/frontend/src/routes.jsx
varun2948, line 1: we can integrate a boolean value and use protectedroute component just based on that boolean value but going ahead on this PR for now

spwoodcock: That would be nice - we can add that in after 👍 

Something else that would be very cool is replacing `return <Navigate to="/" replace />;` in `ProtectedRoute` with something else to prevent calling the API `/summaries` endpoint on the homepage again.

Ideally we only want the login selection popup to show, but I couldn't work out how to do that! (probably using the approach you mention with a flag!)

Issue #1499, opened 2024-04-30
Form submissions bar chart not visible although submissions are made 


NSUWAL123:
**Describe the bug**
Although submissions are made via ODK, it is not reflected on the char since empty list is returned by the `/submission/submission_page/49?days=10` endpoint.

**To Reproduce**

1. Open network settings tab
2. Go to `project-submissions` page
3. Click the above mentioned API 
4. See empty list being returned

**Expected behavior**
If submissions are made, the endpoint should return submissions data.

**Screenshots**
![image](https://github.com/hotosm/fmtm/assets/81785002/5a86d127-ad66-4e42-95d0-f75a2e27381f)

PR #1509, opened 2024-05-08
Submissions in project dashboard and bar graph.

Sujanadh:

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation
- [X] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related Issue #1499

## Describe this PR

This PR updates the project dashboard endpoint to correctly fetch the submission metadata from odk central which used to depend on the cached submissions in s3 that is not required now onwards as we can directly get it from odk central.

## Screenshots

image

## Review Guide

### DIFF
--- a/src/backend/app/projects/project_routes.py
+++ b/src/backend/app/projects/project_routes.py
@@ -64,7 +64,6 @@ from app.models.enums import (
 )
 from app.organisations import organisation_deps
 from app.projects import project_crud, project_deps, project_schemas
-from app.submissions import submission_crud
 from app.tasks import tasks_crud
 
 router = APIRouter(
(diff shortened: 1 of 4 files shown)


src/backend/app/projects/project_routes.py
```diff
@@ -1224,17 +1221,7 @@ async def project_dashboard(
     Returns:
         ProjectDashboard: The project dashboard details.
     """
-    data = await project_crud.get_dashboard_detail(db_project, db_organisation, db)
-
-    background_task_id = await project_crud.insert_background_task_into_database(
```
nrjadkry, line 1229: We are not saving the submission in s3 cache anymore?
Is fetching the data from central api everytime we need the submissions efficient ?  We should think of the case when there are thousands of submission json.

Sujanadh: yes we don't need to save the submissions for regular fetch we can do that directly from central, but we might have to consider it when downloading submissions , need to check the performance for that. if it is considerably good then no more s3 if not we not think about it.

Sujanadh: Later, if the performance doesn't meet expectations, we can create a separate endpoint to save submissions in s3 instead of running background tasks in every endpoint, which can be hit from the frontend whenever someone loads the submission table page.

nrjadkry: Okay, I am merging this PR for now.

Issue #1508, opened 2024-05-08
State loss whenever the submission page is reloaded.


Sujanadh:
**Describe the bug**
If the submission page or submission table is reloaded , the task list in both submission page and drop down in table doesn't appear.

**submission page**
![image](https://github.com/hotosm/fmtm/assets/109404840/21fb0202-ed81-4d91-96cb-6d38cdf66f97)

**submission table**
![Screenshot from 2024-05-08 17-33-37](https://github.com/hotosm/fmtm/assets/109404840/2fb5cbc5-4e40-49d3-a2c6-ebae84538079)

**To Reproduce**

1. Go to view infographics'
2. Reload the page
3. Scroll down to bottom
4. See blank
5. Click on table tab
6. Reload the page
7. Click on filter
8. Click on task id filter
9. See empty drop down list

**Expected behavior**
Task list shouldn't be lost even if the page is reloaded.

PR #1513, opened 2024-05-09
Fix/frontend state api issues

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #1508

## Describe this PR
This PR fix the following issues: 
1. Profile image not displaying due to incorrect image key
2. Relplace taskId with taskIndex on `Zoom to Task` click resolving page crash issue
3. UpdateReviewStateModal modal not opening issue solve by replacing `decodedProjectId` with `projectId`
4. Show default task review state to `recieved` if no `reviewState` is present
5. `/statuses` API fetch on projectSubmission initial render

## Screenshots
image
image

### DIFF
--- a/src/frontend/src/components/DialogTaskActions.jsx
+++ b/src/frontend/src/components/DialogTaskActions.jsx
@@ -30,20 +30,22 @@ export default function Dialog({ taskId, feature, map, view }) {
   const projectIndex = projectData.findIndex((project) => project.id == currentProjectId);
   const currentStatus = {
     ...taskBoundaryData?.[projectIndex]?.taskBoundries?.filter((task) => {
-      return task.id == taskId;
+      return task?.index == taskId;
     })?.[0],
   };
   const projectTaskActivityList = CoreModules.useAppSelector((state) => state?.project?.projectTaskActivity);
 
   useEffect(() => {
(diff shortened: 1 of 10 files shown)


NSUWAL123:
NOTE:
@spwoodcock, it seems that you removed the `task-submission`  API from front end as well as back end which is leading to issues.
1. Task List is being retrieved from that `task-submission` API. Currently, taskList is displayed if the user navigates to the projectSubmission page from projectDetails page. But if the user reloads on projectSubmissions page or goes directly to that page the taskList is not being displayed since the `/statuses` API is not being hit.
image
To solve this issue, I have included fetching `/statuses` API on projectSubmissions page as well. This is not an ideal solution but it's for a temporary fix. I don't see an alternative solution rather that using the previous `/task-completion` API on the project submission page. The `projects/:id` is also fetched on this page that returns `geojson`, `feature_count` but it doesn't return `submission_count`.

CC: @varun2948 @Sujanadh @nrjadkry

Issue #1539, opened 2024-05-26
Used JWT-based authentication regardless of OAuth method


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- For a while we used OSM login only, via osm-login-python.
  - This library takes the OAuth token from OSM, encrypts it with a SECRET_KEY, then provides it to the user.
  - This encrypted token can only be decrypted by osm-login-python (using the same SECRET_KEY), in FastAPI we deserialise the token to OSM details (username, profile picture, etc) via a `login_required` dependency (Depends).
- We have since added a temporary authentication for mappers.
  - This also uses OSM authentication, but instead uses a service account `svcfmtm`.
  - Each user with a temporary login token is technically logged in as svcfmtm.
  - This works fine, but isn't ideal for self-hosted FMTM instances, requiring additional work to set it up (creating an OSM service account, getting the OAuth token).
- It might be better to encapsulate the OSM login in a standardised JWT-based authentication, which will remain the same regardless of login method.
  - This way we could also add other login providers (e.g. Google), although is not preferred as login methods outside of OSM do not provide the user attribution for their edits.

**Describe the solution you'd like**

- Implement JWT-based authentication using [pyjwt](https://github.com/jpadilla/pyjwt/).
- ES256 would be the preferred encrpytion algorithm due to smaller size and better security than RS256 (also asymmetric, pub/priv key pair). However, pyjwt has limited curves available for ES256 (secp256k1 only), which is quite limiting. RS256 is acceptable for our use case, secure until at 2030-2050 with computing advances.
  - First a keypair must be generated:
      ```bash
      $ openssl genrsa -out private_key.pem 4096
      $ openssl rsa -in private_key.pem -pubout -out public_key.pem
      ```
  - The algorithm can be specified in pyjwt: https://pyjwt.readthedocs.io/en/latest/usage.html#encoding-decoding-tokens-with-rs256-rsa
- The payload should include the following [claims](http://www.iana.org/assignments/jwt/jwt.xhtml):

```json
{
"sub": "fmtm|userid",  // iss claim not required if included as unique key 'fmtm|' in sub
"aud": "fmtm.hotosm.org",  // change domain for each instance
"iat": "1716679595"  // UTC Unix token issue time (seconds from the Unix epoch)
"exp": "1716683195",  // UTC Unix token expiry time (seconds from the Unix epoch)
"name": "username (i.e. OSM username)",  // public claim
"email": "email if provided",  // public claim
[...]

PR #1574, opened 2024-06-13
Implemented jwt token authentication for login

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- issue #1539 

## Describe this PR

This PR adds new configuration to generate jwt authentication token for temporary login as well as to encapsulate the osm login for consistency. For token generation, the "RS256" algorithm is used, whereas rsa keys are generated for encoding and decoding the token.

## Screenshots

N/A

## Review Guide

### DIFF
--- a/src/backend/tests/test_projects_routes.py
+++ b/src/backend/tests/test_projects_routes.py
@@ -48,7 +48,7 @@ async def test_create_project(client, admin_user, organisation):
         "odk_central_user": odk_central_user,
         "odk_central_password": odk_central_password,
     }
-    odk_credentials = project_schemas.ODKCentralDecrypted(**odk_credentials)
+    odk_creds_models = project_schemas.ODKCentralDecrypted(**odk_credentials)
 
     project_data = {
         "project_info": {
@@ -71,7 +71,7 @@ async def test_create_project(client, admin_user, organisation):
             "type": "Polygon",
         },
     }
-    project_data.update(**odk_credentials.model_dump())
+    project_data.update(**odk_creds_models.model_dump())
 
     response = client.post(
         f"/projects/create_project?org_id={organisation.id}", json=project_data


src/backend/app/auth/auth_routes.py
```diff
@@ -83,19 +84,27 @@ async def callback(request: Request, osm_auth=Depends(init_osm_auth)):
     log.debug(f"Access token returned of length {len(access_token)}")
     response = Response(status_code=HTTPStatus.OK)
 
+    osm_user = osm_auth.deserialize_access_token(access_token)
+    user_data = {
+        "sub": f"fmtm|{osm_user['id']}",
+        "aud": "fmtm.localhost:8000",
```
spwoodcock, line 90: The actual audience via the proxy is fmtm.localhost:7050

You can build this using the FMTM_DOMAIN var and DEV_PORT vars, see a similar approach for building the URL here: https://github.com/hotosm/fmtm/blob/b143186a9368b41c8fa7a2c122e6dea003e5b0db/src/backend/app/config.py#L173

Hopefully if the audience matches from serialisation to deseralisation, this should work fine 👍

src/backend/app/auth/auth_routes.py
```diff
@@ -83,19 +84,27 @@ async def callback(request: Request, osm_auth=Depends(init_osm_auth)):
...
+        "iat": int(time.time()),
+        "exp": int(time.time()) + 86400,
```
spwoodcock, line 92: Perfect 👌

src/backend/app/config.py
```diff
@@ -275,7 +275,38 @@ def set_raw_data_api_auth_none(cls, v: Optional[str]) -> Optional[str]:
...
+    AUTH_PRIVATE_KEY_PATH: str
+    AUTH_PUBLIC_KEY_PATH: str
+    AUTH_PRIVATE_KEY: Optional[str] = None
+    AUTH_PUBLIC_KEY: Optional[str] = None
+
+    @field_validator("AUTH_PRIVATE_KEY", mode="before")
+    @classmethod
+    def load_private_key(cls, v: Optional[str], info: ValidationInfo) -> str:
+        """Loads the private key for authentication."""
+        if v:
```
spwoodcock, line 288: Shouldn't this be `if v is None` or am I missing something?

We only load from the file if the variable is not set directly.

spwoodcock (commented):
Very nice code!

I think the only thing missing is modifying the introspect endpoint to deserialise the access token and check that it has not expired

src/backend/app/auth/auth_routes.py
Sujanadh: ah crap I used that to test audience , but pushed it mistakenly. Thanks for pointing it out.

src/backend/app/config.py
Sujanadh: Yes, you are actually right.

Sujanadh:
> Very nice code!
> 
> I think the only thing missing is modifying the introspect endpoint to deserialise the access token and check that it has not expired.
> 
> Also the tests are failing and may need to be looked at 😁

introspect depends on `login_required` and expiry check is done in `login_required`.

src/backend/app/auth/auth_routes.py
Sujanadh: I found that audience was failing because I failed to mention it while decoding the encoded token.
`audience = settings.FMTM_DOMAIN` like this while decoding.

Sujanadh:
Since a temporary user is not a real user and is not present in a database, we need to create it in the database, so for that, we need to explicitly specify the fields like id and other fields like profile_img as None.

Sujanadh:
I think test was failing because env file is not updated with variables like `AUTH_PRIVATE_KEY` .

spwoodcock:
Thanks for the feedback - was reviewing on my phone again and missed the login_required link 😁


We already have user svcfmtm in the db - could we use that for the temp user possibly?

I'll update the env vars for the tests 👍

Sujanadh:
> Thanks for the feedback - was reviewing on my phone again and missed the login_required link 😁
> 
> We already have user svcfmtm in the db - could we use that for the temp user possibly?
> 
> I'll update the env vars for the tests 👍

should we use username "svcfmtm" also ?

spwoodcock:
> > Thanks for the feedback - was reviewing on my phone again and missed the login_required link 😁
> > We already have user svcfmtm in the db - could we use that for the temp user possibly?
> > I'll update the env vars for the tests 👍
> 
> should we use username "svcfmtm" also or just the id ?

Probably good to just hardcode `svcfmtm` as the user.
We can then pick up that username/id during data conflation and instead insert the username of the project manager for OSM submissions.

spwoodcock:
This is pretty much good to go, but I will wait to merge until we finalise the 2024.3.1 release.

Todo:
- Update hardcoded username to svcfmtm
- Update env config throughout repo:
  - [ ] Tests
  - [ ] Example dotenv
  - [ ] Docs
  - [ ] FMTM Installer
  - [ ] Usage in Justfiles etc

spwoodcock:
In addition to the above @Sujanadh and I just discussed the possibility of using JWT refresh tokens to keep the token alive.

If the user is inactive for many days it will expire, but if accessed regularly the token will refresh.

Seems like a good idea to me 👍

src/backend/app/auth/auth_routes.py
```diff
@@ -81,28 +91,21 @@ async def callback(request: Request, osm_auth=Depends(init_osm_auth)):
...
-        domain=settings.FMTM_DOMAIN,
-        secure=False if settings.DEBUG else True,
-        httponly=True,
-        samesite="lax",
-    )
-    return response
+    osm_user = osm_auth.deserialize_access_token(access_token)
+    user_data = {
+        "id": osm_user["id"],
+        "sub": f"fmtm|{osm_user['id']}",
```
spwoodcock, line 98: Can the sub be changed to `osm|xxx` here?

The temp auth would be something like `fmtm|temp`

src/backend/app/auth/auth_routes.py
```diff
@@ -81,28 +91,21 @@ async def callback(request: Request, osm_auth=Depends(init_osm_auth)):
...
-        path="/",
-        domain=settings.FMTM_DOMAIN,
-        secure=False if settings.DEBUG else True,
-        httponly=True,
-        samesite="lax",
-    )
-    return response
+    osm_user = osm_auth.deserialize_access_token(access_token)
+    user_data = {
+        "id": osm_user["id"],
```
spwoodcock, line 97: Just checking, do we still need id, now we have sub?

src/backend/app/auth/auth_routes.py
```diff
@@ -81,28 +91,21 @@ async def callback(request: Request, osm_auth=Depends(init_osm_auth)):
...
+        "sub": f"fmtm|{osm_user['id']}",
+        "aud": settings.FMTM_DOMAIN,
+        "iat": int(time.time()),
+        "exp": int(time.time()) + 86400,  # expiry set to 1 day
+        "username": osm_user["username"],
+        "email": osm_user.get("email"),
+        "img_url": osm_user.get("img_url"),
```
spwoodcock, line 104: I think the standard oauth claim for this is `picture` - worth double checking!

src/backend/app/auth/auth_routes.py
```diff
@@ -248,33 +270,16 @@ async def temp_login(
...
-    return response
+    username = "svcfmtm"
+    user_data = {
+        "id": 20386219,
+        "sub": f"fmtm|{username}",
+        "aud": settings.FMTM_DOMAIN,
+        "iat": int(time.time()),
+        "exp": int(time.time()) + 86400 * 7,  # expiry set to 7 days
+        "username": username,
+        "img_url": None,
```
spwoodcock, line 281: Comments to osm jwt also apply here 👍

src/backend/app/auth/auth_routes.py
```diff
@@ -81,28 +91,21 @@ async def callback(request: Request, osm_auth=Depends(init_osm_auth)):
...
-        max_age=31536000,  # OSM currently has no expiry
-        expires=31536000,  # OSM currently has no expiry
-        path="/",
-        domain=settings.FMTM_DOMAIN,
-        secure=False if settings.DEBUG else True,
-        httponly=True,
-        samesite="lax",
-    )
-    return response
+    osm_user = osm_auth.deserialize_access_token(access_token)
```
spwoodcock, line 95: Do we need error handling here of the OSM token fails to deserialise? That would essentially mean OSM login failed and we should not proceed

src/backend/app/auth/auth_routes.py
```diff
@@ -248,33 +270,16 @@ async def temp_login(
...
-        max_age=604800,
-        expires=604800,  # expiry set to 7 days,
-        path="/",
-        domain=settings.FMTM_DOMAIN,
-        secure=False if settings.DEBUG else True,
-        httponly=True,
-        samesite="lax",
-    )
-    return response
+    username = "svcfmtm"
```
spwoodcock, line 273: We can probably remove reference to svcfmtm in this token, as it's no longer an osm account, just a temp jwt

spwoodcock (changes requested):
Some small comments. 

Nice job keeping the code DRY 👍

src/backend/app/auth/auth_routes.py
Sujanadh: yes we need it for `me` endpoint

src/backend/app/auth/auth_routes.py
Sujanadh: We talked about this previously in the above comment to hard-code the username to `svcfmtm`. If not, we need to create a temporary user.

src/backend/app/auth/auth_routes.py
Sujanadh: its img_url,
```json
{'id': 18427944, 'username': 'Sujan Adhikari', 'img_url': 'https://www.openstreetmap.org/rails/active_storage/representations/redirect/eyJfcmFpbHMiOnsiZGF0YSI6MzI4NTE3MTksInB1ciI6ImJsb2JfaWQifX0=--6b73d47fac28854f227f3d0ecd56f2f6554c8709/eyJfcmFpbHMiOnsiZGF0YSI6eyJmb3JtYXQiOiJwbmciLCJyZXNpemVfdG9fbGltaXQiOlsxMDAsMTAwXX0sInB1ciI6InZhcmlhdGlvbiJ9fQ==--473f0190a9e741bb16565db85fe650d7b0a9ee69/sujan.webp'}
```

src/backend/app/auth/auth_routes.py
Sujanadh: We are deserializing to get the user data from osm and encode it again using jwt. What do you mean by proceeding after osm login failed? How should we proceed even after osm login fails? If the user fails osm login, they can choose a temporary login.

src/backend/app/auth/auth_routes.py
Sujanadh: `osm|username` or `osm|id` ?

src/backend/app/auth/auth_routes.py
spwoodcock: Oh of course, my bad!
I shouldn't do a review after 3 hours sleep 😂

src/backend/app/auth/auth_routes.py
spwoodcock: https://www.iana.org/assignments/jwt/jwt.xhtml

The standard claim in the JWT spec is `picture`

`img_url` would be a custom claim

src/backend/app/auth/auth_routes.py
spwoodcock: The sub should always have the id, so good point what I wrote for the temp auth is inconsistent. 

osm|123 or fmtm|123

We have a username claim for the username

src/backend/app/auth/auth_routes.py
spwoodcock: But only because the frontend expects an `id` field right?

We could change the frontend, but a nicer approach might be to use a computed property on the pydantic model returned. 

We could have the `id` field computed by splitting sub on | and then taking the second value

src/backend/app/auth/auth_routes.py
spwoodcock: If the token fails to deserialise then we get two scenarios:
- the JWT is generated with empty OSM details
- a possible 500 response

Neither scenario is great. We should handle failure gracefully and return a HTTPError

src/backend/app/auth/auth_routes.py
Sujanadh: So if we aren't using svcfmtm, then we need a temporary user profile and its id to use as sub.

spwoodcock:
I'll work on the todo actions on Monday 👍 

Todo:
- [X] Update hardcoded username to svcfmtm
- Update env config throughout repo:
  - [ ] Tests
  - [ ] Example dotenv
  - [ ] Docs
  - [ ] FMTM Installer
  - [ ] Usage in Justfiles etc

spwoodcock:
Did a bit of additional refactoring and finished off the roles!

Still need to:
- [ ] Set additional env vars in Github environments
- [ ] Fix test_create_project that is showing 400 project with name already exists??

Issue #1461, opened 2024-04-18
Add endpoint to refresh the ODK appuser token for a project


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- Currently the appuser token is generated once during project creation.
- This is what is embedded in the QRCode and links to a project from ODK Collect.
- There is no way to refresh this token - for example if it is suspected the token is compromised.

**Describe the solution you'd like**

- Add an endpoint that 'refreshes' the appuser token.
- One way to achieve this would be:
  - Delete the appuser (which will delete the permissions to project forms too).
  - Create a new appuser.
  - Link the appuser to the project forms.

PR #1583, opened 2024-06-17
Added new endpoint to refresh the app user token

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [X] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue No. #1461 

## Describe this PR

- This PR creates a new endpoint `refresh_appuser_token` which allows to create an app user if it has not created already , if it exists then it deletes it and creates again and updates the `odk_token` in the database.
- refactors existing code to create app user during project file generation to use this new function.

## Screenshots

N/A

## Review Guide

### DIFF
--- a/src/backend/app/central/central_routes.py
+++ b/src/backend/app/central/central_routes.py
@@ -17,7 +17,7 @@
 #
 """Routes to relay requests to ODK Central server."""
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, HTTPException
 from fastapi.concurrency import run_in_threadpool
 from fastapi.responses import JSONResponse
 from sqlalchemy.orm import Session
@@ -25,6 +25,7 @@ from sqlalchemy.orm import Session
 from app.auth.roles import project_admin
 from app.central import central_crud
 from app.db import database, db_models
+from app.models.enums import HTTPStatus
 from app.projects import project_deps
 
 router = APIRouter(
@@ -58,9 +59,8 @@ async def get_form_lists(
     return forms
 
 
-@router.post("/refresh_appuser_token/{project_id}")
+@router.post("/refresh_appuser_token")
 async def refresh_appuser_token(
-    project_id: int,
     current_user: db_models.DbUser = Depends(project_admin),
     db: Session = Depends(database.get_db),
 ):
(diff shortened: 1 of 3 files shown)


src/backend/app/central/central_routes.py
```diff
@@ -54,3 +56,35 @@ async def get_form_lists(
...
+    Args:
+        project_id (int): The ID of the project to refresh the app user token for.
+        current_user: The current authenticated user with project admin privileges.
+        db: The database session to use.
+
+    Returns:
+        The refreshed app user token.
+    """
+    odk_credentials = await project_deps.get_odk_credentials(db, project_id)
+    project = await project_deps.get_project_by_id(db, project_id)
```
spwoodcock, line 78: Could it be possible to return both the user and the project in a single dict / Depends here, like some of the other endpoints?

It be nice to do in a single transaction instead of two, if possible 😃

src/backend/app/central/central_routes.py
```diff
@@ -54,3 +56,35 @@ async def get_form_lists(
...
+    project_odk_id = project.odkid
+    db_xform = await project_deps.get_project_xform(db, project_id)
+    odk_token = await central_crud.get_appuser_token(
+        db_xform.odk_form_id, project_odk_id, odk_credentials, db
+    )
+    if odk_token:
+        project.odk_token = odk_token
+    db.commit()
+    return {
```
spwoodcock, line 87: We return a success even if no token is generated. 

Could we handle if there is an error too, and return a HTTPException perhaps?

src/backend/app/central/central_routes.py
```diff
@@ -54,3 +56,35 @@ async def get_form_lists(
...
+        "status_code": 200,
```
spwoodcock, line 88: There id an enum available that would be good to use here:

200 --> HTTPStatus.OK (or maybe SUCCESS instead, I forget now)

spwoodcock (changes requested):
Looks good!

Great call reusing the same code during project creation too 👍

Sujanadh:
Changes:
- raise exception when app user fails to refresh
- reduce db transaction using dependency `current_user` to retrieve user and project from single transaction

Issue #1548, opened 2024-06-05
Create endpoint for conflation of FMTM data with existing OSM data


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- ~~https://github.com/hotosm/conflator requires OSM data in a local postgres instance.~~
- We need an endpoint that can conflate FMTM submissions against existing OSM data.
- It should take the final submission geojson from FMTM, including geometries and tags.
- The conflation code should ideally return the certainty of the conflation:
  - No overlap (new): add geometry and tags with **certain** flag.
  - Exact building overlap: conflate tags and return geom, with **certain** flag.
  - Partial building overlap: return both geometries, including conflated tags on both, with **uncertain** flag.
- The resulting GeoJSON will have geometries and tags, with a certainty attached, ready for manual validation via the FMTM frontend.

**Describe the solution you'd like**

- ~~Option 1: host our own OSM database with updates (not ideal, wasted money), and create an endpoint.~~
- ~~Option 2: integrate the conflation with raw-data-api, that already has an existing postgres instance with updated OSM data (i.e. a new endpoint).~~

Discussing the long term solution with @kshitijrajsharma over the coming weeks.
The most achievable option now is something like https://github.com/kshitijrajsharma/osmconflator/blob/3ecf12b0d31773750cf1f806fd4547532f19d046/osmconflator/utils.py#L18

Process:
- User requests conflation
- Latest OSM data for the AOI is downloaded via raw-data-api
  - The data should include all geometry types and all tags.
  - The data is used for conflation against the FMTM geometries and tags.
- The endpoint can return the certainties of conflation as described above.
- The returned GeoJSON is used for visual validation, as described in the follow on issues in this milestone.

**Additional considerations**

- How do we handle the case when the offset is so large that there is no building overlap (for matching buildings).
- Is this a likely scenario?

spwoodcock:
For now this should be adapted from: https://github.com/kshitijrajsharma/osmconflator/blob/3ecf12b0d31773750cf1f806fd4547532f19d046/osmconflator/utils.py#L18

- This would provide very basic conflation of the geometries.
- At first, we could just merge all the tags.
- The next step would be tag conflation, but most likely this will be handled in the near future in hotosm/conflator.

spwoodcock:
More context on this after a lengthy internal discussion with team members.

### OSM Refs & Raw-Data-API

[...]

PR #1594, opened 2024-06-21
Endpoint to conflate the submission with osm data

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue No. #1548 

## Describe this PR

This PR creates a new endpoint in order to conflate the submission data with osm features.
- if both geometries (buildings for now) are equal or almost equal, then only their tags/properties are appended and returned for further validation
- if both geometries overlaps to some extent then in that case both geometries are returned along with their properties and tags appended
- is_duplicate and is_partial_overlap boolean flags (True or False) are appended, respectively.
-
## Response

```python
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              85.297789,
              27.723315
            ],
            [
              85.297842,
              27.723297
            ],
            [
              85.297856,
              27.723329
            ],
            [
              85.297908,
              27.723312
            ],
            [
              85.297924,
              27.723351
            ],
            [
              85.297819,
              27.723386
            ],
            [
              85.297789,
              27.723315
            ]
          ]
        ]
      },
      "properties": {
        "start": "2024-06-11T14:01:37.472+05:45",
        "end": "2024-06-11T14:02:11.013+05:45",
        "today": "2024-06-11",
        "phonenumber": null,
        "deviceid": "collect:N9J5nVWtA9ApE60K",
        "username": "localadmin",
        "email": "NOT IMPLEMENTED",
        "instructions": null,
        "task_id": null,
        "existing": "be3d424a-a5a4-4544-a0c5-3b38dd522dd7",
        "point": null,
        "digitisation_correct": "yes",
        "digitisation_problem": null,
        "digitisation_problem_other": null,
        "image": null,
        "form_category": "Unknown",
        "xid": "650020468",
        "status": "2",
        "category": "housing",
        "name": null,
        "building_material": null,
        "building_levels": null,
        "religion": null,
        "denomination": null,
        "muslim": null,
        "hindu": null,
        "religious": null,
        "service": null,
        "Shop": null,
        "tourism": null,
        "education_details": null,
        "link": null,
        "government": null,
        "medical": null,
        "food": null,
        "cuisine": null,
        "heritage": null,
        "emergency": null,
        "operator": null,
        "inscription": null,
        "housing": null,
        "provider": null,
        "rooms": null,
        "ref": null,
        "building_neighbor": null,
        "power": null,
        "generator_source": null,
        "generator": null,
        "water": null,
        "water_source": null,
        "pump_unit": null,
        "capacity": null,
        "age": null,
        "building_prefab": null,
        "building_floor": null,
        "building_roof": null,
        "condition": null,
        "access_roof": null,
        "levels_underground": null,
        "comment": null,
        "osm_id": 650020468,
        "osm_type": "ways_poly",
        "version": 2,
        "tags": {
          "building": "yes"
        },
        "changeset": 98649258,
        "timestamp": "2021-02-03T14:09:58",
        "is_duplicate": true,
        "is_partial_overlap": false
      }
    }
  ]
}
```

## Review Guide


src/backend/app/db/postgis_utils.py
```diff
@@ -777,3 +785,160 @@ def parse_featcol(features: Union[Feature, FeatCol, MultiPolygon, Polygon]):
     elif isinstance(features, Feature):
         feat_col = geojson.FeatureCollection([feat_col])
     return feat_col
+
+
+def request_snapshot(geometry):
```
spwoodcock, line 790: This function and the `poll_task_status` function can be replaced with osm_fieldwork.PostgresClient.
There is already functionality in place to get a data extract using provided config 👍 
Example: https://github.com/hotosm/fmtm/blob/319f65e61e67f7cb4819129f1141703e12e31dad/src/backend/app/projects/project_crud.py#L449

src/backend/app/submissions/submission_crud.py
```diff
@@ -524,3 +530,45 @@ async def get_submission_detail(
         odk_form.getSubmissions(project.odkid, db_xform.odk_form_id, submission_id)
     )
     return submission.get("value", [])[0]
+
+
+async def get_submission_geojson(
```
spwoodcock, line 535: I noticed we already have central_crud.convert_odk_submission_json_to_geojson and submission_routes.download_submission_geojson that all basically do the same thing.

Could we refactor to only have the logic once, then reuse the code?

src/backend/app/db/postgis_utils.py
```diff
@@ -777,3 +785,160 @@ def parse_featcol(features: Union[Feature, FeatCol, MultiPolygon, Polygon]):
...
+    Returns:
+        bool: True if geometries have a partial overlap, else False.
+    """
+    intersection = geom1.intersection(geom2)
+    return not intersection.is_empty and (
+        0 < intersection.area < geom1.area and 0 < intersection.area < geom2.area
+    )
+
+
+def conflate_features(
```
spwoodcock, line 896: Nice!

spwoodcock (changes requested):

spwoodcock:
In the long term we should look into integrating this: https://github.com/hotosm/conflator/pull/9 👍

src/backend/app/submissions/submission_crud.py
Sujanadh: if i want geojson only for the conflation, then this function won't work since it returns submission in bytes. like you said, we can refactor the `download_submission` endpoint to use `convert_odk_submission_json_to_geojson`.

spwoodcock: The second option sounds best in the spirit of DRY code.

A single function to return a geojson, then encode as bytes where specifically needed.

If absolutely necessary it's also possible to decode bytes back to geojson using `.decode("utf--8")` 👍

json.load would probably also do that

Sujanadh:
updates:
- used PostgresClient to extract features from osm.
- used task outline instead of bbox of submitted feature.
- updated the `convert_odk_submission_json_to_geojson` to return geojson and respective calls to convert geojson to bytes

src/backend/app/db/postgis_utils.py
```diff
@@ -785,3 +791,124 @@ def parse_featcol(features: Union[Feature, FeatCol, MultiPolygon, Polygon]):
...
+        extract_config,
+        auth_token=settings.RAW_DATA_API_AUTH_TOKEN
+        if settings.RAW_DATA_API_AUTH_TOKEN
+        else None,
+    )
+    return pg.execQuery(
+        geometry,
+        extra_params={
+            "fileName": (
+                f"fmtm/{settings.FMTM_DOMAIN}/data_extract"
```
spwoodcock, line 823: Minor thing - the folder structure starting with `fmtm/...` should only be used if we want to keep the data extract.

Raw data api has a rule if we use auth and a set structure, the file is retained for the project. 

In this case we could remove this param as we only need a temp extract 👍

src/backend/app/db/postgis_utils.py
```diff
@@ -785,3 +791,124 @@ def parse_featcol(features: Union[Feature, FeatCol, MultiPolygon, Polygon]):
...
+                if settings.RAW_DATA_API_AUTH_TOKEN
+                else "fmtm_extract"
+            ),
+            "outputType": "geojson",
```
spwoodcock, line 827: I believe bind zip true and geojson are defaults on raw data api, but no harm being explicit 👍

spwoodcock (approved):

src/backend/app/db/postgis_utils.py
Sujanadh: okay

src/backend/app/db/postgis_utils.py
Sujanadh: I think its better to mention, as we won't know what are the default values.

Sujanadh:
Updates:
- bypass check for public project in wrap_check_access
- removed filename parameter in extra params when requesting data extracts as we don't need it since it is only temp extracts and we don't have to save it in fmtm

src/backend/app/auth/roles.py
```diff
@@ -242,6 +242,12 @@ async def wrap_check_access(
     role: ProjectRole,
 ) -> ProjectUserDict:
     """Wrap check_access call with HTTPException."""
+    # If project is public, skip permission check
+    if project.visibility == ProjectVisibility.PUBLIC:
```
spwoodcock, line 246: Adding this logic here means that all users have access to all endpoints if the project is public.

So they could delete the project, modify anything, etc.

It's currently only present in the `mapper` permission, as we only want to bypass the auth if the endpoint requires `mapper` role only:

https://github.com/hotosm/fmtm/blob/fd11d08f9c5c5c88c6d4b1c7394e517649f0c628/src/backend/app/auth/roles.py#L327

Was this logic needed for the code in this PR to work?

spwoodcock (changes requested):

src/backend/app/auth/roles.py
spwoodcock: Everything else is all set to merge!

spwoodcock: I did a minor update, fixing the return type on the `mapper` role if `ProjectVisibility.PUBLIC`.

I'm going to merge as this PR has been open a while - let me know if any issues!

Sujanadh: Uh I totally missed that. Yes , you are right. I changed that in order to get ProjectUserDict.

Sujanadh: Uh I totally missed that. Yes , you are right. I changed that in order to get ProjectUserDict.

Issue #1599, opened 2024-06-27
Map not being displayed


NSUWAL123:
**Describe the bug**
Currently, the Mapbox layer is being used as a default baselayer. It seems the token for the Mapbox baselayer is expired, the base layer is not visible. Until Mapbox baselayer is visible, use OSM baselayer. 

**To Reproduce**

1. Go to 'FMTM.'
2. See the map section showing a blank white screen.

**Expected behavior**

**Screenshots**
![image](https://github.com/hotosm/fmtm/assets/81785002/3416b66e-4dc2-45d3-953d-448af2fc89bd)

PR #1600, opened 2024-06-27
fix(map): set default baseLayer to OSM

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #1599

## Describe this PR
This PR replaces the current default baselayer(Mapbox) to OSM to solve the map not visible issue.

## Screenshots
image


spwoodcock:
@varun2948 can you remember why we used a mapbox layer instead of direct OSM?

Is it so we can add our own content on top at some point?

spwoodcock:
This will affect dev stage and prod.

I suggest we:
- Renew the token a last time if possible (so we don't need a hotfix to prod).
- Then push this code through to replace using mapbox in the next release.

Issue #1614, opened 2024-07-01
Add functional pytest tests for main mapping endpoints


spwoodcock:
**Is your feature request related to a problem? Please describe.**

Key endpoints for mappers are:

- [ ] ~GET auth/introspect~ not tested as requires frontend to handle cookies etc
- [x] GET projects/summaries
- [x] GET projects/{project_id}
- [x] GET projects/{project_id}/entities/statuses
- [x] GET tasks/{task_id}/history
- [x] POST tasks/{task_id}/new-status/{status_number}
- [x] POST projects/{project_id}/entity/status

**Describe the solution you'd like**

Write tests for the above endpoints.

azharcodeit:
This issue will be resolved once the rest of the endpoints are covered with tests that require an access token:

- GET authentication/introspection 
- POST tasks/{task_id}/new-status/{status_number} 

As soon as the access token is figured out, I will prepare the corresponding PRs. 🚀

spwoodcock:
If the introspect one takes a while to work out, then don't worry about it for now.

We can probably more easily handle the login stuff via Playwright including the frontend too.

To test as backend only I believe it would require setting DEBUG=False on the API and also updating the OSM redirect URL to go to the backend and not the frontend

PR #1624, opened 2024-07-03
Test coverage for projects mapping endpoints

azharcodeit:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [x] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Partially closes Issue #1614

## Describe this PR

This PR covers with tests following endpoints:

- POST projects/{project_id}/entity/status
- GET projects/{project_id}/entity/status
- GET projects/{project_id}/entities/statuses

It verifies that the endpoint correctly returns the entities' mapping statuses by comparing the response data with expected entities data.

To retrieve entities `odk_project` and `entities` fixtures have been created.
`odk_project` extends `project` fixture by generating project files through `project_crud` methods.

## Screenshots

Screenshot 2024-07-09 at 12 43 24

## Review Guide

Command to run all tests defined in the project using pytest:

`docker compose run --rm api pytest`

### DIFF
--- a/src/backend/tests/test_projects_routes.py
+++ b/src/backend/tests/test_projects_routes.py
@@ -31,6 +31,7 @@ from loguru import logger as log
 from shapely import Polygon
 
 from app.central.central_crud import create_odk_project
+from app.central.central_schemas import TaskStatus
 from app.config import encrypt_value, settings
 from app.db import db_models
 from app.projects import project_crud, project_schemas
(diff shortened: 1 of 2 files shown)


src/backend/tests/test_projects_routes.py
```diff
@@ -367,6 +367,33 @@ async def test_project_by_id(client, project):
...
+        {
+            "id": "455d0982-f4d9-4a08-ab94-dba22aa951b2",
+            "task_id": 1,
+            "osm_id": 302546443,
+            "status": 0,
+            "updatedAt": None,
+        }
+    ]
+
+    with patch(
```
spwoodcock, line 383: Nice idea with using unittest.patch, but I think this test isn't actually testing anything as a result!

If you patch the response to be what you want it to be, then testing the response equals what you set it to doesn't help much 😄 

I would recommend writing this test to incorporate both posting an entity status and then reading a single entity, and then possibly reading `/entities/statuses` after to check it returns correctly as a list of entities too.

spwoodcock (changes requested):
see comment

src/backend/tests/test_projects_routes.py
azharcodeit: @spwoodcock, I have addressed this in the last commit and added additional tests for updating and retrieving the status of a single entity.

The PR description and title have been updated accordingly, please have look. Thanks!🙏

spwoodcock (approved):
Very nice! Great work on this 💥

Issue #1650, opened 2024-07-14
Use XForwarded Middleware for Kubernetes  Deployments


spwoodcock:
**Is your feature request related to a problem? Please describe.**

Described better than I could here:

https://safir.lsst.io/user-guide/x-forwarded.html

**Describe the solution you'd like**

Use a middleware to correctly pass the user IP through any proxy to the application server for logging. 

There is one provided here https://github.com/lsst-sqre/safir/tree/main/src/safir/middleware

The code is MIT, so to save pulling in the stuff we don't need we could just take the middleware and attribute the user.

spwoodcock:
Hey Ronald! Thanks so much for the message and offer to contribute 🙏 

We have since integrated [Bunkerweb](https://github.com/bunkerity/bunkerweb) into the software stack, so that needs to be taken into account.

In production we specify the container IP range:
https://github.com/hotosm/fmtm/blob/fd654fdc4567299624388527e6a0d8bac5c0fa2d/deploy/compose.main.yaml#L38-L44

Then for the Bunkerweb (Nginx) proxy we whitelist the internal IPs:
https://github.com/hotosm/fmtm/blob/fd654fdc4567299624388527e6a0d8bac5c0fa2d/deploy/compose.main.yaml#L64

I wrote this a while ago & I'm not 100% certain what the cause is and if this is the possible solution - I think it is, but definitely worth double checking!
To avoid this becoming an [XY problem](https://xyproblem.info), I will describe what we need:
- Essentially, the user connects to a service via the the Nginx reverse proxy, which logs the IP of the internal service that the user connected to, instead of the IP of the connecting user.
- I believe the IP of the user needs to be forwarded, perhaps by XForwarded middleware, so that it can be displayed in the proxy logs.

An example of the nginx logs (I redacted some info):

```
# This one is from the ODK Collect mobile app
[NGINX.ACCESS] odk.dev.fmtm.hotosm.org 10.20.30.1 - - [08/Apr/2025:21:59:44 +0000] "GET /v1/key/xxxxxx/projects/14/forms/buildings/manifest HTTP/2.0" 304 0 "-" "org.odk.collect.android/v2024.3.5-entity-select-5-gd9dcd5e43e-dirty Dalvik/2.1.0 (Linux; U; Android 13; xxx Build/xxx)"

# This one is from FastAPI backend --> ODK Central
[...]

PR #1695, opened 2024-07-23
Replace custom nginx / certbot config with BunkerWeb

### LINKED ISSUE #1479: Add rate limiting to API calls

- After we add a task queue, Redis will be part of our stack.
- We can also use Redis to effectively handle rate limiting.

- Rate limiting should be implementing based on:
  - Number of requests from an IP address per minute.
  - Possibly total number of requests per minute to avoid DDOS?

- We could add something like fail2ban to the Nginx proxy, which isn't a bad idea.

**Additional context**

- For most endpoint we are protected by `login_required`, making distributed attacks harder.
- But we do have some unprotected endpoints out of necessity, so this is a moot point.

### LINKED ISSUE #1686: Assess usage of BunkerWeb to replace custom Nginx & certbot containers

- BunkerWeb is basically just a configuration for Nginx, using best practice security and many nice options.
- It's easy to use with docker and has good Kubernetes integration into Ingress.

Security options, from their own site:
- HTTPS support with transparent Let's Encrypt automation : Easily secure your web services with automated Let's Encrypt integration, ensuring encrypted communication between clients and your server.
- State-of-the-art web security : Benefit from cutting-edge web security measures, including comprehensive HTTP security headers, prevention of data leaks, and TLS hardening techniques.
- Integrated ModSecurity WAF with the OWASP Core Rule Set : Enjoy enhanced protection against web application attacks with the integration of ModSecurity, fortified by the renowned OWASP Core Rule Set.
- Automatic ban of strange behaviors based on HTTP status code : BunkerWeb intelligently identifies and blocks suspicious activities by automatically banning behaviors that trigger abnormal HTTP status codes.
- Apply connections and requests limit for clients : Set limits on the number of connections and requests from clients, preventing resource exhaustion and ensuring fair usage of server resources.
- Block bots with challenge-based verification : Keep malicious bots at bay by challenging them to solve puzzles such as cookies, JavaScript tests, captcha, hCaptcha, reCAPTCHA or Turnstile, effectively blocking unauthorized access.
- Block known bad IPs with external blacklists and DNSBL : Utilize external blacklists and DNS-based blackhole lists (DNSBL) to proactively block known malicious IP addresses, bolstering your defense against potential threats.

This would:
- Fix https://github.com/hotosm/fmtm/issues/1479
- Fix https://github.com/hotosm/fmtm/issues/1650
As both can be configured easily.

- Replace the custom built proxy image with BunkerWeb in docker-compose config.
- Also add to Kubernetes Helm chart if time.

spwoodcock:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [x] 🤖 Build or CI

## Related Issue

Fixes #1686
Also fixes #1479

Related to: https://github.com/hotosm/fmtm/issues/1650, but does not fix it.
(the FastAPI middleware is probably required after all)

## Describe this PR

Bunkerweb repo: https://github.com/bunkerity/bunkerweb/

- Replaces the nginx configuration on this repo entirely with BunkerWeb, which has a lot of nice security defaults (see original issue).
  - I was able to remove the entire `nginx` directory and related code / configs / scripts.
  - This also manages all certs for us, rather than our custom certbot config.
- @dakotabenjamin this is a pretty neat all-in-one web proxy that I would recommend to put all our services behind! It has rate limiting, brotli compression, optional antibot, good ModSecurity defaults, 'bad behaviour' checking and banning.
- I added the config for local development. Still todo `development` and `main` compose configs.
- It's a bit of a pain that the architecture of this system relies on a `scheduler` and docker socket binding service. Thankfully I managed to get this to work as initialisation only, rather than running continuously.
  - Using `timeout 120` for each service allows the configs to generated and copied to the nginx webserver, then the scheduler and docker socket service shut down.

- Adding modsecurity etc to our existing nginx config.
- But why bother maintaining our own nginx module building (e.g. brotli, modsecurity) and configuration, when we an outsource this to a well adopted upstream service.
- Resource usage is no more than the standard nginx server we were using, about 100-200mb memory usage at idle.

https://github.com/hotosm/fmtm/blob/development/docs/decisions/0006-web-app-firewall.md

## Notes

- Once merged, we can remove the proxy images from the repo.
- The NGINX_IMG_TAG var can also be removed from Github environment vars.
- Local projects will no longer work as the ODK_CENTRAL_URL was updated from https://proxy --> https://odkcentral
  - Existing projects for local development should be removed `docker compose down -v`

## Review Guide


dakotabenjamin:
I've used a turnkey-like server solution Caddy before, but this seems like it's got even more and its security focused. Looks good to me.

spwoodcock:
Thanks!

I like that it reduces the maintenance burden for building nginx images with brotli and maintaining configs 😄

Hoping to finish this at some point & also roll it out for DroneTM eventually too 👍

spwoodcock:
Tests failing I presume because the CI image doesn't have the correct dev cert for ODK Central.

Merging pre-emptively in the hope that a new CI image build will fix this 😄

Issue #1719, opened 2024-07-29
Better frontend validation of input TMS URL for custom imagery


spwoodcock:
**Describe the bug**

- We don't do much validation of the input TMS URL

**To Reproduce**

- `I tested with this URL: https://tiles.openaerialmap.org/5e6785161804990005e37072/0/5e6785161804990005e37073/{ z}/{x}/{y}`
- There was a simple space in the `{ x}`, causing parsing of the URLs and downloads to fail.

**Expected behavior**

- There should be better checking / parsing of the provided URL, to prevent errors.
- Validation should run in project creation and the basemaps modal.

manjitapandey:
but Validation on frontend during project creation is necessary, so that project aren't created with invalid TMS url.

PR #1725, opened 2024-08-01
Added a validator to check for valid custom tms url

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue #1719 

## Describe this PR

This PR adds a field validator to check for the valid custom tms url. These changes checks for the:
- valid place holders such as `{z} {x} {y}` return exception if there are any invalid place holders for example, space in `{ z}`
- checks for valid file extensions such as `['png', 'jpg', 'jpeg', 'tiff', 'webp', 'pbf']` so it will raise exception if invalid extension such as `.json`
- raises `HTTPException` if any error besides above conditions

## Screenshots

N/A

## Review Guide

Try creating project with invalid tms url it will raise error at the last step when create project is submitted.

### DIFF
--- a/src/backend/app/projects/project_schemas.py
+++ b/src/backend/app/projects/project_schemas.py
@@ -256,21 +256,21 @@ class ProjectIn(BaseModel):
                 full_pattern = placeholders_pattern + extension_pattern
 
                 if not re.match(full_pattern, custom_tms_url):
-                    if re.search(r'\.\w+$', custom_tms_url):
+                    if re.search(r"\.\w+$", custom_tms_url):
                         raise ValueError(
                             "Invalid file extension in TMS URL. Valid extensions are: "
                             f"{', '.join(valid_extensions)}."
-                            )
+                        )
                     else:
                         raise ValueError(
                             "Invalid TMS URL format. Please check the URL structure."
-                            )
+                        )
             return custom_tms_url
         except ValueError as e:
                 raise HTTPException(
                     status_code=HTTPStatus.BAD_REQUEST, 
                     detail=str(e)
-                    )from e
+                ) from e
 
 
 class ProjectUpload(ProjectIn, ODKCentralIn):
--- a/src/backend/app/projects/project_schemas.py
+++ b/src/backend/app/projects/project_schemas.py
@@ -267,7 +267,10 @@ class ProjectIn(BaseModel):
                             )
             return custom_tms_url
         except ValueError as e:
-                raise HTTPException(status_code=HTTPStatus.BAD_REQUEST, detail=str(e)) from e
+                raise HTTPException(
+                    status_code=HTTPStatus.BAD_REQUEST, 
+                    detail=str(e)
+                    )from e
 
 
 class ProjectUpload(ProjectIn, ODKCentralIn):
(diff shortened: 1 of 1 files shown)


spwoodcock (changes requested):
Nice!

TMS URL format is very variable though, and the {x} {y} {z} placeholders are not actually mandatory 😆

The file extension is also entirely optional, but if included it's good to check it's a valid one 👍

This is a good final check to avoid errors, but ideally we also need to pair with frontend validation too for real time user feedback if there are issues (better user experience)

spwoodcock:
In osm-fieldwork there is a basic regex to strip the {x} {y} {z}

Perhaps that could be updated to be more flexible and strip { x} too (if possible)

Sujanadh:
in case of file extension i have just added couple of extensions to check if user provides false extension , if no extension is provided it won't check for extension validation as url may not always have extension so it works and checks for the place holders.

Can you explain me what things should be checked for a valid tms url?

Sujanadh:
what do you think validation should be in fmtm or osm-fieldwork ?
And yes we need validation in frontend for the real time user experience

spwoodcock:
Ideally the backend validation should probably be in osm-fieldwork, with the frontend validation done in FMTM 👍 

I'm open to ideas on the criteria to validate a TMS URL.
But honestly it may not be possible.

If we were provided a URL like: `https://someserver.com/mytiles/15/444/333.png`, then we could validate the structure of the final part of the URL {num 0-24}/{num}/{num}

But as we get a URL more like this: `https://someserver.com/mytiles/` or `https://someserver.com/mytiles/{z}/{x}/{y}` then there is no good way I can think of to validate this, other than extracting the tile image format from an extension like `.png` and removing the `{z}/{x}/{y}` from the URL.

A good start would be to make the regex more flexible in osm-fieldwork so that `{ z}/{x   }/{ y}` won't break it

Sujanadh:
we won't get number in the URL, so whats the point of checking {num 0-24}/{num}/{num} it is only generated when creating tiles for different zoom level. Ideally, we can only check if there are place holders for x y z we can check {   x} and validate it. in terms extension should we check it? if it is valid or not?

spwoodcock:
Sorry my message above wasn't clear - that's what I meant 😄

I was giving an example 'if we got the URL in this format, then we could...'. But we don't get the URL with actual grid coordinates, so we can't check that!

As you say, we should check for and remove any placeholder values, if present. 

They will mess up the download if any of the placeholder remains

spwoodcock:
And yea to your last question 👍

If there is an extension present, validate it & then remove it (this has to be in osm-fieldwork though as we need to know what the file format was for the pmtile or mbtile generation

Sujanadh:
In `osm-fieldwork` , only two extensions (.jpg' , '.png') are checked and removed. Should I add more or leave it as it is?

spwoodcock:
Should be fine - I have yet to see a `.webp` TMS server or anything else - if this changes in future we can address it 😄

spwoodcock:
Rebased this to test an update to the PR Labeler workflow 👍 

Should we change this PR to be frontend based validation?
If the URL checks are good enough in osm-fieldwork we can just use those.
If not, the checks here may be useful: if extension exists, validate it is in list of options

Sujanadh:
Yes, we can change this to frontend based validation.
URL checks in `osm-fieldwork` do work; it checks for any invalid placeholders in the URL. We can use that.

Issue #1615, opened 2024-07-01
Add Playwright tests for mapper workflow


spwoodcock:
**Is your feature request related to a problem? Please describe.**

The mapping workflow:
- Load a project detail page.
- Click on a task and a popup displays (with a QR code).
- Lock the task area.
- The features load in the background for the task area.
- Click on a feature within the task.
- A popup displays.
- Click the 'Map Feature In ODK' button.
- Click on the task area again and mark the task area as complete.

**Describe the solution you'd like**

- Frontend tests for the above workflow using Playwright.
- Perhaps not as one large test, to aid maintainability.
- Could be divided into:
  - Task locking / unlocking.
  - Feature mapping.

PR #1766, opened 2024-09-03
Add Playwright tests for mapper flow

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [x] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #1615

## Describe this PR
This PR contains the following works: 
- Status not showing issue on task popup
- Mapper flow test
- Entity Feature map

## Screenshots
image_720


src/frontend/e2e/entityPopup.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  await page.waitForTimeout(4000);
+  await page.locator('canvas').click({
+    position: {
+      x: 387,
+      y: 211,
+    },
+  });
+  await expect(page.getByRole('heading', { name: 'Feature:' })).toBeVisible();
+  await expect(page.getByRole('button', { name: 'MAP FEATURE IN ODK' })).toBeEnabled();
+  await page.getByRole('button', { name: 'MAP FEATURE IN ODK' }).click();
```
spwoodcock, line 37: After this click I think we should expect a console warning that the redirect to protocol `odkcollect://` failed (it attempts to open the link via mobile intent).

We could set a console listener via playwright and check this warning is displayed (it's not a user facing thing, but it helps check the button works as intended)

spwoodcock (commented):
This is great!

A few other things to check could be:

- Adding a comment works
- Reading the comment on the tab

In future I am wondering if it's possible to make a mock form submission too, so we can do the next round of tests (check the submission download, validator workflow, etc) - a future PR

(Update: to do the above we can replicate something like this https://forum.getodk.org/t/creating-fake-odk-data/26144/3
using JavaScript instead of Python. We simply need to POST some pre-generated XML data to the ODK Central submission API)

Issue #1722, opened 2024-07-31
Injecting mandatory FMTM fields into custom XLSForms


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- Currently our XLSForm has hidden fields that are mandatory: feature / task selection, digitization_correct questions.
- This is prone to breakage if the user uploads a custom XLSForm and accidentally deletes fields (despite being hidden).
- We need a more resilient mechanism to allow for custom XLSForm upload flexbility, but also impose our restrictions / requirements that FMTM needs to function well.

**Describe the solution you'd like**

- Option 1: append the fields in the XLSForm, either via pyxlsx or possibly via Pandas, **prior to XLS --> XML**.
  - Load the XLSForm as a dataframe.
  - Have pre-prepared separate XLSForms with building questions & digitization_correct questions.
  - Read the building selection questions into a dataframe and append to start.
  - Read the digitization_correct questions into a dataframe and append to end.
  - However, this will introduce a big dependency: Pandas, which is not ideal.

- Option 2: use the current XML parsing / manipulation method to inject all required params, **after XLS --> XML**.
  - This method is possible, but may have many fields we need to insert into specific places.
  - Needs testing and is more prone to breakage.

**Additional context**

- This first option would probably require we update all form handling to use Pandas, which would be quite an improvement to XLS wrangling.

> [!Note]
> See comment below. We should not use Pandas, but instead `python-calamine`.

spwoodcock:
This would make the XLSForms a whole lot more maintainable.

Currently if we have say 10 categories = 10 XLSForms, we need to keep the extra fields we use in FMTM in sync for all of them.

Ideally we have:
- 1 XLSForm with mandatory fields for FMTM.
- 10 XLSForms with questions only related to the specific mapping category.
- On project creation we merge the selected form with the 'FMTM mandatory fields' XLSForm.

spwoodcock:
I just did a small assessment of package size between various XLS and XLSX parser libraries:

![image](https://github.com/user-attachments/assets/13c0fa7e-d1cd-4278-a498-cd84e1f11646)

- Pandas is nice to work with, but pulls in so many unnecessary libraries it's not worth it.
- openpyxl and xlrd would be required together to parse XLSX and XLS files respectively.
[...]

PR #1763, opened 2024-08-29
Update form by injecting mandatory fields and validate it

Sujanadh:
## What type of PR is this? (check all applicable)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI
- [ ] ❓ Other (please specify)

## Related Issue

- Issue #1722 

## Describe this PR

This PR depends on `osm-fieldwork` in order to inject `mandatory_fields` and `digitisation_fields` into a custom form if uploaded.

### Changes:-
- custom form if uploaded gets updated with mandatory and digitisation fields during form validation
- `validate-form` used to return a message if the form is valid or not; now it returns form data in bytes along with the message if the form is valid.

### Frontend Changes:
- On the frontend side, the bytes returned need to be stored, converted to file and passed during `generate-files` of project creation
- Or backend changes needed to accept bytes/file both.

## Screenshots

N/A

## Alternative Approaches Considered

Did you attempt any other approaches that are not documented in code?

## Review Guide

N/A

## Checklist before requesting a review

- 📖 Read the FMTM Contributing Guide: <https://github.com/hotosm/fmtm/blob/main/CONTRIBUTING.md>
- 📖 Read the HOT Code of Conduct: <https://docs.hotosm.org/code-of-conduct>
- 👷‍♀️ Create small PRs. In most cases, this will be possible.
- ✅ Provide tests for your changes.
- 📝 Use descriptive commit messages.
- 📗 Update any related documentation and include any relevant screenshots.

Issue #1653, opened 2024-07-15
Support multiple feature selection within single form submission


manjitapandey:
**Is your feature request related to a problem? Please describe.**
I am trying to upload xlsx file where there is the need for selection of more than one feature. The building feature and  road connecting that building. The another case could be selection of main building in case we are surveying associated building.  

**Describe the solution you'd like**
User should be allowed to select multiple features. This increases the use case on FMTM for multiple field surveying scenario. May be we could store the geojson of only the entity we are surveying and in case of main feature, we can just store the feature id. 

**Describe alternatives you've considered**

**Additional context**
This is the user's requirement from the Janakpur field mapping activity. 
[Survey_ready_form (12).zip](https://github.com/user-attachments/files/16230196/Survey_ready_form.12.zip)

manjitapandey:
@Sujanadh , @spwoodcock , need your suggestions on how can we technically achieve this on FMTM.

Sujanadh:
I think we can add geopoint of the main building for the survey of its associate buildings and select the geometry of the associate building.

spwoodcock:
My knowledge of the use case is hazy, but as Sujan says one option and probably the best solution for selecting 'associated' features, is to add a geopoint question in the XLSForm asking the user to 'pinpoint associated feature' on the map.

It is possible to enable `select_multiple_from_file`, but then you would have to distinguish somehow which is the 'main' building being mapped and which is the 'associated'.

If we need the select_multiple capability for something then we can explore this option further.

Otherwise if it's just an issue of form design using geopoints, I think this issue could be closed (maybe we add docs on how to do this)

manjitapandey:
@spwoodcock, the use case here is that user should be able to select associated building as well as the main building which is also present within the same geojson file. Moreover, They also needs to select the road (linear feature) associated with the building from the same geojson. This means we should allow user to select multiple feature. 

> It is possible to enable `select_multiple_from_file`, but then you would have to distinguish somehow which is the 'main' building being mapped and which is the 'associated'.

Exactly, Enabling "Select_multiple_from_file" solves the problem. How do we distinguish which is the main building being mapped and which one is associated.

spwoodcock:
We need to fit around what XLSForms and ODK Collect support, and as far as my knowledge goes we can't distinguish between the geometries using `select_multiple_from_file`.

We could do one of the following:
- Use multiple `select_one_from_file` fields: one of the primary feature, another for an associated feature.
[...]

PR #1792, opened 2024-09-17
Use XLSForm injection during project creation

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes https://github.com/hotosm/fmtm/issues/1722

## Describe this PR

- PR https://github.com/hotosm/fmtm/pull/1763 added the osm-fieldwork update_xlsform work to the `validate-form` endpoint.
- This PR also adds the functionality during the project creation.
- Users can upload a custom form full of question and have the required FMTM fields injected automatically.
- Removes the custom XML manipulation that was difficult to understand / maintain, replaced with much easier Pandas-based XLSForm concatenation.

## Review Guide

- Rebuild the docker image
- Run the API
- Create a new project using a custom form
- Also test using the default / bundled forms

### DIFF
--- a/src/backend/app/central/central_crud.py
+++ b/src/backend/app/central/central_crud.py
@@ -379,7 +379,7 @@ async def update_project_xform(
     category: str,
     task_count: int,
     odk_credentials: project_schemas.ODKCentralDecrypted,
-) -> BytesIO:
+) -> None:
     """Update and publish the XForm for a project.
 
     Args:
@@ -390,23 +390,9 @@ async def update_project_xform(
         task_count (int): The number of tasks in a project.
         odk_credentials (project_schemas.ODKCentralDecrypted): ODK Central creds.
 
-    Returns:
-        BytesIO: updated XLSForm for the project.
-            NOTE this is the XLSForm, not the XForm XML.
+    Returns: None
     """
-    xform_bytesio = await validate_and_update_user_xlsform(
-        xlsform=xlsform,
-        form_category=category,
-        task_count=task_count,
-        existing_id=xform_id,
-    )
-
-    xlsform_updated = await append_fields_to_user_xlsform(
-        xlsform=xlsform,
-        form_category=category,
-        task_count=task_count,
-        existing_id=xform_id,
-    )
+    xform_bytesio = await read_and_test_xform(xlsform)
 
     xform_obj = get_odk_form(odk_credentials)
 
@@ -417,9 +403,9 @@ async def update_project_xform(
         form_name=xform_id,
     )
     # The draft form must be published after upload
+    # NOTE we can't directly publish existing forms
+    # in createForm and need 2 steps
     xform_obj.publishForm(odk_id, xform_id)
-
-    return xlsform_updated
 
 
 async def convert_geojson_to_odk_csv(
--- a/src/backend/app/central/central_crud.py
+++ b/src/backend/app/central/central_crud.py
@@ -19,17 +19,15 @@
 
 import csv
 import json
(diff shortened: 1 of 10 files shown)


spwoodcock:
@Sujanadh sorry to distract from the fmtm-splitter work, but could you please take a look at this PR tomorrow?

I was coding into the night, so no doubt have missed something. Please check everything works as intended / looks good to you 😄 

This will finalise the work on injecting XLSForm fields and allow https://github.com/hotosm/fmtm/issues/1653 to be easily solved.

Thank you!

Sujanadh:
sorry i didn't see the PR , i will test it

Sujanadh:
when trying to check custom form also there is issue 
```
{
  "detail": "XLSForm is invalid: There has been a problem trying to replace ${existing} with the XPath to the survey element named 'existing'. There is no survey element with this name."
}
```

Sujanadh:
i don't think we need to validate it again in `generate-project-date` since we append and validate using validate-form endpoint then return valid appended form which is used while generating project data. For default form we don't need to do anything , instead we can change xls form to allow to create new entity with new geopoint.

spwoodcock:
Thanks for the tests!

The issue about `existing` is from the old FORL that has calculated fields based on this. As you say, using a custom form without this logic works fine.

Agree we could validate once only!
The reason I left it in is because the API is independent from the frontend. 

While it's quite likely our frontend is the only user of the API, calling validate-form prior to generating project files, we can't guarantee it. So validating again is just an addition check, without much performance penalty - what do you think?

spwoodcock:
> i don't think we need to validate it again in `generate-project-date` since we append and validate using validate-form endpoint then return valid appended form which is used while generating project data. For default form we don't need to do anything , instead we can change xls form to allow to create new entity with new geopoint.

Regarding the default forms, I already updated the forms to remove the verification fields (so they are injected).

I will update them again now to remove the FMTM mandatory fields entirely, and have them injected the same way as a custom form.

This should improve maintainability, as we don't need to update the entire XLSForm library when we change something in the mandatory fields 😄 

Related issue I just made: https://github.com/hotosm/osm-fieldwork/issues/300

spwoodcock:
> `HTTPStatus.UNPROCESSABLE_ENTITY: XLSForm is invalid: There are two sections with the name survey_questions.` yes while appending there is a problem since we need to get common fields and then append likewise
> 
> I used default form here.

This is a valid error though! We are having issues with the dataframe merging logic

Sujanadh:
> Thanks for the tests!
> 
> The issue about `existing` is from the old XLSForm that has calculated fields based on this. As you say, using a custom form without this logic works fine.
> 
> Agree we could validate once only! The reason I left it in is because the API is independent from the frontend.
> 
> While it's quite likely our frontend is the only user of the API, calling validate-form prior to generating project files, we can't guarantee it. So validating again is just an addition check, without much performance penalty - what do you think?

Ah got it, you are right that makes sense.

spwoodcock:
Made some updates!

Hopefully fixed project creation and XLSForm updating 😄

spwoodcock:
I made some updates to osm-fieldwork for this, so the main things that need to be done:
- [ ] Do some tests all the way through to the ODK Collect usage, to check the logic all works as intended.
- [ ] Deploy this to staging for Monday.

I am merging this now so we can more easily test - we can make another PR if needed.

Issue #1780, opened 2024-09-11
Error while opening custom ODK collect after selecting a feature and clicked on 'Go to ODK'


manjitapandey:
**Describe the bug**
An error is occured when go to odk is clicked after selecting a feature on fmtm mobile. 

**To Reproduce**

1. Open fmtm on mobile
2. select any feature and click on go to odk.

**Expected behavior**
The ODK should open with prefilled fields, when go to odk is clicked

**Screenshots**

https://github.com/user-attachments/assets/9c8cd716-c354-4e5e-bd6c-b58f9617d792

spwoodcock:
'Go to ODK' will open with the task id prefilled.

If you click on the building, then the button 'Map Feature in ODK', does this work loading the entity id?

manjitapandey:
Yes, clicking on building 'go to odk' feature work well. but its on task level.

spwoodcock:
Fixed in latest build 😄

https://github.com/hotosm/odkcollect/releases/tag/v2024.2.4-entity-select

I will link this in the sidebar for FMTM as the default download soon

spwoodcock:
Please test the APK @manjitapandey and close if it works for you 👍

manjitapandey:
Sure, is the updated apk linked on sidebar of FMTM??

manjitapandey:
The custom odk keeps crashing on my device. 
and also when I go to odk by locking a task, it redirect me to a section to scan the QR instead of form. 
when I go to odk selecting an entity, it just crashes or show the error message as in image below. I also tried deleting the cache and overall staging data. 

Uploading Record_2024-09-13-11-19-11.mp4…

![Screenshot_2024-09-13-11-11-07-40_171725cd2e369f62c287134214eaed67](https://github.com/user-attachments/assets/0fb3392a-2b25-436c-8c8b-b79cfabba8c3)

spwoodcock:
Can you retry this?

People were using the app in the field it seems, so it must work

spwoodcock:
Nice! The feature is actually pre-selected if you skip the task_id choice, but choosing a task id causes the feature to deselect!

First of all, the task_id should be prefilled when the form is opened - this is a bug.

Next, the XLSForm needs logic to:
- Hide the task_id question if feature is selected already (possible, but this might complicate things)
- Hide the 'do you want to add a new geometry' question if a feature is pre-selected.

spwoodcock:
[...]

PR #1810, opened 2024-09-26
Fix/staging bugs

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- QRcode size a bit small (trouble in scanning and loading project on first go)
- on Go to ODK btn click on task selection, no task selected on ODK | Related to #1780

## Describe this PR
- QRcode size increase
- taskId filter add to ODK redirection

### DIFF
--- a/src/frontend/src/components/DialogTaskActions.tsx
+++ b/src/frontend/src/components/DialogTaskActions.tsx
@@ -232,7 +232,7 @@ export default function Dialog({ taskId, feature }: dialogPropType) {
               );
 
               if (isMobile) {
-                document.location.href = `odkcollect://form/${projectInfo.xform_id}`;
+                document.location.href = `odkcollect://form/${projectInfo.xform_id}?task_filter=${taskId}`;
               } else {
                 dispatch(
                   CommonActions.SetSnackBar({
--- a/src/frontend/src/components/ProjectDetailsV2/MapControlComponent.tsx
+++ b/src/frontend/src/components/ProjectDetailsV2/MapControlComponent.tsx
@@ -74,7 +74,7 @@ const MapControlComponent = ({ map, projectName, pmTileLayerData }: mapControlCo
   };
 
   return (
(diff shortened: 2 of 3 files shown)


src/frontend/src/components/DialogTaskActions.tsx
```diff
@@ -232,7 +232,7 @@ export default function Dialog({ taskId, feature }: dialogPropType) {
               );
 
               if (isMobile) {
-                document.location.href = `odkcollect://form/${projectInfo.xform_id}`;
+                document.location.href = `odkcollect://form/${projectInfo.xform_id}?task_filter=${taskId}`;
```
spwoodcock, line 235: This was actually removed intentionally, as I deemed the task_filter param to be unnecessary - it confused things I think!

The custom ODK Collect should be default, then backup users can just select the geom they are standing next too quite easily with geolocation

spwoodcock: @manjitapandey not sure if you agree, but I think it's ok for 'Go to ODK' to just open the unfilled form

manjitapandey: Currently, when users access ODK at the task level, they are presented with all the project's features, which may cause confusion as they likely expect to see only the features relevant to the specific task. Skipping the task-specific question might not be a good idea in this context.

spwoodcock: But should we actually even have that as a flow? The main one we should target is *building* selection, not *task id* selection. 

From what I have seen actual mapping in practice is that the task id selection question is a source of confusion. Perhaps we should get some feedback from recent users.

I personally don't see it as much of an issue having the buildings loaded for users *in the field*. I agree it's difficult to select a building on a remote computer with no geolocation, but when in the field you are geolocated to show a dot next to the building to be mapped

PR #1848, opened 2024-11-04
feat: allow user deletion without deleting task references

### LINKED ISSUE #1661: Backend API to delete user account

- This is essential for GDPR. 

- Deletion of a user db entry, in addition to keeping references to the user for task history in place.
- How to keep the references in place? Save the username and profile picture in the task history?

Anuj-Gupta4:

- [x] 🍕 Feature

## Related Issue

Fixes #1661 

## Describe this PR

On user deletion, task history will retain username from user
and rest of the user data will be deleted.

### DIFF
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -222,6 +222,41 @@ class DbUser(BaseModel):
             return await cur.fetchall()
 
     @classmethod
+    async def delete(cls, db: Connection, user_id: int) -> bool:
+        """Delete a user and their related data."""
+        async with db.cursor() as cur:
+            await cur.execute(
+                """
+                UPDATE task_events SET user_id = NULL WHERE user_id = %(user_id)s;
+            """,
+                {"user_id": user_id},
+            )
+            await cur.execute(
+                """
+                UPDATE projects SET author_id = NULL WHERE author_id = %(user_id)s;
+            """,
+                {"user_id": user_id},
+            )
+            await cur.execute(
+                """
+                DELETE FROM organisation_managers WHERE user_id = %(user_id)s;
+            """,
+                {"user_id": user_id},
+            )
+            await cur.execute(
+                """
+                DELETE FROM user_roles WHERE user_id = %(user_id)s;
+            """,
+                {"user_id": user_id},
+            )
+            await cur.execute(
+                """
+                DELETE FROM users WHERE id = %(user_id)s;
+            """,
+                {"user_id": user_id},
+            )
+
+    @classmethod
     async def create(
         cls,
         db: Connection,
@@ -605,11 +640,11 @@ class DbTaskEvent(BaseModel):
 
     project_id: Annotated[Optional[int], Field(gt=0)] = None
     user_id: Annotated[Optional[int], Field(gt=0)] = None
+    username: Optional[str] = None
     comment: Optional[str] = None
     created_at: Optional[AwareDatetime] = None
 
     # Computed
-    username: Optional[str] = None
     profile_img: Optional[str] = None
     # Computed via database trigger
     state: Optional[MappingState] = None
@@ -662,13 +697,12 @@ class DbTaskEvent(BaseModel):
 
         sql = f"""
             SELECT
-                *,
-                u.username,
+                the.*,
                 u.profile_img
             FROM
-                public.task_events
-            JOIN
-                users u ON u.id = task_events.user_id
+                public.task_events the
+            LEFT JOIN
+                users u ON u.id = the.user_id
             WHERE {filters_joined}
             ORDER BY created_at DESC;
         """
@@ -696,18 +730,19 @@ class DbTaskEvent(BaseModel):
                         INSERT INTO public.task_events (
                             event_id,
                             project_id,
+                            username,
                             {columns}
                         )
                         VALUES (
                             gen_random_uuid(),
                             (SELECT project_id FROM tasks WHERE id = %(task_id)s),
+                            (SELECT username FROM users WHERE id = %(user_id)s),
                             {value_placeholders}
                         )
                         RETURNING *
                     )
                     SELECT
                         inserted.*,
-                        u.username,
                         u.profile_img
                     FROM inserted
                     JOIN users u ON u.id = inserted.user_id;
--- a/src/backend/app/users/user_routes.py
+++ b/src/backend/app/users/user_routes.py
@@ -19,11 +19,13 @@
 
 from typing import Annotated, List
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, Response
+from loguru import logger as log
 from psycopg import Connection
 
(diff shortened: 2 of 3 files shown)


src/backend/app/tasks/tasks_crud.py
```diff
@@ -176,7 +175,7 @@ async def update_task_status(
 
 
 async def create_task_history_for_status_change(
-    db_task: db_models.DbTask, new_status: TaskStatus, db_user: db_models.DbUser
+    db_task: db_models.DbTask, new_status: TaskStatus, db_user: db_models.DbUser, db
```
Sujanadh, line 178: provide `Session` type to db also.

src/backend/app/tasks/tasks_crud.py
```diff
@@ -185,15 +184,34 @@ async def create_task_history_for_status_change(
...
+        "action": get_action_for_status_change(new_status).name,
+        "action_text": msg,
+        "action_date": datetime.now(),
+        "user_id": db_user.id,
+    }
+
+    result = db.execute(query, params)
+    db.commit()
+
+    row = result.fetchone()
```
Sujanadh, line 213: instead of `fetchone() `simply use` first()`, `first()` will call only first row all the time, while `fetchone()` returns all the rows but only one at a time.

spwoodcock: This is true for SQLAlchemy, but not for psycopg actually!

Psycopg only has fetchone fetchmany and fetchall 😄

Apologies for not mentioning that!

spwoodcock:
So sorry @Anuj-Gupta4 , but this PR is based on staging, which is using SQLAlchemy 😅

We have since done a huge refactor on development to remove this and use psycopg, so this PR might need to be started from scratch with development as the base 😥

src/backend/app/tasks/tasks_crud.py
spwoodcock: Actually you are technically correct!

I noticed this PR is using SQLAlchemy! Commented in the main thread

spwoodcock:
There is some really nice code here 😄

Could you possibly rebase with development (might be a lot of hassle considering the amount of change), or stash your changes somewhere, then reset the branch and reapply?

Also feel free to make a new PR if it's easier, then I'll take a look!

Anuj-Gupta4:
@spwoodcock I have reset my local branch and am working over development code. I will be overwriting the git history once its completed.

src/backend/app/db/models.py
```diff
@@ -95,7 +95,7 @@ def dump_and_check_model(db_model: BaseModel):
 class DbUserRole(BaseModel):
     """Table user_roles."""
 
-    user_id: int
+    user_id: Optional[int] = None
```
spwoodcock, line 98: Is there a use case for a UserRole entry with no user_id set?

As we won't know the user for which the role relates to

src/backend/app/db/models.py
```diff
@@ -221,6 +221,41 @@ async def all(
...
+            """,
+                {"user_id": user_id},
+            )
+            await cur.execute(
+                """
+                DELETE FROM users WHERE id = %(user_id)s;
+            """,
+                {"user_id": user_id},
+            )
+
```
spwoodcock, line 258: Perfect 👍

src/backend/app/db/models.py
```diff
@@ -520,7 +555,7 @@ class DbOrganisationManagers(BaseModel):
     """Table organisation_managers."""
 
     organisation_id: int
-    user_id: int
+    user_id: Optional[int] = None
```
spwoodcock, line 558: Same for this - I don't know when user_id would be empty for a record

src/backend/migrations/010-delete-user.sql
```diff
@@ -0,0 +1,47 @@
...
+-- Add column 'username' to 'task_events' table
+DO $$
+BEGIN
+    IF NOT EXISTS (
+        SELECT 1
+        FROM information_schema.columns
+        WHERE table_name='task_events' AND column_name='username'
+    ) THEN
+        ALTER TABLE public.task_events
+        ADD COLUMN username VARCHAR(255);
```
spwoodcock, line 16: No need to specify max 255 chars - it has no efficiency gain in Postgres. Just VARCHAR is good 👍

spwoodcock (changes requested):
Very nice work! 🙌

src/backend/app/db/models.py
```diff
@@ -221,6 +221,41 @@ async def all(
             )
             return await cur.fetchall()
 
+    @classmethod
+    async def delete(cls, db: Connection, user_id: int) -> bool:
+        """Delete a user and their related data."""
+        async with db.cursor() as cur:
```
Sujanadh, line 227: Can we change this to a single SQL and await the execution at once? What do you think?

Anuj-Gupta4:
@spwoodcock The tests are passing locally. I am not sure why it is failing here.

src/backend/app/users/user_routes.py
```diff
@@ -66,3 +68,18 @@ async def get_user_roles(current_user: Annotated[DbUser, Depends(mapper)]):
...
+    current_user: Annotated[DbUser, Depends(super_admin)],
+    db: Annotated[Connection, Depends(db_conn)],
+):
+    """Delete a single user."""
+    log.info(
+        f"User {current_user.username} attempting deletion of user {user.username}"
+    )
+    await DbUser.delete(db, user.id)
+    log.info(f"User {user.id} deleted successfully.")
+    return Response(status_code=HTTPStatus.NO_CONTENT)
```
Sujanadh, line 85: I think it's best to return a response with a similar message as in the log instead of `NO_CONTENT`.

src/backend/app/db/models.py
spwoodcock: Unfortunately that's not possible with psycopg! It's only one statement per cursor 😅
(SQLAlchemy hides this fact!).

It looks inefficient, but it still executes them all in the same session, so it's not bad

Sujanadh: I see you have done a similar delete api for the organisation with a single sql.

spwoodcock: Oh right! Does it actually work? I would think it might error, but could be wrong. 

I think the project deletion has it separated for this reason

spwoodcock: Ah ignore me, I see what you mean now!

The org deletion is a single prepared statement, so that works fine

Sujanadh: I am not sure if it works, I haven't tested it yet. :thinking:

spwoodcock: > I am not sure if it works, I haven't tested it yet. :thinking: 

I think it does!

spwoodcock:
Oops forgot to wait for the other PR comments to be addressed before merge. 

It would be nice to possibly address them in another PR if time @Anuj-Gupta4 🙏

Issue #1849, opened 2024-11-04
Make submission data downloads only possible for project contributors


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- Currently anyone can download data for a public project.
- It might not be very clear to the project manager / users that this is the case.
- Sometimes sensitive data is collected.

**Describe the solution you'd like**

- Limit download of submissions to either:
  - Org manager, project manager.
  - Contributors (i.e. contributing mappers or validators).

**Other considerations**

Do we need an extra flag for projects `sensitive` that applies a submission view / downloading restriction to project admins and above?

PR #1873, opened 2024-11-12
feat(submission): add project_contributors dependency for permission …

Anuj-Gupta4:
## What type of PR is this?

- [x] 🍕 Feature

## Related Issue

Fixes Make submission data downloads only possible for project contributors #1849

## Describe this PR

The code adds a dependency to the routes responsible for submission downloads to only allow access if they are a project contributor. 

## Review Guide

The criteria to be a project contributor may need further discussion.

### DIFF
--- a/src/backend/app/auth/roles.py
+++ b/src/backend/app/auth/roles.py
@@ -390,5 +390,5 @@ async def project_contributors(
 
     raise HTTPException(
         status_code=HTTPStatus.FORBIDDEN,
-        detail="You do not have permission to access this resource.",
+        detail="You must be a project contributor to access this resource.",
     )
--- a/src/backend/app/auth/roles.py
+++ b/src/backend/app/auth/roles.py
@@ -334,3 +334,61 @@ async def mapper(
         current_user,
         ProjectRole.MAPPER,
     )
+
+
+async def project_contributors(
+    project: Annotated[DbProject, Depends(get_project)],
+    db: Annotated[Connection, Depends(db_conn)],
+    current_user: Annotated[AuthUser, Depends(login_required)],
+) -> ProjectUserDict:
+    """A contributor to a specific project."""
+    user_id = current_user.id
+    project_id = project.id
+    org_id = project.organisation_id
+
+    query = """
+        SELECT * FROM users
+        WHERE id = %(user_id)s
+            AND (
+                CASE WHEN role = 'ADMIN' THEN true
+                ELSE
+                    EXISTS (
+                        SELECT 1 FROM organisation_managers
+                        WHERE organisation_managers.user_id = %(user_id)s
+                          AND organisation_managers.organisation_id = %(org_id)s
+                    )
+                    OR EXISTS (
+                        SELECT 1 FROM user_roles
+                        WHERE user_roles.user_id = %(user_id)s
+                          AND user_roles.project_id = %(project_id)s
+                          AND user_roles.role = 'PROJECT_MANAGER'
+                    )
+                    OR EXISTS (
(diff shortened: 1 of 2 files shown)


src/backend/app/auth/roles.py
```diff
@@ -334,3 +334,60 @@ async def mapper(
...
+    project_id = project.id
+    org_id = project.organisation_id
+
+    query = """
+        SELECT
+            EXISTS (
+                SELECT 1 FROM projects
+                WHERE author_id = %(user_id)s
+            ) OR
+            EXISTS (
```
Sujanadh, line 356: I think it will be good to use a `CASE` statement like we do in check_access?  first checking higher roles followed accordingly.

src/backend/app/auth/roles.py
```diff
@@ -334,3 +334,60 @@ async def mapper(
...
+            ) AS has_permission;
+    """
+    async with db.cursor() as cur:
+        await cur.execute(
+            query,
+            {"user_id": user_id, "project_id": project_id, "org_id": org_id},
+        )
+        result = await cur.fetchone()
+
+    print("result is")
```
Sujanadh, line 385: I think it's good to avoid placing print statements.

Sujanadh, line 385: I think it's good to avoid placing print statements.

src/backend/app/auth/roles.py
```diff
@@ -334,3 +334,60 @@ async def mapper(
...
+    print(result, result[0])
+    if result and result[0]:
+        return True
```
Sujanadh, line 388: instead of returning True we can return user_dict and use that to get information of user, project.

src/backend/app/submissions/submission_routes.py
```diff
@@ -64,6 +64,7 @@ async def read_submissions(
 @router.get("/download")
 async def download_submission(
     project_user: Annotated[ProjectUserDict, Depends(mapper)],
+    _: Annotated[bool, Depends(project_contributors)],
```
Sujanadh, line 67: We don't need two dependency here, we can simple use project_contributor instead of mapper.
```python
project_user: Annotated[ProjectUserDict, Depends(project_contributors)]
like this.

File: src/backend/app/submissions/submission_routes.py [hunk 5]
```diff
@@ -365,6 +366,7 @@ async def update_review_state(
 
 @router.get("/download-submission-geojson")
 async def download_submission_geojson(
+    _: Annotated[bool, Depends(project_contributors)],
```
Sujanadh, line 369: same here

Sujanadh (approved):

src/backend/app/auth/roles.py
```diff
@@ -334,3 +334,61 @@ async def mapper(
...
+
+    if db_user:
+        return {
+            "user": db_user,
+            "project": project,
+        }
+
+    raise HTTPException(
+        status_code=HTTPStatus.FORBIDDEN,
+        detail="You do not have permission to access this resource.",
```
spwoodcock, line 393: Could this be changed to better inform the user when displayed on the frontend:

'You must be a project contributor to access this'

spwoodcock (changes requested):
Minor change to a comment needed, but this is a really neat addition 😄

My only concern is that this is quite restrictive. 

In the long term I think we need to go the route discussed a few meetings ago, with different level of project access being properly implemented. So this logic would perhaps be moved to the mapper role, determining access based on the 4 project visibilities discussed.

Let's go with that for now though!

PR #1887, opened 2024-11-15
refactor(logging): enhance error logging with stack information across multiple modules

### LINKED ISSUE #1756: Exceptions should use logger.exception, not logger.error
**Describe the bug**

- We use `log.error` in most places, even when exceptions are raised.
- This is fine when viewing logs in the terminal.
- However to send via OTEL to tools like Sentry, the traceback is only included if `log.exception` is used.
  - This is essentially the same as `log.error(exc_info=True)`.
- If we want the full traceback stack to be included, then we need `log.exception("some info", stack_info=True)`

Anuj-Gupta4:
- [x] 🧑‍💻 Refactor

## Related Issue

Resolves #1756 

## Describe this PR

Replaced log.error with log.exception and included traceback information where exceptions are raised.

### DIFF
--- a/src/backend/app/auth/auth_routes.py
+++ b/src/backend/app/auth/auth_routes.py
@@ -215,7 +215,7 @@ async def get_or_create_user(
 
     except Exception as e:
         await db.rollback()
-        log.error(f"Exception occurred: {e}")
+        log.exception(f"Exception occurred: {e}", stack_info=True)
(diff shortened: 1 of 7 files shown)


spwoodcock:
Looks good!

Could you purposefully raise some exceptions to test this out & see what the logging format is like?

PR #1891, opened 2024-11-17
Replace `pdm` dependency solver with `uv`

spwoodcock:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [x] 🤖 Build or CI

## Describe this PR

uv:
- `uv` was released not too long ago and has since almost gained feature parity with other tools like `pdm`.
- It's written in `rust` and is orders of magnitude faster at locking dependencies: a minor win, but a nice one nonetheless.
- It also allows you to run different versions of Python on your system very easily.
- Thanks @Pradip-p for pointing this tool out!

What I did:
- Replaced all usage of `pdm` with `uv`.
- Overhauled the backend dockerfile to work with `uv` and be a bit cleaner.
- Updated docs around using `pdm`.

Minor caveats:
- We can no longer use the `dynamic` field in `pyproject.toml`: `dynamic=["version"]`.
- This isn't a big deal - we can just hardcode it. The version updating is done by `commitizen` anyway, so nothing lost.

Note: I plan to talk about using this tool with the dev teams when I visit NAXA 👍 

## Screenshots

image


Pradip-p:
When I heard about UV, it always triggered one question: **what exactly makes it faster than other tools like pip, PDM, or Poetry?**  

it seems UV uses heavy parallelization, advanced caching, and an optimized way to access metadata from PYPI. Instead of downloading entire package wheels, it queries the zip file’s Central Directory and fetches only the needed metadata using file offsets, avoiding unnecessary downloads and speeding up dependency resolution.  

Am I understanding this correctly? Let me know your thoughts!  @spwoodcock @nrjadkry  @Sujanadh

spwoodcock:
You are definitely right when I comes the install part!

When it comes to speeding up the locking, the key factor here is that this is a library written in `rust`, a much lower level programming language with better parallelization than python (as you mention).

They most likely use pyo3 to generate Python bindings for the underlying `rust` library, meaning we get to install and use it from Python 😄

Same concept for `ruff`, one of astral's other tools

PR #1903, opened 2024-11-21
Mapper frontend login + remove temp auth from React frontend

NSUWAL123:
## What type of PR is this? (check all applicable)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI
- [ ] ❓ Other (please specify)

## Related Issue
- Related to #1890

## Describe this PR
This PR contains the following works: 
**Mapper Frontend**
- OSM and Temp login add to mapper frontend
- Remove Temp login from react frontend
- Change hot-header component with custom header component with drawer

**React Frontend**
- Replace Legend Toggle to map control component
- Add button named `START MAPPING` to link react frontend to mapper frontend

## Screenshots
image
image
image
image

PR #1910, opened 2024-11-22
Persist osm login across playwright tests (temp login removed)

spwoodcock:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [x] ✅ Test
- [ ] 🤖 Build or CI

## Related

#1903 as the PR where this issue was discussed

## Describe this PR

- Didn't get a chance to finish this due to my flight, but hopefully you get the idea here @NSUWAL123 
- If I don't get to look at this until Monday, we could go over it together in person 😄
- The context should probably be set just after the OSM login succeeds, then waiting for the redirect will allow the cookie to be set for the subsequent test. 
- The context wont save the cookie as it's httpOnly, but should save the OSM login part hopefully, to avoid having to do the Oauth flow for OSM again - let's see!
- In the end the main goal is automated OSM login. To do that we need to make a test OSM account and can just hardcode the creds to this repo. If we optimise by reducing calls to OSM via the auth context, this is a bonus, but not essential for now.

- Disable auth on the backend when DEBUG=true.
- But this would mean we can't test properly with roles, and this could be important in future. 

### DIFF
--- a/src/backend/app/auth/auth_routes.py
+++ b/src/backend/app/auth/auth_routes.py
@@ -315,7 +315,6 @@ async def temp_login(
     setting it as a cookie.
 
     Args:
-        request (Request): The incoming request object.
         email: email of non-osm user.
 
     Returns:
--- a/src/frontend/.gitignore
+++ b/src/frontend/.gitignore
@@ -122,5 +122,5 @@ dist
 /playwright-report/
 /blob-report/
 /playwright/.cache/
-
-playwright/.auth
+/e2e/.cache/
+e2e/.auth
--- a/src/frontend/e2e/01-create-new-project.spec.ts
+++ b/src/frontend/e2e/01-create-new-project.spec.ts
@@ -3,18 +3,14 @@
 
 import { test, expect } from '@playwright/test';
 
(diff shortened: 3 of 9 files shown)


spwoodcock:
This was a lot more complicated that I imagined, due to OpenStreetMap oauth restrictions:
- `https` must be used for all URLs, other than `127.0.0.1`.
- As we are testing in one container, but accessing the frontend in another, we cannot use `127.0.0.1`.
- This forces us to use `https` for the domain `fmtm.dev.test`, so we have to configure a certificate on the vite dev server and add the CA cert to the playwright container trusted cert chain.
- Documented the work in that direction here: https://github.com/hotosm/fmtm/pull/1920

**Instead**, in this PR I decided to make a simple workaround via temp auth for the React frontend:
- The auth is not actually used anyway, as we have `DEBUG=True` on the backend, so the `localadmin` user is used.
- This simply sets the temp auth cookie & authDetails in the frontend state, so we can continue as if we are logged in.

PR #1917, opened 2024-11-26
Add tests for submission routes

### LINKED ISSUE #1913: Add functional pytest Test cases for submission module
We don't have test cases for the submission related workflow. Since being able to see submission and its visualisation is one of the important component of FMTM. We need test cases to cover submission module. 

 Add functional pytest tests for submission related endpoints.

- [ ] Read submission
- [ ] Download submission on both csv and json format
- [ ] Get submission count (Should include submission both by selecting entities as well as new geopoints )
and many more

Anuj-Gupta4:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [x] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Example: Resolves #1913

## Describe this PR

A fixture for submission has been created and some tests have been added for submission routes.

### DIFF
--- a/src/backend/tests/conftest.py
+++ b/src/backend/tests/conftest.py
@@ -351,17 +351,12 @@ async def submission(client, odk_project):
     """
 
     with Client(config_path=odk_config_file) as client:
-        client.submissions.create(
+        submission_data = client.submissions.create(
             project_id=odk_project_id,
             form_id=odk_form_id,
             xml=submission_xml,
             device_id=None,
             encoding="utf-8",
-        )
-        submission_data = client.submissions.get(
-            project_id=odk_project_id,
-            form_id=odk_form_id,
-            instance_id=submission_id,
         )
 
     yield {
--- a/src/backend/tests/conftest.py
+++ b/src/backend/tests/conftest.py
@@ -19,6 +19,7 @@
 
 import json
 import os
+import uuid
 from io import BytesIO
 from pathlib import Path
 from typing import Any, AsyncGenerator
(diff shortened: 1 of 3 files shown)


src/backend/tests/conftest.py
```diff
@@ -275,6 +276,97 @@ async def odk_project(db, client, project, tasks):
     yield project
 
 
+@pytest_asyncio.fixture(scope="function")
+async def submission(client, odk_project):
+    """Set up a submission for a project in ODK Central."""
+    odk_project_id = odk_project.odkid
+    odk_credentials = odk_project.odk_credentials
+    odk_creds = odk_credentials.__dict__
```
spwoodcock, line 284: I think this is a Pydantic model, so the way to dump would be `.model_dump()` 👍

Using __dict__ is never advised

src/backend/tests/conftest.py
```diff
@@ -275,6 +276,97 @@ async def odk_project(db, client, project, tasks):
...
+    base_url = odk_creds["odk_central_url"]
+    auth = (
+        odk_creds["odk_central_user"],
+        odk_creds["odk_central_password"],
+    )
+
+    def forms(base_url, aut, pid):
+        """Fetch a list of forms in a project."""
+        url = f"{base_url}/v1/projects/{pid}/forms"
+        return requests.get(url, auth=aut)
```
spwoodcock, line 294: Does this work? I think there is a spelling mistake `aut` --> `auth`

spwoodcock (changes requested):
I love that this adds another test, so going to approve after minor fixes 👍

There's definitely more scope to test the different submission download types after, like CSV and geojson (this is in a different endpoint I think).

But I just wanted to comment that using basic auth via requests `auth=` isn't great. I'm surprised that ODK Central accepts this & support could possibly be removed in the future. In osm-fieldwork we have wrappers for the Central API (as does pyodk). Both of those libraries handle Central authentication via the sessions API (request a session token, then use the token to authenticate)

src/backend/tests/conftest.py
Anuj-Gupta4: it works. i have updated spelling for the sake of clarity.

Anuj-Gupta4:
@spwoodcock 
Should I add more tests for submission routes?

spwoodcock:
No need for that if this approach works 😄

I think it would be more valuable adding `pyodk` as a dependency in FMTM, as it might be useful elsewhere anyway:

`uv add pyodk`

The submission could be created with the pyodk client, then more tests written for json CSV and geojson downloads =)

I'll leave you to assess how much time this may take though and if it's worth it. Requirements relates to Tokha could be addressed instead if this will take a long time 👍 Either way is fine!

Anuj-Gupta4:
I haven't included the .pyodk_config.toml file in tests directory.

spwoodcock:
As the credentials are only used for testing, it's ok to commit that file. I think we need it for the tests to work (pyodk doesn't accept environment variables unfortunately 🥲)

src/backend/tests/conftest.py
```diff
@@ -275,6 +278,99 @@ async def odk_project(db, client, project, tasks):
...
+
+    with Client(config_path=odk_config_file) as client:
+        client.submissions.create(
+            project_id=odk_project_id,
+            form_id=odk_form_id,
+            xml=submission_xml,
+            device_id=None,
+            encoding="utf-8",
+        )
+        submission_data = client.submissions.get(
```
spwoodcock, line 361: I think submission.create returns the submission json data if you assign a variable, so there is probably no need for the extra `.get` call afterwards 👍

spwoodcock (changes requested):
This is excellent, well done!

Two tiny changes and all good to merge

spwoodcock:
Sorry Anuj! I didn't mean the dev server credentials!

The credentials you can use are for the locally installed version of ODK Central, which always has user `admin@hotosm.org` and password `Password1234`. The URL is `https://proxy` (which resolves to the proxy container, and then forwards traffic to the ODK Central container over HTTPS)

Anuj-Gupta4:
Oops my bad.

spwoodcock (approved):

Issue #1897, opened 2024-11-18
Small issues with form questions on mandatory XLSForm fields


spwoodcock:
**Issue 1**

In the XLSForm, I can select a geometry fine:

![image](https://github.com/user-attachments/assets/7fa8f693-4079-4df2-ac7f-ecb8704d4962)

But then it asks me another question after to select a building:

![image](https://github.com/user-attachments/assets/926f3d8e-c747-4789-9b2c-406a5ecea998)

I think related to recent additions to the form logic (about additional entities)

**Issue 2**

- If the user selects 'building does not exist', it would be nice if they could take a photo of the area.
  - Currently the form skips the photo question in this scenario.
- Perhaps we move the photo question outside the verification section as a standalone question, and always ask this question (photos are always useful)?
  - Other options to achieve this?

manjitapandey:
Issue 1: I can confirm this is due to additional entities being introduced and can be replicated when no optional field is added. 
Expected behavior: Only show the question if additional features are added during project creation.

Issue 2: It would be nice but we have to make this question optional since, most users on those case are not sure what to click if there is no feature.

PR #1919, opened 2024-11-26
Fix null value of additional entity when no additional entities are uploaded

Sujanadh:

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue: #1897 

## Describe this PR

This PR fixes the parsing of  `null` value of additional entities as a string in `form-data` when no additional entities are uploaded. It updates the body of generate-project-data by parsing null as a JSON instead of form data.

## Screenshots

N/A

## Review Guide

### DIFF
--- a/src/frontend/src/api/CreateProjectService.ts
+++ b/src/frontend/src/api/CreateProjectService.ts
@@ -190,34 +190,37 @@ const GenerateProjectFilesService = (url: string, projectData: any, formUpload: 
       let isAPISuccess = true;
       try {
         let response;
+
         const additional_entities: string[] =
           projectData?.additional_entities?.length > 0
             ? [projectData?.additional_entities?.[0]?.replaceAll(' ', '_')]
             : [];
+
         if (projectData.form_ways === 'custom_form') {
           // TODO move form upload to a separate service / endpoint?
           const generateApiFormData = new FormData();
           generateApiFormData.append('xlsform', formUpload);
+
           if (additional_entities?.length > 0) {
             generateApiFormData.append('additional_entities', additional_entities);
           }
+
           response = await axios.post(url, generateApiFormData, {
             headers: {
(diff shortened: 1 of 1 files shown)


NSUWAL123 (approved):

src/frontend/src/api/CreateProjectService.ts
```diff
@@ -190,34 +190,37 @@ const GenerateProjectFilesService = (url: string, projectData: any, formUpload:
...
-            additional_entities?.length > 0 ? additional_entities : null,
-          );
-          response = await axios.post(url, generateApiFormData, {
+          const payload = {
+            additional_entities: additional_entities.length > 0 ? additional_entities : null,
+          };
+          response = await axios.post(url, payload, {
             headers: {
-              'Content-Type': 'multipart/form-data',
+              'Content-Type': 'application/json',
```
spwoodcock, line 219: Well done finding the problem!
It makes sense now 😄

But here I think form data was used to provide a file name, so that can give a name to the additional entity list.

Does this work correctly when using a json upload?

spwoodcock (commented):
This may need to be solved on the backend instead, where we check for null and do not process the additional entities if so

spwoodcock:
But equally if this is working well, then we can keep this approach too! Just wanted to check

src/frontend/src/api/CreateProjectService.ts
Sujanadh: Yes, this works correctly. :tada:

Sujanadh: The form data was used because of custom xlsform upload, we don't need form data only for the additional_entities list.

Issue #1923, opened 2024-11-28
Validated vs Mapped Task graph is not populated in infographics page


manjitapandey:
**Describe the bug**
The Validated vs Mapped Task graph is not populated in infographics page. When mark a task as fully mapped and go on infographics page, the data is not updated and i see the message _No Tasks Validated or Mapped in the Last 30 Days_. 

**To Reproduce**

1. Go to 'project details page.'
2. Mark a task as fully mapped.'
3. Go to infographics page 
4. See error on mapped vs validated task graph

**Expected behavior**
The graph should be populated and updated when the task is mapped and validated. 

**Screenshots**
![Image](https://github.com/user-attachments/assets/4a0b41c4-b307-4e16-b6fb-76cc3bfbe908)

manjitapandey:
The chart is populated but the hover component shows 0 number of task mapped and validated. 

![Image](https://github.com/user-attachments/assets/b43c1a15-ea39-4b80-8821-237c869893d6)

PR #1924, opened 2024-11-28
Resolve route ambiguity for /tasks/activity endpoint

Sujanadh:

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue: #1923
```python
{
  "errors": [
    {
      "loc": [
        "path",
        "task_id"
      ],
      "msg": "Input should be a valid integer, unable to parse string as an integer",
      "error": "Input should be a valid integer, unable to parse string as an integer['path', 'task_id']"
    }
  ]
}
```

## Describe this PR

This PR resolves a routing issue in the `/tasks/activity` endpoint. A trailing slash (/) was added to the endpoint definition to ensure unambiguous route matching.

Previously, the `/tasks/activity` route conflicted with the dynamic route` /tasks/{task_id}`, causing the framework to misinterpret activity as a task_id, resulting in validation errors.

## Screenshots

N/A

## Review Guide

### DIFF
--- a/src/backend/app/tasks/task_routes.py
+++ b/src/backend/app/tasks/task_routes.py
@@ -55,6 +55,27 @@ async def get_tasks_near_me(
     return "Coming..."
 
 
+@router.get("/activity", response_model=list[task_schemas.TaskEventCount])
+async def task_activity(
+    project_id: int,
+    db: Annotated[Connection, Depends(db_conn)],
+    project_user: Annotated[ProjectUserDict, Depends(mapper)],
+    days: int = 10,
+):
+    """Get the number of mapped or validated tasks on each day.
+
+    Return format:
+    [
+        {
+            date: DD/MM/YYYY,
+            validated: int,
+            mapped: int,
+        }
+    ]
+    """
+    return await task_crud.get_project_task_activity(db, project_id, days)
+
+
 @router.get("/{task_id}", response_model=task_schemas.TaskOut)
 async def get_specific_task(
     task_id: int,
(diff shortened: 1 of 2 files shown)


spwoodcock:
One thing that is key to account for here is the route ordering in the xxx_routes.py file!

The `/projects/{project_id}/tasks/{task_id}` endpoint should be defined as late as possible,
then endpoints like `/projects/{project_id}/do-something` are defined above.

The routes are resolved in order in the file, so if they were the other way around then FastAPI would try to load: `/projects/{project_id}/do-something/{task_id}`, as `do-something` would be passed as the `{task_id}`

Sujanadh:
Yes, you are right. The actual issue is due to the ordering of endpoints. In addition to ordering, we need to define distinct endpoints with trailing slash '/' in every endpoint. What else would be the best approach to avoid this kind of issue? @spwoodcock  Yes, we can maintain ordering and place endpoints with placeholders in the path following other endpoints.

spwoodcock:
> Yes, you are right. The actual issue is due to the ordering of endpoints. In addition to ordering, we need to define distinct endpoints with trailing slash '/' in every endpoint. What else would be the best approach to avoid this kind of issue? @spwoodcock  Yes, we can maintain ordering and place endpoints with placeholders in the path following other endpoints.

We should be sure to **not** use a trailing slash for endpoints, but instead ensure the ordering is correct 👍

But note this is only ordering per **router**, so an alternative could be a different grouping of endpoints:
- We could better adhere to rest principals by putting nested tasks under the projects router. /projects/1/tasks/1
- Perhaps have an events router, using query params. /events?project_id=1&task_id=1
- Remove 'action' based endpoints, replacing them with a more generic version. For example on the organisations router, don't have a /approve endpoint, but instead a generic /status endpoint where we post the approved status. 
- Loads of options here for improvement...

I digress 😅 In summary, for now let's just fix the ordering!

spwoodcock (approved):

PR #1948, opened 2024-12-04
Feat/OSM login enforced for management, additional temp login option for mappers

### LINKED ISSUE #1911: Backend: OSM login enforced for management, additional temp login option for mappers
**Describe the bug**

- Follow up to #1890
- We have removed the temp login button from the management frontend.
The mapper frontend accepts both OSM and temp login options.
- The backend API remains the same though. 
- A user that logs in using 'temp token' in the mapper frontend, but then goes to the management frontend, will still show as logged in.

**Solutions**

- This isn't so simple as it first seems.
- We use separate cookies for each login type which is good.
- But as we use the same API for both frontends, how do we determine from which frontend the user makes the call from?
- The only approach I can think of for now is to gave seaprate endpoints:
  - We call different refresh endpoints, like `refresh_management` and `refresh_mapper` from each frontend. 
  - This would mean we can ignore the `_temp` cookie for the management frontend.

Of course, also open to other ideas!

Anuj-Gupta4:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to: #1911 

## Describe this PR

- created two seperate refresh endpoint for management and mapper frontends.
- enforced osm login only for management
- preferred osm login for mapper but also allowed temp login

### DIFF
--- a/src/backend/app/auth/auth_routes.py
+++ b/src/backend/app/auth/auth_routes.py
@@ -31,8 +31,11 @@ from app.auth.auth_schemas import AuthUser, AuthUserWithToken, FMTMUser
 from app.auth.osm import (
     create_jwt_tokens,
     extract_refresh_token_from_cookie,
+    extract_refresh_token_from_osm_cookie,
+    extract_token_from_cookie,
     init_osm_auth,
     login_required,
+    mapper_login_required,
     refresh_access_token,
     set_cookies,
     verify_token,
@@ -113,7 +116,7 @@ async def callback(
 
         # Get OSM token from response (serialised in cookie, deserialise to use)
         serialised_osm_token = tokens.get("oauth_token")
-        cookie_name = settings.FMTM_DOMAIN.replace(".", "_")
+        cookie_name = settings.cookie_name
         osm_cookie_name = f"{cookie_name}_osm"
         log.debug(f"Creating cookie '{osm_cookie_name}' with OSM token")
         response_plus_cookies.set_cookie(
@@ -140,11 +143,19 @@ async def logout():
     """Reset httpOnly cookie to sign out user."""
     response = Response(status_code=HTTPStatus.OK)
     # Reset all cookies (logout)
-    fmtm_cookie_name = settings.FMTM_DOMAIN.replace(".", "_")
+    fmtm_cookie_name = settings.cookie_name
     refresh_cookie_name = f"{fmtm_cookie_name}_refresh"
+    temp_cookie_name = f"{fmtm_cookie_name}_temp"
+    temp_refresh_cookie_name = f"{fmtm_cookie_name}_temp_refresh"
     osm_cookie_name = f"{fmtm_cookie_name}_osm"
 
-    for cookie_name in [fmtm_cookie_name, refresh_cookie_name, osm_cookie_name]:
+    for cookie_name in [
+        fmtm_cookie_name,
+        refresh_cookie_name,
+        temp_cookie_name,
+        temp_refresh_cookie_name,
+        osm_cookie_name,
+    ]:
         log.debug(f"Resetting cookie in response named '{cookie_name}'")
         response.set_cookie(
             key=cookie_name,
@@ -252,8 +263,8 @@ async def my_data(
     return await get_or_create_user(db, current_user)
 
 
-@router.get("/refresh", response_model=AuthUserWithToken)
-async def refresh_token(
+@router.get("/refresh/mgmt", response_model=AuthUserWithToken)
+async def refresh_mgmt_token(
     request: Request,
     current_user: Annotated[AuthUser, Depends(login_required)],
 ):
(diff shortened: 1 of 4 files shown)


src/backend/app/config.py
```diff
@@ -161,6 +161,11 @@ class Settings(BaseSettings):
 
     EXTRA_CORS_ORIGINS: Optional[str | list[str]] = None
 
+    @property
```
spwoodcock, line 164: This is pretty nice, pydantic computed isn't necessarily needed as we don't serialise the model, so a property works well here

spwoodcock (approved):
Haven't been able to test yet, but just reviewing the code this looks good!

src/backend/app/auth/auth_routes.py
```diff
@@ -252,8 +263,8 @@ async def my_data(
     return await get_or_create_user(db, current_user)
 
 
-@router.get("/refresh", response_model=AuthUserWithToken)
-async def refresh_token(
+@router.get("/refresh/mgmt", response_model=AuthUserWithToken)
```
spwoodcock, line 266: Explicit is probably preferred over abbreviated: this could be /refresh/management 👍

spwoodcock:
@Anuj-Gupta4 this is causing an issue: when I refresh the page on the management frontend (i.e. call the /refresh endpoint), my cookies seem to be cleared and I am no longer logged in.

The cookie needs to be refreshed correctly on the /refresh call, meaning the user is still logged in 😄 
Please prioritise this issue for the release today 🙏

spwoodcock:
Temporary login on the mapper frontend is also failing

spwoodcock:
Ignore the above - this works nicely for the management frontend 👍 

One small fix I will do now - if the token is invalid in the mapper frontend refresh, then we need to return a `HTTPStatus.UNAUTHORIZED` so the mapper frontend picks this up correctly (suggesting the user logs in)

spwoodcock:
OK so this mostly works, but temp login seems to fail for some reason

Sujanadh:
The project detail endpoint currently checks for mapper role in react frontend and same endpoint is used in the mapper frontend where it only checks for osm login in the login required. I think we might need a separate project endpoint in the mapper frontend with only the minimal required details of the project; I don't think we will require everything like in the project of the react frontend. 

The project details page of react frontend might require project mapper role not a global mapper, we can change this as per requirements. In the mapper frontend we will check only login and the token validation. Use login_required as dependency, handle both type of cookies in login required. To make it simpler, frontend needs to remove temporary cookie if i switch back to  react frontend. Then we will get whatever cookies is stored in both frontend and test its validation.
@spwoodcock @Anuj-Gupta4 

- #1969

spwoodcock:
The endpoint could be /projects/1/mapper-data and will use the `minimal` flag for getting the project

Sujanadh:
I don't find any significance of temp login at all. Its purpose is to allow anyone to contribute to the project; for that, we don't even need a login just for field submissions. Task locking and updating status will all require osm login. This temp login does have an issue with our access model, so creating an endpoint with minimal details will not leak any sensitive data. We can consider it for now.

And, about removing cookies when I redirect to the react frontend, it might create a requirement to login every time I go to mapper frontend. 
For this, I think we can auto log in with temporary login as a default, no need to do extra click to login for temporary users (mappers). We can add a pop-up to allow them to use osm login if they have an osm profile; otherwise, they can simply skip that.

What do you think? @spwoodcock

spwoodcock:
The only purpose of the 'temp login' is an extra layer of obfuscation really - it doesn't provide any security benefit.

So in theory I'm hoping it might reduce bots / spamming / scraping endpoints as it requires a login of some sorts, before anything else can be accessed.

It's not essential though! We considered it previous as we had everyone on one frontend, all using the same endpoints.
Now we have the separation, it's less important.

We could actually remove it entirely with this scenario:

- By default any user can access the mapper frontend (no login required).
- It has minimal info included / no sensitive data.
- It will have to include the ODK Central appuser token (QRCode data), meaning for now, anyone can contribute to the project (we can revisit how to restrict this in the future).
- We prompt the user to log in via OSM, if they want to get credit for their contributions.

What do you think?

spwoodcock:
This all said, for now we rely on endpoints to update task/entity statuses - it will be removed in the future, but is a requirement for now.

If we had no login at all, it would mean we have to update the access on all those endpoints too, or create separate 'mapper' endpoints 🤔

spwoodcock:
The idea you proposed above could work well:

Mapper:
- When a user accesses the mapper frontend and calls `/refresh/mapper`, we either attempt to refresh the OSM cookie if it exists, or we by default provide a 'temporary login' cookie.
- The user will be seamlessly logged in either way, but if they are using a temp cookie, we prompt them to log in via OSM in the frontend (if they want to be attributed).

Management:
- When they access the React frontend, we invalidate / remove any 'temp login' cookie they might have.

Sujanadh:
Yeah that could work for now. :+1:

Issue #921, opened 2023-10-18
Update documentation throughout repo


spwoodcock:
Replacement of https://github.com/hotosm/fmtm/issues/164

We currently have some minimal documentation, which could be greatly improved.
We would really welcome PRs for this 🙏 

## What needs to be done:

- Proof reading of current docs and correction.
- Reduce docs duplication - we currently have the same content duplicated in a few places.
- Improve general README, contributors guidelines, etc.
- Improve developer documentation with additional examples, FAQs (if you struggled setting up FMTM anywhere, feel free to document it!).
- Improve user documentation for how to use FMTM.
- Python docstrings throughout the backend, where missing (especially on routes).
- If you particularly like writing JSDoc docstrings, then it wouldn't harm having a few in the frontend too.

vinayakhegde1:
@spwoodcock I've imrpoved the contributing guidelines please do check

spwoodcock:
Reopening as q more general issue for contributors to help on.

It could definitely be split into sub issues based on the description though, to make it clearer what needs to be done

manjitapandey:
Agreed, Sorry I mistakely closed it adding fixes and this issue. Not intended.

Joshua-Tihabwangye:
hello, can i take on this isuue and complete it?

manjitapandey:
Sure, Is there anything specific in your mind that you are thinking to add? 

Otherwise, I was thinking you could also add few video tutorials on mapper's page as well. or the product tool video?

Joshua-Tihabwangye:
![Image](https://github.com/user-attachments/assets/0c519127-6494-4228-ba8a-13f41fb40f2e)

1.    I have read on the Field Mapper (The person who collects data)  section and I think  we should add in all the necessary apps by documenting there names and also providing the Links to those apps to enable the new or first time mappers to easily know the apps they are supposed to install which are compatible or which work hand in hand with FMTM.
2.  For easy access and to make the FMTM easy to work with also a link to the OSM registration site is needed below that step.

Joshua-Tihabwangye:
![Image](https://github.com/user-attachments/assets/f2cffa90-1c0c-4358-aba1-014c620e7e8b)

manjitapandey:
To do: 
1. Proof read the documentation for project manager and video manuals. 
2. Update with existing updates on workflow
3. Proofread if the user flow diagram is up to date or any section requires updating
4. Developer documentation updates if any part is outdated. 
[...]

PR #2012, opened 2024-12-20
Update mapping.md

### LINKED ISSUE #2010: Documentation on addition of user's metadata in odk collect so we get proper statistics
Currently, we haven't documented the workflow where field mappers  can set their username, email and device ID so that they can be used for better tracking user's contributions and applying necessary filters. 

**Expected Behavior**
 To generate meaningful statistics and insights, it is crucial to capture user metadata alongside submitted data. Though its the odk section of documentation, still we need to document it and encourage users to update theur details before filling up form on odk collect.

manjitapandey:
documented updating of users metadata in osm.

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [x] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Example: Fixes #2010 , #921 

## Describe this PR

Added more details on mapper's workflow
Addition of user's metadata in odk

## Review Guide

### DIFF
--- a/docs/manuals/mapping.md
+++ b/docs/manuals/mapping.md
@@ -1,7 +1,7 @@
 # Mapper Frontend Documentation
 
 The **Mapper's page** was developed to provide a simpler, more intuitive
-mapping experience. 
+mapping experience.
 
 ## Part (A): Setup For Mapping
 
@@ -9,12 +9,11 @@ mapping experience.
 
 - The first time you load the project, you should be prompted to download
   the custom `.apk` from the upper right corner of your browser window
-   usually represented by three vertical dots.
+  usually represented by three vertical dots.
 
 - Once downloaded, you should install the custom ODK Collect application.
   For installation, locate the downloaded ODK APK file usually found on
-  downloads section of your file manager and tap install. 
-
+  downloads section of your file manager and tap install.
 
 !!! tip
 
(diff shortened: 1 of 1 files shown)


docs/manuals/mapping.md
```diff
@@ -1,15 +1,19 @@
...
 mapping experience.
 
 ## Part (A): Setup For Mapping
 
 ### Step 1: Install The Custom ODK Collect Mobile App
 
 - The first time you load the project, you should be prompted to download
-  the custom `.apk` from the sidebar.
+  the custom `.apk` from the upper right corner of your browser window
+  usually represented by three vertical dots.
```
NSUWAL123, line 12: It's not `dot`. It shall be represented by a hamburger menu(might be technical)/menu/☰

spwoodcock:
Love the changes 👍

The stuff regarding adding these three metadata fields doesn't have to be done byvthr user though. They are automatically injected field by us & are always collected (username, email, device id)

manjitapandey:
I observed we don't add email and phone numbers automatically but usernames are added and I have provided them info about that too.

docs/manuals/mapping.md
```diff
@@ -19,7 +23,19 @@ mapping experience.
       You may have to enable installing from unknown sources in your device
       settings too.
 
-### Step 2: Access the Mapper Frontend
+### Step 2: Add metadata on odk application(Optional)
```
spwoodcock, line 26: I think this section should be removed, as its not something the user should have to configure manually. All the fields you mention are already included by default.

Sure we don't actually inject an email into the field yet (as its not available, we may never do this), but its enabled anyway

spwoodcock (changes requested):

docs/manuals/mapping.md
manjitapandey: But I am only inform the users that if their project needs email address, phone number or any other metadata that we don't have, they don't need to fill up in every form rather they can use this method. Otherwise its not compulsory and completely optional behavior for them. 
However, some projects need their user's phone number to contact them if needed. In such cases, they can add it in instruction and this document can help was my point of view.

spwoodcock: But we have no ability to configure this with the current code: the fields are always enabled by default. Perhaps this would make sense if we added the config option

spwoodcock: Maybe I'm misunderstanding this, as I have never tested the approach you mention

manjitapandey:
This has been included in field mapping guidelines document. So we are good to close it here for now.

PR #1998, opened 2024-12-17
Rename docker-compose.yml files

casdal:
## What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI
- [x] ❓ Other (files rename)

## Related Issue

Fixes #1769

## Describe this PR

Rename docker-compose.yml files and references to them.

## [optional] What gif best describes this PR or how it makes you feel?
😄

PR #2015, opened 2024-12-23
fix: compose file name for backend test stage

nischalstha9:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [X] 🤖 Build or CI

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -26,6 +26,7 @@ jobs:
       compose_command: pytest
       tag_override: ci-${{ github.ref_name }}
       coverage: true
+      compose_file: "compose.${{ github.ref_name }}.yaml"
     secrets: inherit
 
   frontend-unit-test:


Sujanadh (approved):

Sujanadh:
```
stat /home/runner/work/fmtm/fmtm/docker-compose.yml: no such file or directory
Error: Process completed with exit code 14.
```

spwoodcock:
Related to #1998. I did wonder if it might cause unintended issues somewhere, my apologies.

You need to include the `compose_file` var in the workflow I think

spwoodcock:
Or is the issue perhaps within another workflow that has the image build workflow nested within? 😅 That would be a hassle.

If so, it might require updating the gh-workflows repo default compose file name, incrementing as a breaking MAJOR version, and creating a release. Then updating workflows in FMTM.

Alternatively, just revert the PR I linked above and we can address this later 😃

nischalstha9:
I added the `compose_file` parameter but there was some error related to coverage. Not sure why it was not found in the test.
image
Meanwhile I have also created a PR reverting the commit #2021.
Will you review it.

spwoodcock:
Mystery to me!

If it was going to be missing, I would think it would happen after the upgrade to use `uv`.

As far as I remember nothing related to the build has been done recently 🤔

Issue #1975, opened 2024-12-11
Addition of highlighting effect on all rejected submissions


manjitapandey:
**Is your feature request related to a problem? Please describe.**
When validator rejects the submission, user are most likely to ignore if they are not highlighted and move on. 

**Describe the solution you'd like**
We should highlight the bad submissions of whole project area so that project managers as well as mapper in the field will be aware of rejected submission and they can resurvey the feature. 
- [x] API to pass all the bad submissions
- [x] Highlighting effect on all the bad submissions on both mapper frontend and management frontend

**Describe alternatives you've considered**

**Additional context**
This was discussed on in person team meeting.

manjitapandey:
when one submission is rejected a slight blue layer of feature is shown on mapper's view and clicking of this feature doesn't show the  pop-up to open form on odk. 

![Image](https://github.com/user-attachments/assets/ee0f02cb-8d79-415a-a221-629fa5567a06)

PR #2018, opened 2024-12-23
Entity pulse effect on rejected submissions

NSUWAL123:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
 - Related to #1975

## Describe this PR
- Pulse entity if submission to that entity marked as `hasIssues`
- Remove `rejected` entity submission state, retaining only `approved` and `hasIssues` state

## Screenshots
[Screencast from 17-1-25 03:32:55 अपराह्न +0545.webm](https://github.com/user-attachments/assets/9d454a2f-d276-4ee6-ae96-40348c21741b)

giphy (1)


NSUWAL123:
Now everything works fine but one small problem due to the current workflow: 
For example, if a submission has been made to an entity and it has been marked as bad then the obvious behavior would be that the entity's color would be red and the highlighting effect would be visible. Now the issue arises when another submission is made to that entity. The entity's color is changed to green but the highlighting effect remains until the submission is marked as `approved` since delete geom record API is triggered on marking the submission as approved.
So, basically in our current workflow, we are not able to delete geometry records once a submission is made to the entity whose previous submission has been marked `hasIssues` leading to the entity color green while having red pulse effect.

image

spwoodcock (approved):
Lots of changes, two frontend 😅
Good work!

Hopefully we don't have many changes that affect both the React and Svelte frontends at once.

Thanks for commenting on the case of two submissions too 👍 we can return to that in the future if needed

Issue #1960, opened 2024-12-09
Split by Sqaure can results to some tiny tasks.


Sujanadh:
**Is your feature request related to a problem? Please describe.**
Currently our split by square can results to tiny tasks which is not good. It will only increase the tasks number and records in table.

![Image](https://github.com/user-attachments/assets/17bcc2f0-2959-4f49-99ac-02483c38f46f)

**Describe the solution you'd like**
Merge them to nearset task if they fall under category of minimun threshold for example: less than the 20 % of task area (square generated). 

**Describe alternatives you've considered**
User changes the parameter until they get desired tasks area. 

What would be best approach here? 
@manjitapandey @ramyaragupathy  @spwoodcock

spwoodcock:
Agree! Merging into neighbouring squares is the way.

The approach in drone tm is good

https://github.com/hotosm/drone-tm/blob/1b7c81ded92fbf9417cd48e3ad32f043ddce7f0f/src/backend/app/tasks/task_splitter.py#L124

Although if it could be in SQL / PostGIS instead it would be even nicer!

Also I'm sure they have a reason, but I cant quite work out why they used that threshold calculation. If it makes sense to you, then use it,  but otherwise a 35% threshold could be appropriate

Sujanadh:
To use SQL/Postgis, we won't have tasks geom stored in db while splitting. For that, we might need to create a temporary table and do all the operations using PostGIS. And, I think it is still good to use temporary temp and do the operations.

Sujanadh:
@spwoodcock  One question: should we also consider tasks with no features?

Sujanadh:
So shall we fetch extracts and avoid tasks without features or only if users provide custom extracts?

spwoodcock:
If the user provides an extract, remove squares with no features.

If the user doesn't provide an extract, simply split by square (don't account for features) 👍

PR #2020, opened 2024-12-23
Add db parameter to split_by_square after recent changes to fmtm-splitter

Sujanadh:

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue: #1960

## Describe this PR

After a recent upgrade of split by square in the fmtm-splitter, it now requires an extra db parameter as a unicode string of db url, which fixes the mentioned issue. 

## Screenshots

N/A


src/backend/tests/conftest.py
```diff
@@ -230,6 +230,7 @@ async def task_event(db, project, tasks, admin_user):
         new_event = TaskEventIn(
             task_id=task.id,
             user_id=user.id,
+            username=user.username,
```
spwoodcock, line 233: I'm surprised this isn't calculated in the SQL, using the user id?

If you insert both the user id and username, then it opens the potential for inconsistency: inserting a username that does not match the user id

Sujanadh: Ah you are right , I got what you meant to say. I was trying to solve the test case failure issue, but it seems to have github action issue?

Issue #2025, opened 2024-12-25
Basemap generation failure for tms source


Anuj-Gupta4:
**Describe the bug**
Basemap generation fails for custom tms source.

**To Reproduce**

1. Go to any project
2. Click on basemap.
3. Choose custom tms as tile source and fill up all the fields.
4. Click generate
5. See error

**Expected behavior**
Expected a new record to be generated showcasing the new basemap generation.

**Screenshots**
![Image](https://github.com/user-attachments/assets/29c0f596-79a5-4d3b-b8fc-92486c837d52)

**Additional context**
Looks like a field for custom tms source is missing in the backend.

Anuj-Gupta4:
@NSUWAL123 check valid tile sources before using tiles-generate api after above pr is merged.

Issue #1953, opened 2024-12-04
Generate PMTiles automatically if custom TMS url is provided during project creation


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- Currently generating PMTiles for mapping is a manual process for the project manager.
- - It might be nice that if a custom TMS URL is provided during project creation, we automatically generate a PMTiles basemap archive.
- This would be available to mappers immediately.

**Describe the solution you'd like**

- Call the basemap creation API after generate project data is done.
- In future this will be a call to the basemap-api separate microservice.

PR #2027, opened 2024-12-25
Auto basemap generation if tms url is input during project creation

Anuj-Gupta4:

- [x] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- [x] Fixes: #2025 
- [x] Resolves: #1953 

## Describe this PR

- Added extra option to tile source of BasemapGenerate pydantic model.
- Added tms basemap generation during project data generation if tms url was present during project creation.

### DIFF
--- a/src/backend/app/projects/project_routes.py
+++ b/src/backend/app/projects/project_routes.py
@@ -946,7 +946,7 @@ async def generate_files(
 
     if project.custom_tms_url:
         basemap_in = project_schemas.BasemapGenerate(
-            tile_source="tms", file_format="pmtiles", tms_url=project.custom_tms_url
+            tile_source="custom", file_format="pmtiles", tms_url=project.custom_tms_url
         )
         org_id = project.organisation_id
         await generate_basemap(project_id, org_id, basemap_in, db, background_tasks)
--- a/src/backend/app/projects/project_routes.py
+++ b/src/backend/app/projects/project_routes.py
@@ -843,6 +843,7 @@ async def add_additional_entity_list(
 async def generate_files(
     db: Annotated[Connection, Depends(db_conn)],
     project_user_dict: Annotated[ProjectUserDict, Depends(project_manager)],
+    background_tasks: BackgroundTasks,
     xlsform_upload: Annotated[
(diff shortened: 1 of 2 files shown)


src/backend/app/projects/project_routes.py
```diff
@@ -942,6 +944,13 @@ async def generate_files(
             },
         )
 
+    if project.custom_tms_url:
+        basemap_in = project_schemas.BasemapGenerate(
+            tile_source="tms", file_format="pmtiles", tms_url=project.custom_tms_url
+        )
+        org_id = project.organisation_id
+        await generate_basemap(project_id, org_id, basemap_in, db, background_tasks)
```
spwoodcock, line 952: Its key that this returns immediately when the background task is submitted - just checking this works well upon testing?

spwoodcock (approved):
This is great to have in place 😄

In theory this should as minimal time to the project file creation API response time.

So the frontend will redirect and the project can be viewed. The basemap generation will lock a thread, but there should generally be another thread available to serve the project response afterwards.

This will have more guarantees in future with a separate [basemap-api](https://github.com/hotosm/basemap-api) microservice though

src/backend/app/projects/project_schemas.py
```diff
@@ -274,7 +274,9 @@ class ProjectUserContributions(BaseModel):
 class BasemapGenerate(BaseModel):
     """Params to generate a new basemap."""
 
-    tile_source: Annotated[Literal["esri", "bing", "google"], Field(default="esri")]
+    tile_source: Annotated[
+        Literal["esri", "bing", "google", "tms"], Field(default="esri")
```
spwoodcock, line 278: Good catch 👍

src/backend/app/projects/project_routes.py
Anuj-Gupta4: Yes this worked upon testing

Sujanadh (approved):

PR #2051, opened 2025-01-02
Load entities after page load

NSUWAL123:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
The initial mapper frontend page load is too slow due to the `/entities/statuses` API call which leads to a blank white screen for some time. 
The slow API is blocking the UI from being rendered.

## Describe this PR
The /entities/statuses API is called after the UI load rather than before which speeds up the UI load. 

## Screenshots
[Screencast from 2-1-25 09:42:48 पूर्वाह्न +0545.webm](https://github.com/user-attachments/assets/ec0727dc-ef77-43c2-be36-d325dda6d51c)

### DIFF
--- a/src/mapper/src/routes/[projectId]/+page.svelte
+++ b/src/mapper/src/routes/[projectId]/+page.svelte
@@ -25,6 +25,9 @@
 	import More from '$lib/components/more/index.svelte';
 	import { getProjectSetupStepStore, getCommonStore } from '$store/common.svelte.ts';
 	import { projectSetupStep as projectSetupStepEnum } from '$constants/enums.ts';
+	import { ShapeStream } from '@electric-sql/client';
+
+	const API_URL = import.meta.env.VITE_API_URL;
 
 	interface Props {
 		data: PageData;
@@ -44,7 +47,6 @@
 	const commonStore = getCommonStore();
 
 	const taskEventStream = getTaskEventStream(data.projectId);
-	const entityStatusStream = getEntityStatusStream(data.projectId);
 
 	// Update the geojson task states when a new event is added
 	$effect(() => {
@@ -98,9 +100,6 @@
 	}
 
 	onMount(async () => {
-		// In store/entities.ts
-		await entitiesStore.subscribeToEntityStatusUpdates(entityStatusStream, data.entityStatus);
-
 		// In store/tasks.svelte.ts
 		await taskStore.subscribeToEvents(taskEventStream);
 		await taskStore.appendTaskStatesToFeatcol(data.project.tasks);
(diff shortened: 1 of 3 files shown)


NSUWAL123:
@spwoodcock what do you think about this? Is it a good approach?

src/mapper/src/routes/[projectId]/+page.ts
```diff
@@ -36,15 +36,11 @@ export const load: PageLoad = async ({ parent, params, fetch }) => {
 	if (projectResponse.status === 404) {
 		throw error(404, { message: `Project with ID (${projectId}) not found` });
 	}
-	const entityStatusResponse = await fetch(`${API_URL}/projects/${projectId}/entities/statuses`, {
```
Comment (spwoodcock) [inline src/mapper/src/routes/[projectId]/+page.ts:39]: This part is key - its causing the slow render

src/mapper/src/routes/[projectId]/+page.svelte
```diff
@@ -98,19 +100,32 @@
...
 		await taskStore.subscribeToEvents(taskEventStream);
 		await taskStore.appendTaskStatesToFeatcol(data.project.tasks);
 	});
 
 	onDestroy(() => {
 		taskEventStream?.unsubscribeAll();
-		entityStatusStream?.unsubscribeAll();
 	});
 
+	$effect(() => {
```
Comment (spwoodcock) [inline src/mapper/src/routes/[projectId]/+page.svelte:112]: Does moving this from onMount change the load speed?

onMount should be called after the component is first rendered, so I'm surprised it would affect load speed.

If this works, its a bit hacky but we can leave it.
Perhaps I'm missing something key about svelte rendering order.

But if it has no impact, then onMount is a more natural place to init the ShapeStream, ensuring its only loaded once

spwoodcock (approved):

spwoodcock (changes requested):
Oops, see comments

src/mapper/src/routes/[projectId]/+page.svelte
Comment (NSUWAL123) [reply src/mapper/src/routes/[projectId]/+page.svelte:112]: Yeah, I also thought about the same thing. Last time I wrote this code block in onMount there was a problem when instantiating a new shape: `entitiesShape = new Shape(entitiesStream);`
Also today I tried to move the code block to onMount for a couple of hours but it didn't work. The status colors won't change. 
Is there something that I need to consider when shifting this code block to onMount due to which it's not working?

Comment (spwoodcock) [reply src/mapper/src/routes/[projectId]/+page.svelte:112]: Ah makes sense, I think intialising the ShapeStream may take some time (needs a web request), plus it happens outside of the Svelte lifecycle.

So the issue is probably if the component mounts, but the call to the electric service has not complete to initialise the ShapeStream, then creating the Shape with an empty ShapeStream will fail.

So by using an $effect, we trigger the Shape initialisation only after the ShapeStream is initialised.

Thanks for testing 👍

Comment (spwoodcock) [reply src/mapper/src/routes/[projectId]/+page.svelte:112]: (at least thats my hunch without digging into it too deeply!)

spwoodcock (approved):

Issue #1986, opened 2024-12-12
Create new DB table to store 'new' and 'bad' geoms


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- When a new geometry is mapped (submission made), its not displayed on the map.
- When a geom is marked bad, its hidden within the task. You have to click the task to see bad geom.

**Describe the solution you'd like**

- Database table recording new and bad geoms.
- Synced to user via electric.
- Bad geoms displayed flashing on top of the map.
- New geoms displayed either on top, or when you click a task area only?

**Describe alternatives you've considered**

- Ideally we could add new geoms to the entity list. Does this work currently?
- Will there be issues if its a different geometry type to the other geoms (collecting a point in a building entity list)?
- However this would mean the user has to refresh the page to view new geoms, plus we don't have a nice way to do this currently anyway. Loading geoms is based on the flatgeobuf, then entity properties are mapped to the geometries by id.
- Using a new db table opens the possibility for real time sync of new geoms.

NSUWAL123:
@spwoodcock This approach of storing bad geometries on a new table would also ease the effort on the frontend in pulsing the bad entities right #1975? 

Currently, on PR #2018, I have added another flatgeobuf layer passing the project extent and then filtering the features based on the entity status. This approach might be resource-extensive for low-end devices and makes the UI a bit laggy (especially management frontend) due to the constant fetch of the fgb URL. 
It would be very easy if we have an API that returns a feature collection of bad entities.

spwoodcock:
The approach you took is nice & more calls to the fgb file isn't a huge problem.

This db table will be required for the newly added geoms though, as we can't update the fgb file every time.

So if we have the table there anyway, it makes sense too also add the 'bad' geoms to it too.

Then we can display as a layer on top of all the others, ignoring the task area filtering

spwoodcock:
Plus also note the database table allows for the realtime update / loading bad geoms in the mapper frontend to be re-mapped immediately by the mapper still in the field

NSUWAL123:
The where clause in electric SQL doesn't support columns of the enum type. So I think we need to convert the type of column. I looked on electric SQL docs and couldn't find any solution.
https://electric-sql.com/docs/guides/shapes#where-clause
https://github.com/electric-sql/electric/issues/1709

CC: @spwoodcock , @Sujanadh

spwoodcock:
Damn! That's useful info (electric has a few limitations for SQL currently).

[...]

PR #2063, opened 2025-01-08
Update geom type to jsonb, int id to uuid in geometrylog table

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [X] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue:  #1986

## Describe this PR

This PR updates the existing table geometry log to have geometry field as jsonb type and id as UUID type. It is required as pglite doesn't support postgis; it may be in the near future. Now, using electric sql we can retrive geom as geojson without any functional support.

## Screenshots

N/A

## Review Guide

### DIFF
--- a/src/backend/migrations/003-jsonb-geom-geometry-log.sql
+++ b/src/backend/migrations/003-jsonb-geom-geometry-log.sql
@@ -21,7 +21,7 @@ ALTER TABLE geometrylog
 ALTER COLUMN id TYPE UUID USING gen_random_uuid(),
 ALTER COLUMN id SET DEFAULT gen_random_uuid();
 
-CREATE INDEX IF NOT EXISTS idx_geom_gin ON geometrylog USING gin (geom);
+CREATE INDEX IF NOT EXISTS idx_geom_gin ON geometrylog USING gist (geom);
 
 -- Commit the transaction
 COMMIT;
--- a/src/backend/migrations/003-jsonb-geom-geometry-log.sql
+++ b/src/backend/migrations/003-jsonb-geom-geometry-log.sql
@@ -1,0 +1,27 @@
+-- ## Migration to update geom column to jsonb type and int id to uuid.
+
+-- Start a transaction
+
+BEGIN;
+-- drop existing indexes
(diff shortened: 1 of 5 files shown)


src/backend/migrations/003-jsonb-geom-geometry-log.sql
```diff
@@ -0,0 +1,27 @@
...
+-- First, drop the default if it's currently not a UUID
+ALTER TABLE geometrylog
+ALTER COLUMN id DROP DEFAULT;
+
+-- Change the column type to UUID and set the default value
+ALTER TABLE geometrylog
+ALTER COLUMN id TYPE UUID USING gen_random_uuid(),
+ALTER COLUMN id SET DEFAULT gen_random_uuid();
+
+CREATE INDEX IF NOT EXISTS idx_geom_gin ON geometrylog USING gin (geom);
```
spwoodcock, line 24: From the PostGIS docs:

```
    GIN index lookups are about three times faster than GiST

    GIN indexes take about three times longer to build than GiST

    GIN indexes are moderately slower to update than GiST indexes, but about 10 times slower if fast-update support was disabled [...]

    GIN indexes are two-to-three times larger than GiST indexes
```

With that info, I would think GIST is more appropriate here.
An index is nice to have, but it's unlikely we will actually use it, due to passing directly to ElectricSQL.
So we won't see any of the advantages of GIN in this case.

Perhaps we would look at GIN indexes for other db tables though 👍

spwoodcock (approved):

spwoodcock:
Nice, thanks for including the revert migration too 🚀 

I'll just tweak the GIN index to GIST for now as commented & merge.

Let me know if you disagree / we can change it back 👍

spwoodcock:
Actually I'm totally wrong - apologies!

I forgot the `geom` field is now a JSONB, in which case you chose the correct index - gin is better for JSONB it seems 😄 

I may rename the `geom` field to `geojson` instead to help clarify 👍

Issue #1979, opened 2024-12-11
Add extra project advanced config option for new point collection: point, polygon, polyline


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- We can currently add a new point in FMTM and have it injected as the `new_feature` field in the XLSForm.
- The XLSForm field is currently `geopoint`.
- Ideally we can have this configurable, so we inject the field as geotrace (polygon) or geoshape (polygon), and allow the user to draw the relevant geometry in FMTM to inject.

This is a follow up issue from https://github.com/hotosm/fmtm/issues/1915

**Describe the solution you'd like**

- [x] Add a config option for this during project creation (frontend, the backend accepts the param)
- [x] Add a db field for `projects` to capture this, as an enum with three options **edit: done, added field new_geom_type with enum options `POINT`, `POLYLINE`, `POLYGON`.
- [x] Based on the enum value in the project JSON in the frontend, we change the terradraw geom type: point, polyline, polygon.
- [x] We may need to add buttons for deleting and modifying vertices if drawing more complex polygons.
- [x] The `append_mandatory_fields` function in osm-fieldwork should have the option to specify the new geom type, so the XLSForm has the relevant field type. https://github.com/hotosm/osm-fieldwork/pull/329 (released in v0.18.0)

**Additional context**

- This may be dependent on moving the mandatory XLSForm questions sheet out of the XLS file and directly into code (to make things easier).

manjitapandey:
> We can currently add a new point in FMTM and have it injected as the new_feature field in the XLSForm.

Is this new_feature becomes an entity and can we use it in visualising the bad submissions?? 

cc: @spwoodcock , @Sujanadh

spwoodcock:
Yes we can for sure!

Either:

- Make an API call to get entity ids and statuses. Then fetch the geoms via their entity id.

- (preferred) insert the new geoms and bad geoms into an FMTM database table. Have the table reactive via electric again. The geometries will be loaded on the map real time as they are mapped or marked bad.

spwoodcock:
I'm working on it - ticking off the tasks in the description as they are done

manjitapandey:
@spwoodcock Does this needs frontend side of action now?? I am not sure as i am not being able to draw new polygon on FMTM mapper's page.
cc: @NSUWAL123

spwoodcock:
> @spwoodcock Does this needs frontend side of action now?? I am not sure as i am not being able to draw new polygon on FMTM mapper's page.
[...]

PR #2082, opened 2025-01-10
Add capability to draw new polygon and linestring geoms

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Related to #1979

## Describe this PR

- We should be able to collect new polygon and linestring geoms with this update.
- By default all new projects are type 'polygon' / 'geoshape'.
- The type of geometry to draw needs to be configured at project creation.
- Setting the `new_geom_type` param will also modify the XLSForm as needed to the correct geometry type.

Future todo:

- [ ] Add field / selection to project creation workflow in frontend.
- [ ] Work out why the polygon drawing is a bit flaky:
  - They seem to complete on the third vertice being drawn.
  - It's possible to draw more by holding the click, dragging the map slightly, then releasing the click.
  - Probably some conflict with the default MapLibre click handler?
- Note that I didn't add any extra toolbar buttons to the map. It's not possible to edit verticies, as this is probably done on a phone and that would be too fiddly. The user will just have to create a new geom if they make a mistake.

## Screenshots

image

## Review Guide

- Create a new project.
- Create new polygons.
- Test they work in ODK Collect.

### DIFF
--- a/src/backend/app/central/central_crud.py
+++ b/src/backend/app/central/central_crud.py
@@ -34,7 +34,7 @@ from pyxform.xls2xform import convert as xform_convert
 
 from app.central import central_deps, central_schemas
 from app.config import settings
-from app.db.enums import EntityState, HTTPStatus
+from app.db.enums import DbGeomType, EntityState, HTTPStatus
 from app.db.models import DbXLSForm
 from app.db.postgis_utils import (
     geojson_to_javarosa_geom,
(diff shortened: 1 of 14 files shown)


Anuj-Gupta4 (approved):
I tested this locally using odk collect in mobile phone.
It works exactly as expected for new polygons. I didn't see any way to create new linestrings yet.
Approving for polygons only.

Issue #1985, opened 2024-12-12
Addition of distance constraints between user's geolocation and feature selected for mapping


manjitapandey:
**Is your feature request related to a problem? Please describe.**
Currently, user can select the feature far away from their current location. This causes enumerators to share incorrect data sometimes. 

**Describe the solution you'd like**
Restrict the user to select the feature that are far away from their current location. For that we need to necessarily ask for geolocation on mapper's page and if users click on feature far away and click on map on odk. The error message saying the feature you are trying to map is xx km away from your current location.

spwoodcock:
See the linked dup issue for some additional info

manjitapandey:
As you mentioned in duplicate issue, 

> Instead open a confirmation dialog / modal: 'The created geometry is far away from your current location. Are you sure you wish to proceed?'

Are you suggesting just the warning message or complete blocking to submission? 

Showing the warning is good but I think we should make the feature a bit dynamic my adding the feature of whether project creator wants to add restriction or not. 
During project creation, we can ask project creator if they want to restrict the selection of feature away from their current location. If project creator ticks on the restriction, they can add their own proximity distance as well. 
We can cover more use cases this way for both open projects and strictly monitored projects. 

What are your opinion? @spwoodcock , @ramyaragupathy

spwoodcock:
I don't think this feature should touch the backend in any way & shouldn't need configuration. 

Its just a helpful warning / indicator to the user in the field.

Adding a dialog to say 'are you sure you wish to proceed' isn't blocking - its just a warning.

The reason we need this and not just a toast warning notification is that clicking to create a new feature currently opens ODK Collect immediately, so the user will not see any warning

manjitapandey:
Talked to tokha team, If we don't wanna implement it in FMTM, we can only build this feature for tokha using 100 meter threshold for now.

spwoodcock:
If it's something that they need, perhaps others will need it in future too.

I think we should solve this like:
- [ ] Add a new db field `projects.geo_restrict_distance_meters` as an int2, default `50`.
- [ ] Add a new db field `projects.geo_restrict_force_error` as a boolean, default `false`.
- [ ] Return these fields in the project response.
- [ ] If the field `geo_restrict_force_error` is false, have a warning on the distance constraint.
- [ ] If the field `geo_restrict_force_error` is true, have a warning on the distance constraint.

I can add these fields on the db / backend today to faciliate.

What do you think @manjitapandey @NSUWAL123 ?

[...]

PR #2084, opened 2025-01-13
Distance constraint add on frontend

NSUWAL123:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #1985

## Describe this PR
Validate the distance between user location and entity if the distance constraint is `true`.  
If the distance constant is set to true, the user must enable their device location & must be within the distance constraint set.

## Screenshots
1. Geolocation must be enabled if the distance constant set
image_720

2. Show a message if the user now within the range
image_720

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
   # Lint / autoformat: Python code
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: "v0.8.6"
+    rev: "v0.9.1"
     hooks:
       # Run the linter
       - id: ruff
@@ -21,7 +21,7 @@ repos:
 
   # Deps: ensure Python uv lockfile is up to date
   - repo: https://github.com/astral-sh/uv-pre-commit
-    rev: 0.5.14
+    rev: 0.5.18
     hooks:
       - id: uv-lock
         files: src/backend/pyproject.toml
--- a/contrib/just/start/Justfile
+++ b/contrib/just/start/Justfile
@@ -32,7 +32,7 @@ backend:
   cd {{justfile_directory()}}
   docker compose up -d api
 
-# Start backend API only
+# Start backend API without docker
 [no-cd]
 backend-no-docker:
   #!/usr/bin/env sh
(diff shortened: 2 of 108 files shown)


spwoodcock (approved):

spwoodcock:
I updated the logic to be:

```js
	// Check if drawn geometry is within set distance constraints from user
	const isDistanceConstaintValid = (): boolean => {
		const coordTo = entitiesStore.selectedEntityCoordinate?.coordinate;
		const coordFrom = entitiesStore.userLocationCoord;

		// Geolocation not enabled, warn user
		if (!coordFrom) {
			alertStore.setAlert({
				message:
					'This project has a distance constraint set. Please enable device geolocation for optimal functionality',
				variant: 'warning',
			});
			return false;
		}

		const entityDistance = distance(coordFrom as Coord, coordTo as Coord, { units: 'kilometers' }) * 1000;
		if (entityDistance && entityDistance > projectData?.geo_restrict_distance_meters) {
			// Feature is far away from user, warn user
			alertStore.setAlert({
				message: `The feature must be within ${projectData?.geo_restrict_distance_meters} meters of your location`,
				variant: 'warning',
			});
			if (projectData?.geo_restrict_force_error) {
				// Not valid coord, prevent user from continuing
				return false;
			}
			// Valid coord: it's outside of range, but only 'warn' user
			return true;
		}

		// Valid coord
		return true;
	};
```

The goal is to:
- Warn the user if geo_restrict_force_error=False
- Error / stop the user if geo_restrict_force_error=True

Hope this makes sense! Please check it works as intended, then merge 🙏

NSUWAL123:
So do we want user's to restrict map the feature if their device location is turned off? 
If yes then we shall update the toast message and let the user map the feature else we shall only update the toast message to `Please enable device geolocation to map this feature`.
image

spwoodcock:
I didn't test the logic, so perhaps made a mistake.

But you make a good point.
We can't **force** users to enable geolocation, hence its impossible to actually enforce this constraint.

Anyway, this should happen:
- No geolocation, no force geo constraint: ignore / do not show warning.
- No geolocation, force geo constraint: prompt to enable geolocation.
- Geolocation, no force geo constraint: warning if triggered.
- Geolocation, force geo constraint: error if triggered.

Does that sound right?

manjitapandey:
Yes the conditions above sounds right. 

also In such cases, For Tokha projects, field managers and monitoring team will always be available on call to assist mappers in field. For other project this may not be relevant so only warnings and prompt to enable geolocation should work but this case is different.

spwoodcock:
Thanks!

Let's block the mapping if geolocation is not enabled the @NSUWAL123 👍

NSUWAL123:
Updated logic to: 
- No geolocation, no force geo constraint: ignore / do not show warning
- No geolocation, force geo constraint: prompt to enable geolocation
- Geolocation, no force geo constraint: ignore / do not show warning
- Geolocation, force geo constraint: error if triggered

NSUWAL123:
Sure!  It makes sense

NSUWAL123:
Updated the flow:
image_720

spwoodcock (approved):

Issue #1999, opened 2024-12-17
GDPR Compliance: warn users their account will be deleted after two years inactivity


spwoodcock:
**Describe the bug**

- We must warn users about inactivity and delete their account after two years inactivity.
- If they log in again, then this is reset.
- If they don't respond / login, we delete their user.

**Solution**

- Warn them three times, perhaps three weeks prior to the deletion?
- Warnings via email / OSM message?
- Write out a default email for this / get wording right for GDPR compliance.

Anuj-Gupta4:
We first need to decide what classifies as users being active or inactive in order to figure out how this issue can be addressed.

spwoodcock:
Do you think adding a last_active_at field to the users table would be a good approach?

Then update it at every login

Anuj-Gupta4:
I was thinking the same thing. This probably works the best.

manjitapandey:
Scheduling part is remaining for this.

spwoodcock:
@Anuj-Gupta4 I promised I would look at the scheduler - doing that now 👍

PR #2088, opened 2025-01-15
Warn users and delete user accounts after period of inactivity 

Anuj-Gupta4:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related to: #1999 

## Describe this PR

added a new field to track users last active date based on login in users table
added a task endpoint to warn users before deletion and delete users if the inactivity threshold has been met.


Anuj-Gupta4:
We will probably need a scheduler to be able to periodically warn and delete inactive users.

spwoodcock (changes requested):
If we add a db connection on each login_required call, I think this means the update is done for every single endpoint called.

Instead, would it be better to add this logic to the /me endpoint only, when the login details are retreived (on first page load)?

Could you simply add to the select SQL `.one` logic something like `UPDATE users SET last_login_at = NOW() WHERE id = %(user_id)s;`

spwoodcock:
You could probably merge this into the existing logic, adding a RETURNING clause to get the user after update

Sujanadh:
add migration file in fmtm_base_schema as well

src/backend/app/db/models.py
```diff
@@ -160,59 +160,69 @@ class DbUser(BaseModel):
...
+                    {"username": user_identifier},
+                )
+                user_row = await cur.fetchone()
+
+            if user_row is None:
+                raise KeyError(f"User ({user_identifier}) not found.")
+
+            user_identifier = user_row.id
+
         async with db.cursor(row_factory=class_row(cls)) as cur:
```
Sujanadh, line 186: Can we create a separate function to update user instead of writing it in `one`?

On Second thought
I don't think we would need to add update in `one` unless we are not using `get_or_create_user` function. May be we could get rid of this function and use one instead then. Otherwise we don't need to use update in `one`, since the last_login will be updated from `get_or_create_user` function whenever user logs in.

Anuj-Gupta4: true. lets just update the field only when me endpoint is executed.

spwoodcock (approved):
Excellent work! 🎉

Sujanadh (approved):

Issue #2093, opened 2025-01-16
Backend: API endpoint to change global user roles for admin


manjitapandey:
**Is your feature request related to a problem? Please describe.**
As discussed in the last user management meeting and followed by #1083 , currently we have 3 global roles namely, mapper, admin and blocked users. We lack specific endpoint to change global user roles. 

**Describe the solution you'd like**
An API endpoint to change global user roles for admin.

**Describe alternatives you've considered**

**Additional context**

spwoodcock:
Rather than an endpoint specific to admins, this should have a field admin level permission set.

Roles only really make sense related to a project, so this should take a project id.

Project managers, assistant admins, field admins, should all be able to update roles.

But they can only assign roles of equal level or below their own role!

spwoodcock:
Sorry I read your description more carefully.

The global roles are only there to create new admins or block users, which are infrequent activities.

I guess an endpoint and page for this could make sense, but I don't imagine it will be used much.

PR #2117, opened 2025-01-27
Add endpoint to change global user roles

Anuj-Gupta4:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related to: #2093 

## Describe this PR

Allow admins to change the global role of a user

### DIFF
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -78,7 +78,7 @@ if TYPE_CHECKING:
         ProjectUpdate,
     )
     from app.tasks.task_schemas import TaskEventIn
-    from app.users.user_schemas import UserIn
+    from app.users.user_schemas import UserIn, UserUpdate
 
 
 def dump_and_check_model(db_model: BaseModel):
@@ -315,17 +315,34 @@ class DbUser(BaseModel):
         return new_user
 
     @classmethod
-    async def update_role(cls, db: Connection, user_id: int, new_role: UserRole):
+    async def update(
+        cls, db: Connection, user_id: int, user_update: "UserUpdate"
+    ) -> Self:
         """Update the role of a specific user."""
+        model_dump = dump_and_check_model(user_update)
+        placeholders = [f"{key} = %({key})s" for key in model_dump.keys()]
+        sql = f"""
+            UPDATE users
+            SET {", ".join(placeholders)}
+            WHERE id = %(user_id)s
+            RETURNING *;
+        """
+
         async with db.cursor(row_factory=class_row(cls)) as cur:
             await cur.execute(
-                """
-                UPDATE users SET role = %(new_role)s WHERE id = %(user_id)s
-                RETURNING *;
-                """,
-                {"new_role": new_role.value, "user_id": user_id},
+                sql,
+                {"user_id": user_id, **model_dump},
             )
-            return await cur.fetchone()
+            updated_user = await cur.fetchone()
+
+        if updated_user is None:
+            msg = f"Failed to update user with ID: {user_id}"
+            log.error(msg)
+            raise HTTPException(
+                status_code=HTTPStatus.INTERNAL_SERVER_ERROR, detail=msg
+            )
+
+        return updated_user
 
 
 class DbOrganisation(BaseModel):
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -314,6 +314,19 @@ class DbUser(BaseModel):
 
         return new_user
 
+    @classmethod
+    async def update_role(cls, db: Connection, user_id: int, new_role: UserRole):
+        """Update the role of a specific user."""
+        async with db.cursor(row_factory=class_row(cls)) as cur:
+            await cur.execute(
+                """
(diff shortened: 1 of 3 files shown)


src/backend/app/users/user_routes.py
```diff
@@ -56,6 +56,17 @@ async def get_user_roles(current_user: Annotated[DbUser, Depends(mapper)]):
     return user_roles
 
 
+@router.get("/change-role", response_model=user_schemas.UserOut)
+async def change_user_role(
+    user: Annotated[DbUser, Depends(get_user)],
```
spwoodcock, line 61: This seems like an extra db call for no reason - should it be `user_id: int` instead?

src/backend/app/db/models.py
```diff
@@ -313,6 +313,19 @@ async def create(
 
         return new_user
 
+    @classmethod
+    async def update_role(cls, db: Connection, user_id: int, new_role: UserRole):
```
spwoodcock, line 317: Instead of a specific method for updating a single field, would it be possible to have a generic `.update` method that accepts a pydantic schema to update any field specified for the user?

Sticking to this approach for all models should reduce code duplication 👍

See the other model update methods for examples

spwoodcock (changes requested):

src/backend/app/users/user_routes.py
Anuj-Gupta4: Using the dependency would raise an error if the user doesn't exist for the entered user id.

src/backend/app/db/models.py
Anuj-Gupta4: ok

src/backend/app/users/user_routes.py
Sujanadh: - Use `user_id` in path `{user_id}/change-role`
- Instead of using UserUpdate (which you have created, which is fine; you can leave it as it), just to update roles, you can use the already existing `UserRole`: `user_schemas.UserRole(role=new_role)` like this.
So that there won't be necessary db calls, the code will look clean and optimised.

Sujanadh (changes requested):

src/backend/app/users/user_routes.py
Sujanadh: If the user doesn't exist, you can raise an error from db update?

Anuj-Gupta4: ok

src/backend/app/users/user_schemas.py
```diff
@@ -34,6 +34,24 @@ class UserIn(DbUser):
...
+    registered_at: Annotated[Optional[AwareDatetime], Field(exclude=True)] = None
+    api_key: Annotated[Optional[str], Field(exclude=True)] = None
+    tasks_mapped: Annotated[Optional[int], Field(exclude=True)] = None
+    tasks_validated: Annotated[Optional[int], Field(exclude=True)] = None
+    tasks_invalidated: Annotated[Optional[int], Field(exclude=True)] = None
+    project_roles: Annotated[Optional[dict[int, ProjectRole]], Field(exclude=True)] = (
+        None
+    )
+    orgs_managed: Annotated[Optional[list[int]], Field(exclude=True)] = None
+    is_email_verified: Annotated[Optional[bool], Field(exclude=True)] = None
```
spwoodcock, line 52: I like this approach!

The fields api_key and is_email_verified should probably be included / updatable though

spwoodcock (approved):

src/backend/app/users/user_schemas.py
Sujanadh: Just to update roles, we wouldn't need an extra pydantic model with excluding all those fields. But yeah, when we need an api to update user info, this pydantic could be used. For the fields `api_key` and `is_email_verified` that can be removed from pydantic model, they will be used from DbUser and will be updatable too. 
cc. @Anuj-Gupta4

src/backend/app/users/user_routes.py
```diff
@@ -58,6 +58,18 @@ async def get_user_roles(current_user: Annotated[DbUser, Depends(mapper)]):
     return user_roles
 
 
+@router.patch("/{user_id}/change-role", response_model=user_schemas.UserOut)
```
spwoodcock, line 61: Actually, sorry I missed this before, but now this is a generic user update, we can probably remove `change-role` and have this as simply a `.patch("/{user_id}"`.

Its always good to use this single endpoint as much as possible, to avoid creating specific endpoints for each update task (/change-role, /approve-user, /update-user, etc)

spwoodcock (changes requested):

src/backend/app/users/user_routes.py
Sujanadh: In that case, we can change the endpoint and make it generic user update not just relying on role like you have mentioned. :+1:

spwoodcock (approved):

Issue #2120, opened 2025-01-28
Statuses API is failing for project created before 2025.1.0 release


manjitapandey:
**Describe the bug**
For the project created before 2025.1.0 release, the statuses API for all projects are failing. However, those projects were completely functional before the changes were pushed. I have tested on staging server with project ID 30. The project exists on odk server though. 

**To Reproduce**

1. Go to 'stage.fmtm.hotosm.org '
2. Click on 'any project created before latest release'
3. View the network panel

**Expected behavior**
The statuses API should function properly and all entities should load.

**Screenshots**

![Image](https://github.com/user-attachments/assets/bd545d83-4122-4d98-8111-4aee0c092e98)

**Smartphone (please complete the following information):**

**Additional context**
This issue is there on tokha server too which makes it high priority issue.

Anuj-Gupta4:
This happened because old projects do not have submission_ids field in their entities. This problem will not occur in new projects.

spwoodcock:
Ah nice, thanks @Anuj-Gupta4 

This is a breaking change between releases then 😅 

Could we do a simple workaround for now like:

(pseudocode, needs to be updated to work)
```python
if project.created_at < 24/01/2024
    entities = await central_crud.get_entities_data(
        project.odk_credentials,
        project.odkid,
        fields="__system/updatedAt, osm_id, status, task_id"
    )
else:
    entities = await central_crud.get_entities_data(
        project.odk_credentials,
        project.odkid,
    )
```

probably not ideal, and I'm not sure if this will break anything else? but could be worth testing just to get things working again 👍

spwoodcock:
Actually, saying that, it's not the best of ideas.

The optimal solution would be updating the entity lists (datasets) for all old projects to have the `submission_ids` field 👍 

That would be a migration Python script to iterate all projects before 24/01/2024 and add the extra property to the associated entity list

spwoodcock:
Extra info: I am getting a 500 error viewing the entities in ODK Central UI too!

Logs:

```
central-1  | ::ffff:172.20.0.9 - - [28/Jan/2025:11:30:08 +0000] "GET /v1/projects/1/datasets/features HTTP/1.0" 200 1585
central-1  | error: column entity_defs.sourceId does not exist
central-1  |     at Parser.parseErrorMessage (/usr/odk/node_modules/pg-protocol/dist/parser.js:287:98)
[...]

PR #2126, opened 2025-01-29
feat(migration): add a file to add dataset property to projects

Anuj-Gupta4:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related to: #2120 

## Describe this PR

Added a migration file that will add a dataset property to projects in order to fix statuses api

### DIFF
--- a/src/backend/migrations/add-submission-id-property.py
+++ b/src/backend/migrations/add-submission-id-property.py
@@ -1,0 +1,49 @@
+"""Add dataset property for submission ID."""
+
+import asyncio
+
+from psycopg import AsyncConnection
+
+from app.central import central_deps, central_schemas
+from app.config import settings
+from app.db.models import DbProject
+
+
+async def add_submission_id_to_project(
+    odk_creds: central_schemas.ODKCentralDecrypted,
+    odk_id: int,
+) -> list:
+    """Add dataset property 'submission_ids' for each project."""
+    try:
+        async with central_deps.get_odk_dataset(odk_creds) as odk_central:
+            await odk_central.createDatasetProperty(
+                odk_id,
+                "submission_ids",
+            )
+    except Exception:
+        msg = f"Failed for ODK project ({odk_id}): e"  # usually "already exists"
+        print(msg)
+
+
+async def add_submission_id():
+    """Adds dataset property for all projects."""
+    async with await AsyncConnection.connect(
+        settings.FMTM_DB_URL.unicode_string(),
+    ) as db:
+        projects = await DbProject.all(db)
+        if not projects:
+            print("No projects found.")
+            return
+
+        for project in projects:
+            print("")
+            print("--------------------------")
+            print(f"------- Project {project.id} -------")
+            print("--------------------------")
+            print("")
+            await add_submission_id_to_project(project.odk_credentials, project.odkid)
+    print("success")
+
+
+if __name__ == "__main__":
+    asyncio.run(add_submission_id())


Anuj-Gupta4:
This mostly works. But I have noticed that odk credentials is not always being populated in projects after creation and it could fail if that is the case.
Also, project creation is currently not working in development to be able to test this.
Could someone take a look at the production database?
cc: @Sujanadh @spwoodcock

Sujanadh:
So there is an issue of ODK credentials not being populated in the project; in such a case, we are retrieving credentials from the env. It works until the credentials of env and credentials used during project creation (retrieved from the organisation or input); if the credentials are different, it might cause issues. We need to make sure that credentials are populated in the project table; if not, we retrieve them from the organisation. If the organisation also does not have credentials, then we use them from env.

spwoodcock:
Thanks for looking into this @Sujanadh 🙏

You are right, the order of credentials to look for:

project creds --> org creds --> .env file creds as a final fallback (this shouldn't happen ideally)

spwoodcock (approved):
Looks good!
I'll merge this, then we can test for existing projects on the dev server 😄

src/backend/migrations/add-submission-id-property.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if not projects:
+            print("No projects found.")
+            return
+
+        for project in projects:
+            project["odk_creds"] = get_odk_creds(project)
+            print(f"\n------- Project {project['id']} -------\n")
+
+            async with central_deps.get_odk_dataset(
```
spwoodcock, line 66: Actually I didn't review this closely enough.
This works fine as a migration, as we don't have loads of projects yet.

But just note that the `get_odk_dataset` internally uses `OdkDataset` from osm-fieldwork.
As part of the `__aenter__` method for that class (called when used with a context manager: `async with ...`), the authentication with ODK Central will happen, creating a new session.

With the current code, for each project we start a new connection and session with ODK Central, which isn't ideal when done in quick succession without any `sleep` time between.

An alternative would be:

```python
from time import sleep
async with central_deps.get_odk_dataset(project["odk_creds"]) as odk_central:
    for project in projects:
        project["odk_creds"] = get_odk_creds(project)
        print(f"\n------- Project {project['id']} -------\n")
        await odk_central.createDatasetProperty(
            project["odkid"],
            "submission_ids",
        )
        # wait 1 second before next api call
        sleep(1)
```

spwoodcock: lol actually I should look again!

of course, each project has it's own odk creds, so needs a new session unfortunately!
ignore me 😅

spwoodcock: ooh one last thing, in SQL, always good to cast the types explicity:
```sql
WHERE p.created_at < '2024-01-24 00:00:00.000+05:45'::timestamptz;
```
otherwise I don't think that equivalence check will work as expected

spwoodcock: Also, I made in a mistake in the comment of the related issue: it's 2025, not 2024 😅 
Will update with the small tweaks 👍

Anuj-Gupta4: I see. Thanks for fixing.🤔

Issue #1603, opened 2024-06-27
Automatically unlock tasks that have been locked for more than 3 days


spwoodcock:
Blocker https://github.com/hotosm/fmtm/issues/1604

**Is your feature request related to a problem? Please describe.**

- A task may be locked by a user, then they don't actually map, or forget to unlock.
- The project admin can unlock the task, so this is not a complete blocker.
- But other mappers cannot unlock the task.

**Describe the solution you'd like**

- Automatically unlock the task on the database after 3 days have elapsed.

**Solution**

This is a relatively easy solution via pg_cron directly on the database.

Create a function:

```sql
CREATE OR REPLACE FUNCTION unlock_longstanding_locked_tasks()
RETURNS void LANGUAGE plpgsql AS $$
DECLARE
    svc_user_id INT;
BEGIN
    SELECT id INTO svc_user_id
    FROM users
    WHERE username = 'svcfmtm';

    INSERT INTO task_history (project_id, task_id, action, action_date, user_id)
    SELECT project_id, task_id, 'READY', NOW(), svc_user_id
    FROM (
        SELECT project_id, task_id, action, action_date,
               ROW_NUMBER() OVER (PARTITION BY project_id, task_id ORDER BY action_date DESC) as row_num
        FROM task_history
    ) subquery
    WHERE subquery.row_num = 1
      AND subquery.action = 'LOCKED_FOR_MAPPING'
      AND current_timestamp - subquery.action_date >= interval '3 days';
END;
$$;
```

Create the cron (runs every day at midnight):

```sql
SELECT cron.schedule('0 0 * * *', 'SELECT unlock_longstanding_locked_tasks();');
```

spwoodcock:
The logic could be better here:

- Now we have odk_entities table we can check for any status updates within the task area, I.e. if progress is made.

- As a manager could assign a task (lock it for a user and notify them), we should account for this. If the latest event is ASSIGN, then unlock if it was three days ago.

- But if the latest event was MAP and there has been no progress, then unlock within an **three hours**.

- Perhaps there is a way to trigger this during an endpoint call, instead of as a cron (then this handles the issue of only wanting to run on **active** projects too).

manjitapandey:
I agree with your idea of checking the last activity but i guess only 3 hours will be less for the users. We can implement the logic if there is room to increase the timeout easily in future.

Anuj-Gupta4:
[...]

PR #2147, opened 2025-02-04
Add task scheduler to backend services (cron jobs)

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [x] 🤖 Build or CI

## Related Issue

- https://github.com/hotosm/fmtm/pull/1984
- https://github.com/hotosm/fmtm/issues/1603
- https://github.com/hotosm/fmtm/issues/1999
- https://github.com/hotosm/fmtm/issues/2181
- https://github.com/hotosm/fmtm/pull/2088

## Describe this PR

- Using cron in a container without root is actually really difficult. We need root to run the cron daemon.
- Instead I added supercronic, a go binary specifically built to replace cron in containers.
- This runs as part of the local development compose stack.
- As we don't build a specific container for this, it's much more flexible. We can replace the compose service with a Kubernetes CronJob as needed.
- The python scripts to run are under `/src/backend/scheduled/xxx.py`
- TODO:
  - [x] Update the code in the linked PRs so the logic happens entirely in xxx_crud.py modules. The xxx_routes.py code should simply call the logic. This way we can replicate this in the `/src/backend/scheduled/xxx.py` too.
    - Unlock tasks / features
    - Delete old user accounts
    - Trigger ODK photo upload?
  - [x] Remove any requirement on the `request: Request` var from logic, as we can't use this on a schedule (instead we need to use a service account OSM token to send emails).
  - [x] Test thoroughly.
  - [x] Add compose services to remaining compose files + deploy.

- I tried many ways of using cron. All pretty impossible without root, if we want to receive any kind of logging.
- I also considered an approach using a simple sleep command: https://gist.github.com/mowings/a32fbf4dd46d9fb9661822056ceda91c, but it's very difficult to work the syntax as part of the compose `command:` section.

## Review Guide

- For testing, replace the cron schedule in compose.yaml with `* * * * *` to run every minute.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -374,7 +374,7 @@ services:
         echo '* */3 * * * /opt/scheduler/unlock_tasks.py' > ./crontab
 
         # Task unlocking every Sunday 00:00
-        echo '0 0 * * 0 /opt/scheduler/inactive_users.py' > ./crontab
+        echo '0 0 * * 0 /opt/scheduler/inactive_users.py' >> ./crontab
 
         exec /usr/local/bin/supercronic ./crontab
       "
--- a/compose.yaml
+++ b/compose.yaml
@@ -77,7 +77,7 @@ services:
     volumes:
       - fmtm_logs:/opt/logs
       - ./src/backend/pyproject.toml:/opt/pyproject.toml:ro
-      - ./src/backend/app:/opt/app
+      - ./src/backend/app:/opt/app:ro
       - ./src/backend/tests:/opt/tests:ro
       - ./src/backend/stats:/opt/stats:ro
       # - ../osm-fieldwork/osm_fieldwork:/opt/python/lib/python3.12/site-packages/osm_fieldwork:ro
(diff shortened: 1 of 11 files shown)


spwoodcock:
I won't work any more on this for now - moving onto MdP work.

Everything is in place to continue the work for unlocking inactive users and tasks @Anuj-Gupta4 🙏 

Although I would say this is one of the lower priority items on our roadmap, to be addressed by anyone when there is some free time.

Anuj-Gupta4 (approved):
This looks great! I'll make a few minor adjustments to refine it for now and come back to it later.

compose.yaml
```diff
@@ -344,3 +344,37 @@ services:
...
+    networks:
+      - fmtm-net
+    entrypoint: ["/bin/sh", "-c"]
+    # The approach below allows us to easily switch to Kubernetes CronJob if needed
+    command: |
+      "
+        # Task unlocking every 3hrs
+        echo '* */3 * * * /opt/scheduler/unlock_tasks.py' > ./crontab
+
+        # Task unlocking every Sunday 00:00
```
spwoodcock, line 368: Oops, this should say 'Check inactive users every Sunday 00:00

src/backend/scheduler/inactive_users.py
```diff
@@ -0,0 +1,35 @@
...
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger(__name__)
+
+
+async def scheduled_process_inactive_users():
+    """Process inactive users."""
+    try:
+        async with await AsyncConnection.connect(settings.FMTM_DB_URL) as db:
+            log.info("Starting processing inactive users")
+            await process_inactive_users(db)
```
spwoodcock, line 28: This needs to be fixed

src/backend/scheduler/unlock_tasks.py
```diff
@@ -0,0 +1,36 @@
...
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger(__name__)
+
+
+async def scheduled_unlock_inactive_tasks():
+    """Process inactive tasks."""
+    try:
+        async with await AsyncConnection.connect(str(settings.FMTM_DB_URL)) as db:
+            log.info("Starting processing inactive tasks")
+            await trigger_unlock_tasks(db)
```
spwoodcock, line 29: This needs to be fixed

Anuj-Gupta4 (approved):

Issue #2157, opened 2025-02-06
Bug rendering PMTiles layer in OpenLayers for React frontend


manjitapandey:
**Describe the bug**
The base imagery is not rendered properly and shows multiple imageries when zoomed out while using custom drone imagery. 

**To Reproduce**

1. Go to 'any project and visualise the PM tile'
2. Zoom out the map to view multiple tiles
3.  Go to other projects and see the base imagery

**Expected behavior**
The base imagery should be visualized on its location only and shouldn't be duplicated. Also it should be rendered properly. 

**Screenshots**

![Image](https://github.com/user-attachments/assets/33a61a24-07e3-4040-b58b-2382d34d1634)

**Additional context**
If zoomed in on the actual location of tms tile, the drone imagery is not seen there and is shifted to different location. 
If you open different project too, same base imagery is visualized until you refresh the page.

spwoodcock:
Just two clarify, there are two issues right?

- Load PMTiles for a project --> change project --> PMTiles for previous project still visible

- As above, but the tiles are loaded in the wrong place on the map. Are they displayed in the correct place on the original project they were loaded for?

manjitapandey:
Yes, there are multiple issues:
1. Load PMTiles for a project --> change project --> PMTiles for previous project still visible
2. the tiles are loaded on multiple areas. Among them, one is its correct position, all other locations where tiles are seen are incorrect.

spwoodcock:
> Yes, there are multiple issues:
> 1. Load PMTiles for a project --> change project --> PMTiles for previous project still visible
> 2. the tiles are loaded on multiple areas. Among them, one is its correct position, all other locations where tiles are seen are incorrect. 

### Tile generation issue (2)

The second issue will be because of the swapped xy edit that was done on the Tokha server. 

As an alternative, we could swap out osm-fieldwork basemapper with go-tilepacks quite easily. Will give this some more thought.

@Sujanadh we could look at this tomorrow 👍

### Offline PMTiles issue

The first issue is a frontend bug. The PMTiles loading from OPFS should only load for the specified project id, if present in OPFS, else fallback to loading none.

@NSUWAL123 I know you have a lot of tasks, but it would be amazing if you get a chance to debug this one 🙏

NSUWAL123:
Yeah sure, Sam! I was also looking at the issue related to multiple-tile rendering. There were some issues that I faced: 
[...]

PR #2162, opened 2025-02-07
Fix TMS issues by using go-tilepacks and go-pmtiles

spwoodcock:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Work for #2157

## Describe this PR

- Fixes the TMS issues once and for all!
- `go-tilepacks` is much better than basemapper.py at generating mbtiles - so much faster and better at handling different formats.
- I also fixed an issue where the encoding is gzipped from servers: this was a problem for the Tokha TMS server (perhaps relay this info to devops, as it's useful to know! TMS server responses **can** be gzipped, but the receiving software needs to be able to decode the data). In some cases, it might be good to simply remove the gzip encoding from the responses.
- Also includes `go-pmtiles` to do the conversion to pmtile format.
- osm_fieldwork.basemapper.py is no longer used.

## Screenshots

image

## Additional Considerations

- I'm sure this is probably quite CPU and memory intensive - I haven't tested thoroughly.
- It might lock up the worker / thread until done. Luckily it's reasonably quick (<1min generally).
- Might need to be tested if we experience issues!
- In future this will be migrated to the https://github.com/hotosm/basemap-api microservice anyway.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -76,7 +76,6 @@ services:
     # tty: true
     volumes:
       - fmtm_logs:/opt/logs
-      - fmtm_tiles:/opt/tiles
       - ./src/backend/pyproject.toml:/opt/pyproject.toml:ro
       - ./src/backend/app:/opt/app
       - ./src/backend/tests:/opt/tests:ro
--- a/contrib/basemaps/Dockerfile
+++ b/contrib/basemaps/Dockerfile
@@ -1,0 +1,52 @@
+FROM docker.io/alpine/curl:8.11.1 AS base
+
+
+# Download and verify tilepack binary
+# FROM base AS tilepack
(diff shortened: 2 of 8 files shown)


Sujanadh (approved):

Sujanadh:
Great work :tada: I tested using all the tile sources, and performance is so great; it was quick.
The only issue is that the Bing source doesn't work, which you also have mentioned in the comment line. My device got frozen sometimes, maybe due to its CPU usage; not sure. Besides that, it works really well.  :+1:

PR #2178, opened 2025-02-11
And endpoints to download form XML and create new form submissions

### LINKED ISSUE #2166: Endpoint to submit ODK submission XML via FMTM API

- ODK Collect handles the ODK Central interaction typically: pulling the form / entities, submitting the final XML via OpenRosa.
- For web-forms, we have a prop to inject the form XML into the web-form page.
- There is a submit button on the form. In future it should POST directly to ODK Central using the appuser token, but for now, we decided to simply return the submission XML.
- The XML will be sent back to FMTM, for manual submission to ODK Central.

- [x] GET endpoint to return form XML, for injecting into Web Form.
- [x] POST endpoint to submit XML to ODK Central using REST API
  - Note that ODK Central has the option to submit via OpenRosa or REST - both should work the same in theory.
  - Double check the API docs to see if there are any caveats to using REST over OpenRosa.
    - I have a feeling that attachment submission alongside the XML will require a separate API call.

This will be changed in future:
- Web form will pull forms and Entity data from ODK Central (appuser token).
- Web form will submit data to ODK Central (appuser token).

**Additional context**

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #2166

## Describe this PR

Based on our web-form discussion @DanielJDufour 👍 

Two new endpoints:
- Download form XML for project (to insert into web forms)
  - GET `/projects/{project_id}/form-xml`
  - Docs: http://api.fmtm.localhost:7050/docs#/projects/get_project_form_xml_route_projects__project_id__form_xml_post
- Create new submission from XML (provided by webforms).
  - POST `/submission?project_id=${projectId}`
  - Docs: http://api.fmtm.localhost:7050/docs#/submission/create_submission_submission_post
  - This endpoint also requires file uploads, which are handled by the server and uploaded as attachments to Central.
- Also cleaned up a few unused functions.

> Once deployed to dev, you can view the docs by replacing http://api.fmtm.localhost:7050 --> https://api.dev.fmtm.hotosm.org

## Review Guide

- Probably best testing via Web Forms, as we need a valid submission XML.


src/backend/app/config.py
```diff
@@ -234,6 +234,20 @@ def assemble_db_connection(cls, v: Optional[str], info: ValidationInfo) -> Any:
...
+    def pyodk_env_vars(self) -> None:
+        """Set the pyodk environment variable configuration."""
+        if self.ODK_CENTRAL_URL:
+            os.environ["PYODK_BASE_URL"] = self.ODK_CENTRAL_URL
+        if self.ODK_CENTRAL_USER:
+            os.environ["PYODK_USERNAME"] = self.ODK_CENTRAL_USER
+        if self.ODK_CENTRAL_PASSWD:
+            os.environ["PYODK_PASSWORD"] = self.ODK_CENTRAL_PASSWD.get_secret_value()
+        # This file caches as session token, to avoid creating new sessions
+        os.environ["PYODK_CACHE_FILE"] = "/home/appuser/.pyodk_cache.toml"
```
DanielJDufour, line 248: I'm wondering if we want to use a hard_coded absolute path.  I think using an absolute path would require that we always run the backend in a docker container.  Are there any situations where someone would want to run the API without Docker?

Would it be possible to try to add an if statement to only set PYODK_CACHE_FILE if it's not already set.  Leaving it possible in the future to support someone over-writing the default?

src/backend/app/projects/project_routes.py
```diff
@@ -936,6 +936,18 @@ async def add_new_entity(
...
+@router.post("/{project_id}/form-xml")
+async def get_project_form_xml_route(
+    project_user: Annotated[ProjectUserDict, Depends(mapper)],
+) -> str:
+    """Get the raw XML from ODK Central for a project."""
+    project = project_user.get("project")
+    odk_creds = project.odk_credentials
+    odkid = project.odkid
+    odk_form_id = project.odk_form_id
+    return await central_crud.get_project_form_xml(odk_creds, odkid, odk_form_id)
```
DanielJDufour, line 948: is this await necessary?  I genuinely don't know as I haven't done much async programming in Python.  In JavaScript, you wouldn't need to add an await.  It doesn't hurt though.

src/backend/app/submissions/submission_crud.py
```diff
@@ -250,6 +194,35 @@ async def get_submission_detail(
...
+    """Create a new submission in ODK Central, using pyodk REST endpoint."""
+    submission_attachments = submission_attachments or {}  # Ensure always a dict
+    attachment_filepaths = []
+
+    # Write all uploaded data to temp files for upload
+    # (required by PyODK)
+    for file_name, file_data in submission_attachments.items():
+        temp_path = f"/tmp/{file_name}"
+        with open(temp_path, "wb") as temp_file:
+            temp_file.write(file_data.getvalue())
```
DanielJDufour, line 213: nice use of getvalue!

src/backend/app/submissions/submission_crud.py
```diff
@@ -250,6 +194,35 @@ async def get_submission_detail(
...
+    device_id: Optional[str] = None,
+    submission_attachments: Optional[dict[str, BytesIO]] = None,
+):
+    """Create a new submission in ODK Central, using pyodk REST endpoint."""
+    submission_attachments = submission_attachments or {}  # Ensure always a dict
+    attachment_filepaths = []
+
+    # Write all uploaded data to temp files for upload
+    # (required by PyODK)
+    for file_name, file_data in submission_attachments.items():
```
DanielJDufour, line 210: You can create a temporary directory in Python via [tempfile.TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory).  I'm wondering if that's better as I think it'll provide more flexibility in deployment in the future.

DanielJDufour (commented):
looks awesome!  I liked the async approach.  gave a first-pass review.  I haven't approved yet, since I haven't test the code yet.

src/backend/app/submissions/submission_routes.py
```diff
@@ -237,24 +270,7 @@ async def get_submission_count(
 #     return Response(content=osmoutfile_data, media_type="application/xml")
 
 
-@router.get("/submission_page")
-async def get_submission_page(
-    project_user: Annotated[ProjectUserDict, Depends(mapper)],
```
Sujanadh, line 242: Not sure why this endpoint is deleted?, since this endpoint is used for the infographics page to visualize the submissions for last 10 days or 30 days.

src/backend/app/submissions/submission_crud.py
spwoodcock: Good point, that would probably be better!

In a rush I forgot to add deletion code after the upload anyway 😅 so tempfile would handle that nicely

src/backend/app/projects/project_routes.py
spwoodcock: Yeah you wouldn't need to await, else it will try to serialise a coroutine as a response.

But saying this, it might be best to define the route function itself as a sync def, as the underlying osm-fieldwork code is not async 😅

I'll have a test!

src/backend/app/config.py
spwoodcock: Thanks for mentioning! The default cache path set by pyodk is actually fine, but I wrote this for documentation mostly. Will consider removing it though 👍 else set it correctly based on pathlib determined HOME

src/backend/app/submissions/submission_routes.py
spwoodcock: Oh no, really! I did a search and couldn't find reference in the frontend 😅 my bad, I'll check again!!

Sujanadh: I remember creating this endpoint for that specific purpose, if it is not in use for the infographics then it is totally fine, I guess the graphs are all generated from `statuses` api if I am not wrong; in that case, it is completely fine to remove it  :+1:

spwoodcock: I can't see this use anywhere - I think it might have been replaced?

When we load the infographics we use:
- projects/3?project_id=3
- /projects/3/entities/statuses
- /projects/contributors/3
- /tasks/activity

I think it was stopped being used when we started replacing infographic-specific endpoints with existing data we have via Entity statuses etc

src/backend/app/projects/project_routes.py
spwoodcock: This takes about 0.25s to run on average, so no a huge deal if blocking the event loop for that long.

But to mitigate this problem entirely, I wrapped in `run_in_threadpool`, meaning this will run the task in a new thread in the same process (keeping the main thread responsive to further requests):

```python
    return await run_in_threadpool(
        central_crud.get_project_form_xml,
        odk_creds, odkid, odk_form_id,
    )
```

It doesn't add much overhead to execution time, maybe ~0.05s at most

Issue #2181, opened 2025-02-12
Allow super admin or project creator to unlock task or revert to ready for mapping


manjitapandey:
**Is your feature request related to a problem? Please describe.**
In field mapping exercise there is maximum chances that user can lock task or mark task as mapped without mapping completely or leave the task mapping midway. In such cases, there is no workflow to undo these actions causing users not being able to map new features or provide false information to management team. 

**Describe the solution you'd like**
Provide the access to super admin or project creator to unlock the tasks locked by mapper or reset the mapped task to ready for mapping state. 

**Describe alternatives you've considered**
Alternatively, for now we can add map new feature button on mapper frontend irrespective of task but this will also cause problem in validation workflow later and its good that features and submissions are linked to task ID. 

**Additional context**

spwoodcock:
We have the code in place to already do this, in a PR that unlocks based on a cron.

The simple thing to do would be:
- Endpoint with project admin permissions
- Use the same logic as the cron for unlocking tasks (DRY code)

spwoodcock:
Actually, I realised this has a frontend part I think:

- Currently, if a user has locked a task, another user can't unlock it. This is a restriction imposed by the frontend.
- If the user if a project manager or greater, we want to allow them to unlock the task by calling the API.
- Related to https://github.com/hotosm/fmtm/pull/2207, where we want to use `const isProjectManager = useIsProjectManager();`, the we we have conditional logic to allow the user to trigger an unlock of the task.

Issue #1161, opened 2024-02-02
User Roles: Create user roles in frontend for organization admin


manjitapandey:
**Is your feature request related to a problem? Please describe.**
The roles of organisational admin should be created on the frontend.
The credentials are provided for those roles by backend as per now.
The tasks may be subdivided as below:

- [x] - On logging on as an organisational admin, in project creation, One should see the prefilled organisation name (non editable in case of single). In case user is org admin of multiple orgs, they should be only provided options which they have org admin access.  
- [x] - The New button on manage organisation tab should be removed. 
- [x] - Addition of assign project manager field on create project step 1(Ref: design). 

Additional context

The permissions are specified by the file below:
[https://docs.google.com/spreadsheets/d/1crWztvXvuewE8Zv-qxbv0t_-eYHUqSlrJQgeto8kmf4/edit#gid=1497569844](https://github.com/hotosm/fmtm/issues/url)

related #1068

spwoodcock:
Related to slack discussion:

In general super admin has access to all the below:
- Org creation - super admin only - `isSuperAdmin`
- Org editing - specific org admin - `isOrgAdmin`
- Project creation button - any org admin - `anyOrgAdmin`
- Project editing - specific project admin, or org admin - `isProjectAdmin`
- Project user management - specific field admin, project admin, or org admin - `isFieldAdmin`

Does that seem about right @NSUWAL123 ?

Is there any missed frontend role-specific functionality @manjitapandey ?

Open to ideas / changes - do what you think is most appropriate 👍

manjitapandey:
Blocked by https://github.com/hotosm/fmtm/issues/2221  since i don't have an account of mapper role.

Issue #1068, opened 2023-12-27
Hide/unhide the buttons from frontend based on user roles


manjitapandey:
**Is your feature request related to a problem? Please describe.**
for admin roles most of the features are visible however, the organisational admin shouldn't have permission to create projects using other organisation name. Such restrictions on frontend are different at different level.

**Describe the solution you'd like**

- [x] the organisation oadmin shouldn't be able to choose organisation name other than its own organisation during project creation.
- [x] when creating a project, if the user is only part of one organisation, the value should be prefilled on the project creation form.
- [x] organisation admin should be able to edit their own organisation details.
- [x] allow super admin and  organisational admin to assign project manager.

PR #2207, opened 2025-02-18
Frontend user roles integration

NSUWAL123:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #1068
- Related to #1161
- Related to #2181

## Describe this PR
- Hide/Unhide buttons, provide access to functionality and pages based on user roles

## Screenshots
N/A


src/frontend/src/hooks/usePermissions.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// PROJECT-LEVEL
+export function useCreateProjectAccess() {
+  const authDetails = CoreModules.useAppSelector((state) => state.login.authDetails);
+  return authDetails?.role === user_roles.ADMIN || !!authDetails?.orgs_managed;
+}
+
+// ORGANIZATION-LEVEL
+export function useEditOrganizationAccess(id: number) {
```
spwoodcock, line 17: Instead of defining these based on the action type, could they be defined based on the access level?

We shouldn't need many in that scenario:

```js
export function currentUserIsOrgManager(id: number) {
  const authDetails = CoreModules.useAppSelector((state) => state.login.authDetails);
  return (
    authDetails?.role === user_roles.ADMIN || (authDetails?.orgs_managed && authDetails?.orgs_managed?.includes(id))
  );
}
```

currentUserIsProjectManager

currentUserIsAdmin

src/frontend/src/hooks/usePermissions.ts
```diff
@@ -0,0 +1,22 @@
...
+// ADMIN-ONLY
+export function useAdminAccess() {
+  const authDetails = CoreModules.useAppSelector((state) => state.login.authDetails);
+  return !!authDetails && authDetails?.role === user_roles.ADMIN;
+}
+
+// PROJECT-LEVEL
+export function useCreateProjectAccess() {
+  const authDetails = CoreModules.useAppSelector((state) => state.login.authDetails);
+  return authDetails?.role === user_roles.ADMIN || !!authDetails?.orgs_managed;
```
spwoodcock, line 13: This is good for `orgs_managed: null` 👍 

If for some reason the backend response changed to be `orgs_managed: []` this would evaluate true, but the next check for the specific org id should still prevent access.

spwoodcock (approved):
Looks good to me!

Useful reference: https://docs.google.com/spreadsheets/d/1crWztvXvuewE8Zv-qxbv0t_-eYHUqSlrJQgeto8kmf4/edit?gid=1497569844#gid=1497569844

PR #2212, opened 2025-02-19
Refactor: Replace old button with new button components

NSUWAL123:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
Previously, a new button component was made which needed to be replaced with the old button component

## Describe this PR
- Replace of previously used old button components with new button components overall

## Screenshots
image


spwoodcock (approved):
Love this idea overall 🎉 

But was wondering if in the long run we can replace components like buttons designed for FMTM with generic button components from shoelace-style?

Similar to the components we have integrated into the mapper frontend.

React 19 has first party support for web-components, so it should work well

Issue #2190, opened 2025-02-13
Generate project data endpoint is failing during project creation


manjitapandey:
**Describe the bug**
As a part of Ghana KYC project, while creating a project there was 500 error in upload project data endpoint during project creation. I was using small AOI with high building density extracted from osm. 

**To Reproduce**

1. Go to 'create project'
2. Upload AOI provided below
3. Fetch map features from osm

**Expected behavior**
The project should be created successfully

**Screenshots**

![Image](https://github.com/user-attachments/assets/28188279-60e1-4c6c-96b1-1beea90539eb)

**Additional context**

spwoodcock:
Hmm its a 404, not a 500

manjitapandey:
Oops my bad, edited and updated. 
My assumption is timeout since its 404 error after just 1 minutes

spwoodcock:
Could you upload the AOI for debugging @manjitapandey?

Ideally we get a fix out for this in the next release in the coming days 👍

Sujanadh:
I am working on this, splitting geojson with task areas is taking longer causing time out, additionally fetching fgb data and converting it into geojson is also taking more time. This is the AOI.
```json
{
"type": "FeatureCollection",
"name": "Old Fadama Community_Accra_AoI",
"crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:OGC:1.3:CRS84" } },
"features": [
[...]

PR #2214, opened 2025-02-19
Optimize generating project data

Sujanadh:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [X] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue: #2190

## Describe this PR

- Optimized fetching fgb data from s3 using aiohttp
- Refactored sql that converts fgb to geojson using materialized view
- Simplified sql that splits geojson by task areas using batch insertion
- Simplified adding required properties to geojson

## Screenshots

N/A

## Review Guide


spwoodcock (changes requested):
Some nice updates!

I think there could be some logic issues that I commented on, but could be wrong 😄 

Let me know your thoughts

src/backend/app/db/postgis_utils.py
```diff
@@ -220,130 +222,86 @@ async def split_geojson_by_task_areas(
...
-                DROP TABLE IF EXISTS temp_features CASCADE;
-            """)
+        features = featcol["features"]
+        result_dict = defaultdict(list)
 
+        async with db.cursor() as cur:
             await cur.execute("""
-                -- Create a temporary table to store the parsed GeoJSON features
-                CREATE TEMP TABLE temp_features (
+                DROP TABLE IF EXISTS tmp_project_features;
```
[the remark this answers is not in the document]
Sujanadh: Yes, it does support running two sql blocks separated by a semicolon ';' if no parameters %s are used in the sql. So, you are right in the case where we need to pass parameters; executing those multiple sql won't be supported. We can use DO statements if we want to run multiple sql with parameters.

src/backend/app/db/postgis_utils.py
```diff
@@ -220,130 +222,86 @@ async def split_geojson_by_task_areas(
...
+            data = [
+                (
+                    str(f["properties"].get("osm_id")),
+                    json.dumps(f["properties"]),
+                    json.dumps(f["geometry"]),
+                )
+                for f in batch
+            ]
+
+            async with db.cursor() as cur:
```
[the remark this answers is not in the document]
Sujanadh: yeah you are right i can change that to use one cursor for all the batches.

Sujanadh:
It is still taking time. Can you have a look into this, @spwoodcock? :pray:

spwoodcock:
> It is still taking time. Can you have a look into this, @spwoodcock? :pray: 

So sorry Sujan, I probably can't for quite a while now. I have needed to switch to web forms development for weeks now & its getting urgent.

What I can say, is that while these changes are great and we should look at optimisations for sure, I don't think this is high priority by any means.

Instead, I think an acceptable approach is to just have the user wait some time if they want to make a huge project. The frontend should have changes in place to prompt the user when this is the case - if that isn't working it can be fixed, but I think its the way to go for now

Sujanadh:
This was pushed for ghana project. since it was blocker for them. we can merge and test it. I will get back to it later again after stablising tokha requirement.

src/backend/app/db/postgis_utils.py
```diff
@@ -220,130 +222,83 @@ async def split_geojson_by_task_areas(
...
+            # Batch insert features
+            for i in range(0, len(features), batch_size):
+                batch = features[i : i + batch_size]
+                data = [
+                    (
+                        str(f["properties"].get("osm_id")),
+                        json.dumps(f["properties"]),
+                        json.dumps(f["geometry"]),
+                    )
+                    for f in batch
```
spwoodcock, line 250: Same question as for the other batch operations: what if one fails?

I think this case is easier though, as its inside a single db transaction. Should be easy to rollback the changes on failure.

The question is, is the performance improvement significant to warrant the increased complexity?

I'm not even sure how this improves performance, as the db is local to the API anyway. We aren't doing anything concurrently with async here (and we shouldn't to avoid db issues) - it's simply iterating through batches and inserting sequentially. Theoretically it should be slower as we need to execute multiple commands 🤔

spwoodcock (changes requested):
Sorry, there are merge conflicts since the last change!

Sujanadh:
## Updates:
- prepared arrays as parameters (`feature_ids`, `feature_properties`, `feature_geometries`) as they are more efficient than  parsing a large `JSON` string in `SQL`
-  batch operation with `unnest()` which is typically faster than iterating through `JSON` elements
- used single query execution
- avoid creating and dropping temporary tables
- used `ST_Interescts` rather than `ST_Within(ST_Centroid())` where it  first checks if geometries actually overlap and again checks if the centroid is inside the task outline

src/backend/app/db/postgis_utils.py
```diff
@@ -218,132 +220,91 @@ async def split_geojson_by_task_areas(
...
+                                f.properties,
+                                '{task_id}',
+                                to_jsonb(t.project_task_index)
+                            ),
+                            '{project_id}', to_jsonb(%s)
+                        )
+                    ) AS feature
+                FROM tasks t
+                JOIN feature_data f
+                ON ST_Intersects(f.geom, t.outline)
```
spwoodcock, line 263: If we do an intersect, couldn't a geometry intersect two tasks and be included in both?

With the centroid approach its only possible to be included in one task.

Am I missing something?

spwoodcock (commented):
Nice changes 👍

Just want to clarify the st_intersects change

src/backend/app/db/postgis_utils.py
Sujanadh: I tested it I didn't see any feature present on boundary being shared with two tasks which was avoided by distinct selection; however, there could be a scenario when the centroid is on one task and it is assigned to another task. I just didn't want to calculate centroid and check again using st contains. I think I do not have any other option to just use the single post gis function rather than computing centroid and using st contains.

Sujanadh: I just updated back to ST_within to check if centroid lies on task outline or not :+1:

spwoodcock: Do you know what the potential performance benefit is? PostGIS is pretty optimised C code, so I can't imagine there is much in it.

When it comes to performance, I would say its never good to favour it over data integrity. We want to guarantee the centroid for a geom is present in the task area, as this is also how we filter geoms on the frontend: an inconsistency between the two could cause an error

spwoodcock: > I just updated back to ST_within to check if centroid lies on task outline or not :+1: 

Just read this, OK good with me!

Sujanadh: i agree on that :+1:

src/backend/app/db/postgis_utils.py
```diff
@@ -353,43 +314,31 @@ def add_required_geojson_properties(
...
-                # id is nested in properties, use that
-                feature["id"] = f"{prop_id}"
-                properties["osm_id"] = prop_id
-            elif fid := properties.get("fid"):
-                # The default from QGIS
-                feature["id"] = f"{fid}"
-                properties["osm_id"] = fid
-            else:
-                # Random id
-                # NOTE 32-bit int is max supported by standard postgres Integer
```
spwoodcock, line 378: Always good to keep notes when refactoring if possible 🙏

This sort of note could be useful to future devs

spwoodcock:
Sorry, I'm reviewing this in more details than usual as its such a critical part of the workflow 😄

I will finalise the check, test, and merge when at my desk (waking up currently 😁)

Sujanadh:
> Sorry, I'm reviewing this in more details than usual as its such a critical part of the workflow 😄
> 
> I will finalise the check, test, and merge when at my desk (waking up currently 😁)

I like every detail review you provide :+1:

spwoodcock (approved):
Nice work here 😄

Issue #35, opened 2023-01-15
Option for campaign manager to allocate task to specific user


ivangayton:
In some cases, Carter at the WNA Hub wants to assign tasks to his mappers, rather than allowing them to freely select a task. He can already do so by downloading the QR codes himself and sending them to the relevant team members, but then the tasks will be locked by him, rather than the users, the users will not be able to mark the tasks as complete.

It would be great if a campaign manager had an option to see a list of registered users and allocate a task to each. The system should then send a push notification of some kind to those users.

More detail as we discuss with Carter and his team.

spwoodcock:
This will be addressed soon!

We have an ASSIGN event on the backend.

The events API can be called with the ASSIGN event, then:
- Task is set to locked, by the specified user.
- User is notified via OSM message & email that a task was assigned to them.
- There should be a filter on the mapper frontend for 'my tasks', so the user can see they are actively mapping in the assigned task area(s).

manjitapandey:
This is the UI we have come up with: 
Please drop your suggestions and thoughts regarding the UI. 

![Image](https://github.com/user-attachments/assets/fc782425-e2df-4ea7-8a6a-3900d1ebff88)

![Image](https://github.com/user-attachments/assets/b7c5f285-49ef-4d28-8c7c-a04f0f5bac02)

spwoodcock:
This looks pretty nice, but I have a feeling its more intuitive to go from the map first.

Click on a task --> assign to user --> enter username.

Once we remove some of the mapping capability from the management frontend, there will be room in the task popup for a button.

It it would reduce the need for extra pages / components to be developed and maintained 😃

manjitapandey:
That's also nice way to assigning task but would be difficult if we would be bullk assigning tasks to users. I will brainstrom it more and comeup with UI and discuss further on this.

spwoodcock:
> That's also nice way to assigning task but would be difficult if we would be bullk assigning tasks to users. I will brainstrom it more and comeup with UI and discuss further on this. 

True!

Perhaps we go via the map for individual task assignments, but then have a project management tab called 'Bulk Assign'? (With the design you have made)

PR #2237, opened 2025-02-28
Assign users to tasks (+ project team integration)

Anuj-Gupta4:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related to: #35 

### DIFF
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -76,7 +76,6 @@ if TYPE_CHECKING:
         BasemapUpdate,
         GeometryLogIn,
         ProjectIn,
-        ProjectTeam,
         ProjectTeamIn,
         ProjectUpdate,
     )
@@ -908,10 +907,10 @@ class DbProjectTeam(BaseModel):
     project_id: int
 
     # Computed
-    users: Optional[list[dict]] = None
+    users: Optional[list[dict]] = []
 
     @classmethod
-    async def one(cls, db: Connection, team_id: UUID) -> "ProjectTeam":
+    async def one(cls, db: Connection, team_id: UUID) -> Self:
         """Fetch a single team by its ID along with user details."""
         async with db.cursor(row_factory=class_row(cls)) as cur:
             await cur.execute(
@@ -944,7 +943,7 @@ class DbProjectTeam(BaseModel):
         return team
 
     @classmethod
-    async def create(cls, db: Connection, team_in: "ProjectTeamIn") -> "ProjectTeam":
+    async def create(cls, db: Connection, team_in: "ProjectTeamIn") -> Self:
         """Create a new team for a project."""
         model_dump = dump_and_check_model(team_in)
         columns = ", ".join(model_dump.keys())
@@ -966,7 +965,7 @@ class DbProjectTeam(BaseModel):
     @classmethod
     async def update(
         cls, db: Connection, team_id: UUID, team_update: "ProjectTeamIn"
-    ) -> "ProjectTeam":
+    ) -> Self:
         """Update a team for a project."""
         model_dump = dump_and_check_model(team_update)
         placeholders = ", ".join(f"{key} = %({key})s" for key in model_dump.keys())
@@ -1001,7 +1000,7 @@ class DbProjectTeam(BaseModel):
             )
 
     @classmethod
-    async def all(cls, db: Connection, project_id: int) -> list["ProjectTeam"]:
+    async def all(cls, db: Connection, project_id: int) -> list[Self]:
         """Fetch all teams for a project along with users."""
         async with db.cursor(row_factory=class_row(cls)) as cur:
             await cur.execute(
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -25,7 +25,7 @@ import json
 from datetime import timedelta
 from io import BytesIO
 from re import sub
-from typing import TYPE_CHECKING, Annotated, Optional, Self
+from typing import TYPE_CHECKING, Annotated, List, Optional, Self
 from uuid import UUID
 
 import geojson
@@ -862,7 +862,7 @@ class DbTaskEvent(BaseModel):
                         u.profile_img
                     FROM inserted
                     JOIN users u ON u.id = inserted.user_id;
-            """,
+                """,
                 model_dump,
             )
             new_task_event = await cur.fetchone()
@@ -875,6 +875,51 @@ class DbTaskEvent(BaseModel):
             )
 
         return new_task_event
+
+
+class DbTaskAssignment(BaseModel):
+    """Table task_assignments."""
+
+    project_id: int
+    task_id: int
+    user_id: Annotated[Optional[int], Field(gt=0)] = None
+    assigned_at: Optional[AwareDatetime] = None
+
+    @classmethod
+    async def create(
+        cls, db: Connection, project_id: int, task_id: int, user_ids: List[int]
+    ):
+        """Create new task assignments for multiple users."""
+        print("user_ids in task assignments", user_ids)
+        async with db.cursor(row_factory=class_row(cls)) as cur:
+            await cur.executemany(
+                """
+                    INSERT INTO public.task_assignments (project_id, task_id, user_id)
+                    VALUES (%(project_id)s, %(task_id)s, %(user_id)s)
+                    ON CONFLICT (task_id, user_id) DO NOTHING;
+                """,
+                [
+                    {"project_id": project_id, "task_id": task_id, "user_id": user_id}
+                    for user_id in user_ids
+                ],
+            )
+
+    @classmethod
+    async def get(cls, db: Connection, project_id: int, task_id: int):
+        """Get task assignment by task ID."""
+        async with db.cursor(row_factory=class_row(cls)) as cur:
+            await cur.execute(
+                """
+                SELECT * FROM public.task_assignments
+                WHERE project_id = %(project_id)s
+                AND task_id = %(task_id)s
+                """,
+                {
+                    "task_id": task_id,
+                    "project_id": project_id,
+                },
+            )
+            return await cur.fetchall()
 
 
 class DbTask(BaseModel):
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -915,7 +915,6 @@ class DbProjectTeam(BaseModel):
                 {"team_id": team_id},
             )
             team = await cur.fetchone()
-            print("team", team)
 
         if team is None:
             raise KeyError(f"Team ({team_id}) not found.")
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -890,6 +890,7 @@ class DbProjectTeam(BaseModel):
         """Fetch a single team by its ID along with user details."""
         # NOTE Avoiding cyclical import errors
         from app.projects.project_schemas import ProjectTeam
+
         async with db.cursor(row_factory=class_row(ProjectTeam)) as cur:
             await cur.execute(
                 """
@@ -983,6 +984,7 @@ class DbProjectTeam(BaseModel):
         """Fetch all teams for a project along with users."""
         # NOTE Avoiding cyclical import errors
         from app.projects.project_schemas import ProjectTeam
+
         async with db.cursor(row_factory=class_row(ProjectTeam)) as cur:
             await cur.execute(
                 """
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -76,6 +76,8 @@ if TYPE_CHECKING:
         BasemapUpdate,
         GeometryLogIn,
         ProjectIn,
+        ProjectTeam,
+        ProjectTeamIn,
         ProjectUpdate,
     )
     from app.tasks.task_schemas import TaskEventIn
@@ -126,20 +128,10 @@ class DbUserRole(BaseModel):
                     (user_id, project_id, role)
                 VALUES
                     (%(user_id)s, %(project_id)s, %(role)s)
-                RETURNING *;
+                ON CONFLICT (user_id, project_id) DO NOTHING;
             """,
                 params,
             )
-            new_role = await cur.fetchone()
-
-        if new_role is None:
-            msg = f"Unknown SQL error for data: {params}"
-            log.error(f"Failed user role creation: {params}")
-            raise HTTPException(
-                status_code=HTTPStatus.INTERNAL_SERVER_ERROR, detail=msg
-            )
-
-        return new_role
 
     @classmethod
     async def all(
@@ -756,6 +748,7 @@ class DbTaskEvent(BaseModel):
 
     project_id: Annotated[Optional[int], Field(gt=0)] = None
     user_id: Annotated[Optional[int], Field(gt=0)] = None
+    team_id: Optional[UUID] = None
     username: Optional[str] = None
     comment: Optional[str] = None
     created_at: Optional[AwareDatetime] = None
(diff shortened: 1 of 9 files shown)


src/backend/migrations/009-task-assignments.sql
```diff
@@ -0,0 +1,44 @@
...
+
+-- Create task_assignments table only if it does not exist
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = 'task_assignments') THEN
+        CREATE TABLE public.task_assignments (
+            project_id INTEGER NOT NULL,
+            task_id INTEGER NOT NULL,
+            user_id INTEGER NOT NULL,
+            assigned_at TIMESTAMPTZ NOT NULL DEFAULT now()
```
spwoodcock, line 12: Do we need a new table for this?

I thought we would just have an ASSIGN event in the task events table?

spwoodcock (changes requested):

src/backend/migrations/009-task-assignments.sql
spwoodcock: Adding new things to the database should always be a last resort

Anuj-Gupta4: If we rely solely on the task_events table, we'd either have multiple event records per user (causing redundancy and data bloat) or a separate field for assignees that stores a list of user IDs resulting in slower querying and an empty column for non-assignment events.

spwoodcock: Good point - we need to reference both the assigner (eg manager) and the assigned (mapper) in one event.

Instead, how about we add a field to the tasks table `assigned_to`, then also make a task events entry.

The task event will have the user id for the assigner, while the assigned_to field will have the user id of the assigned.

Do you think that could be good?

(the task unlocking logic may have to be tweaked though to remove the `assigned_to` value from the tasks table though)

Anuj-Gupta4: Since we are to assign multiple users to a task, the field we add to the tasks table would need to be a list of integers (since the table can only have one row for each task ID). We would need to append or extend to the assigned_to field, which would be a list of user IDs. This is not the best approach.

It would be better to store it in a new task assignments table for faster querying, simple insert and delete for assignments and referential integrity with foreign keys. The assignments would also be automatically deleted without any new query code if the user, project or task were to be deleted.

spwoodcock: Fully agree - if we need to assign multiple users then this is the way to go 👍

Assigning multiple users doesn't really fit into our model though. A task can only be locked by one user. The credit for mapping a task will only go to one user. I thought we always discussed team mapping as an informal process, not recorded. 

Anyway, if this was discussed in my absence and agreed on, I will defer to your judgement 😄

src/backend/migrations/009-task-assignments.sql
```diff
@@ -0,0 +1,44 @@
+-- Begin transaction
+BEGIN;
+
+-- Create task_assignments table only if it does not exist
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = 'task_assignments') THEN
```
spwoodcock, line 7: There is a simpler syntax in postgres for this: `create table if not exists task_assignments`

Checking the information schema is only needed when this is not possible, such as with enums

src/backend/migrations/init/fmtm_base_schema.sql
```diff
@@ -467,6 +475,9 @@ ADD CONSTRAINT xlsforms_title_key UNIQUE (title);
 ALTER TABLE ONLY public.geometrylog
 ADD CONSTRAINT geometrylog_pkey PRIMARY KEY (id);
 
+ALTER TABLE ONLY public.task_assignments
+ADD CONSTRAINT task_assignments_pkey PRIMARY KEY (task_id, user_id);
+
 -- Indexing
 
 CREATE INDEX idx_projects_outline ON public.projects USING gist (outline);
```
spwoodcock, line 483: We probably need an index for faster lookup. Choosing the column depends what you will query on most frequently.

spwoodcock (approved):
Left some minor comments 👍

I'm still in two minds about this. Task assignment seems like something ephemeral to me - an event that occurs, but something that can be easily overwritten or chaged and not data that needs to be recorded in a separate table.

The requirement for multiple users to be assigned at once is a new one for me though - we haven't discussed this before - perhaps I missed it. With that in mind as the key criteria this could make sense (although the implication here is that perhaps we need to redesign how task locking and completion works).

So in summary, I kinda disagree with this approach, but don't want to hold up the process if this is needed 😅 Go ahead and merge when you think this is ready!

spwoodcock:
Oh by the way, I remembered what my original idea here was from months ago, as part of the event driven redesign (for reference).

1. Manager assigns task to user
2. Two events occur:
  - A MAP event is added for the assigned user on the task (I.e. it looks like they locked the task. If they visit FMTM they see a task locked by them - we can easily add a filter 'my tasks' on the frontend).
  - Notification / email is sent to user about the assignment.

spwoodcock:
> Will there be a scenario where multiple users will be assigned to a single task? To whom do we assign tasks, mappers? Are assigning tasks and locking tasks the same? I think I am a bit out of context here. @spwoodcock  @Anuj-Gupta4 

- I'm not sure if assigning a task to multiple people was a requirement from Tokha?

- Tasks are assigned to mappers 👍

- We made a new event ASSIGN to handle assignment in another PR. So I guess the idea was to have an assign event with the user id of the manager, then a MAP (lock) event with the user id of the mapper

spwoodcock:
Perhaps I made a mistake here when suggesting the ASSIGN event should automatically set the task state to LOCKED_FOR_MAPPING? As the assign event is by the manager, but we want the task locking to be done by the mapper.

Should we possibly revert this and instead create two event entries as mentioned above?

Sujanadh:
I am not sure there will be only one user/mapper assigned to a task. Some projects might need numbers of enumerators to do field surveys even for a single task (if the number of features within the task is huge), like a group of people assigned to a single task. In the case of Tokha, there are 4,5 people in a single task to do a survey. 

So, the concept of assigning tasks to users is to restrict users to not to select task that is not assigned to them?

spwoodcock:
> I am not sure there will be only one user/mapper assigned to a task. Some projects might need numbers of enumerators to do field surveys even for a single task (if the number of features within the task is huge), like a group of people assigned to a single task. In the case of Tokha, there are 4,5 people in a single task to do a survey. 
> 
> So, the concept of assigning tasks to users is to restrict users to not to select task that is not assigned to them?

Yep basically! Task assignment is pretty arbitrary - it only serves two purposes:
- Lock the task so others don't try to map there 
- Notify the user that their manager wants them to map the task

Sujanadh:
yeah we could do that. what if it restricts users who needs to do mapping on same task?

spwoodcock:
> yeah we could do that. what if it restricts users who needs to do mapping on same task?

I think we removed the restrictions in place previously - any user should be able to map regardless of who locked a task.

Task locking should only be an indication / helper and not restrictive

Sujanadh:
Okay, so we are just assigning users to let them know that it is assigned for you, just to notify?

Anuj-Gupta4:
How about we use the task event table to assign only one user and lock the task but keep the task assignments table to keep track of all the mappers that are expected to map the task and were notified about it? There could be a small section in the frontend to show that (maybe only for the project manager) if needed. I think it's better to not implement any restrictions on the tasks as well.

spwoodcock:
Yeah that could work 👍

But what about if the task gets re-assigned to new people? How do we handle that?

Anuj-Gupta4:
I would say let's implement bulk assign and bulk remove for a single task. In the frontend, there would be a list of users assigned to a task, and we would allow the project manager to select multi users and either add or remove them. There would be OSM message and email for both assign and remove as well. What do you think?

spwoodcock:
We discussed on our team call today - decided to implement teams to keep a 1:1 assignment for tasks.

I suggest we keep this simple and only allow 'mapper' teams. Plus Sujan suggested we make them project specific, instead of organisation specific.

table structure:

```sql
CREATE TABLE IF NOT EXISTS project_teams (
    team_id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
    team_name VARCHAR,
    project_id INTEGER NOT NULL,
    user_id INTEGER NOT NULL
);
```

Any thoughts on if we need extra fields?

Anuj-Gupta4:
Won't this structure result in each team having only one user? Also this table doesn't seem to be connected to tasks at all?

spwoodcock:
You are right - we actually need two tables 👍 
It doesn't need to be linked to tasks - that's done in the task_events table 😄

src/backend/app/db/models.py
```diff
@@ -126,20 +128,10 @@ async def create(
...
-            new_role = await cur.fetchone()
-
-        if new_role is None:
-            msg = f"Unknown SQL error for data: {params}"
-            log.error(f"Failed user role creation: {params}")
-            raise HTTPException(
-                status_code=HTTPStatus.INTERNAL_SERVER_ERROR, detail=msg
-            )
-
-        return new_role
```
spwoodcock, line 142: As this is removed, the methods no longer returns an instance of Self.

The ON CONFLICT part is nice to avoid errors, but why not also return the value?

All other methods in the models return the data as a pydantic model, it seems strange to break that convention for a single method.

(But maybe there is a good reason to do so??

src/backend/app/db/models.py
```diff
@@ -837,6 +830,11 @@ async def create(
         model_dump = dump_and_check_model(event_in)
         columns = ", ".join(model_dump.keys())
         value_placeholders = ", ".join(f"%({key})s" for key in model_dump.keys())
+        username = (
+            "NULL"
+            if model_dump.get("user_id") is None
+            else "(SELECT username FROM users WHERE id = %(user_id)s)"
```
spwoodcock, line 836: Nice, thanks for this 👍

I'm assuming the problem was that if a user is assigned that hasn't signed up yet, then we get an error

src/backend/app/db/models.py
```diff
@@ -877,6 +878,177 @@ async def create(
...
+
+    team_id: UUID
+    team_name: Optional[str] = None
+    project_id: int
+
+    @classmethod
+    async def one(cls, db: Connection, team_id: UUID) -> "ProjectTeam":
+        """Fetch a single team by its ID along with user details."""
+        # NOTE Avoiding cyclical import errors
+        from app.projects.project_schemas import ProjectTeam
```
spwoodcock, line 892: All imports should generally be at the top unless absolutely necessary.

In this file you will see a bunch if imports under `if TYPE_CHECKING`. Does adding it there fix the issue?

src/backend/app/db/models.py
```diff
@@ -877,6 +878,177 @@ async def create(
...
+
+        async with db.cursor(row_factory=class_row(ProjectTeam)) as cur:
```
spwoodcock, line 894: Related to the cyclical import, shouldn't this be returning Self? Don't all other classrow queries in this file do that?

src/backend/app/db/models.py
```diff
@@ -877,6 +878,177 @@ async def create(
...
+                    ON pt.team_id = ptu.team_id
+                LEFT JOIN users u
+                    ON ptu.user_id = u.id
+                WHERE pt.team_id = %(team_id)s
+                GROUP BY pt.team_id;
+                """,
+                {"team_id": team_id},
+            )
+            team = await cur.fetchone()
+            print("team", team)
```
spwoodcock, line 918: Missed print statement I think

spwoodcock (changes requested):
Looking good so far - nice work! 😁

src/backend/app/db/models.py
Sujanadh: I think he added it here to avoid a cyclic import issue. so that it can be used as a class row.

src/backend/app/db/models.py
Anuj-Gupta4: ok will remove

src/backend/app/db/models.py
Anuj-Gupta4: It was causing problems when team id was assigned instead of user id in task event for assign.

src/backend/app/db/models.py
Anuj-Gupta4: `ProjectTeam` contains an additional field to also return users. I originally had it defined in `DbProjectTeam` class itself. But then, we would need another pydantic class without users for other methods and doing that resulted in import errors as well.

src/backend/app/db/models.py
Anuj-Gupta4: I already have it imported for type checking. But `TYPE_CHECKING` didn't seem to work for `row_factory=class_row(ProjectTeam)`. Adding the import here looked like the cleanest fix.

src/backend/app/db/models.py
Anuj-Gupta4: The on conflict portion was added to avoid role creation errors when adding team users that were already assigned to the project with a role higher than mapper. The returned role is not being used anywhere. Maybe I should add that back. Perhaps we should add some logic to upgrade user role on new role creation if the new role is higher than the existing role?

spwoodcock: Yeah good idea! Its easy to update only the role field ON CONFLICT 😃

src/backend/app/db/models.py
```diff
@@ -126,20 +128,32 @@ async def create(
                     (user_id, project_id, role)
                 VALUES
                     (%(user_id)s, %(project_id)s, %(role)s)
-                RETURNING *;
+                    ON CONFLICT (user_id, project_id) DO UPDATE
+                    SET role = EXCLUDED.role
+                    WHERE user_roles.role < EXCLUDED.role;
```
spwoodcock, line 133: 👌

src/backend/app/db/models.py
```diff
@@ -877,6 +900,176 @@ async def create(
...
+
+    team_id: UUID
+    team_name: Optional[str] = None
+    project_id: int
+
+    @classmethod
+    async def one(cls, db: Connection, team_id: UUID) -> "ProjectTeam":
+        """Fetch a single team by its ID along with user details."""
+        # NOTE Avoiding cyclical import errors
+        from app.projects.project_schemas import ProjectTeam
```
spwoodcock, line 914: Sorry to keep on the same topic 😅

But I really think that shouldn't be importing schemas into the models.

The models define all possible fields as optional, so we use `cls` as the class_row.

Once we have an instance of the model, the schemas are all about either:
- Validation on endpoint POST / insert
- Or, important here, serialisation of the model into the JSON response needed on an endpoint.

So in this case the data flow should always go
models --> schemas --> JSON for frontend

Is there no way we can use `cls` like with all the other methods in this file? Then depending on which fields we need, we pass through a schema to filter them down 😃

If you need to include computed fields in the output from the db (the case here), then there are examples in this file of defining computed fields on the model (it would be good to separate them by a line and comment as with the others, to make it clear they are computed and no db table fields)

src/backend/app/db/models.py
```diff
@@ -877,6 +900,176 @@ async def create(
...
+                GROUP BY pt.team_id;
+                """,
+                {"project_id": project_id},
+            )
+            teams = await cur.fetchall()
+
+        return teams
+
+
+class DbProjectTeamUser(BaseModel):
```
spwoodcock, line 1035: As we are only really interested in teams when related to a specific project, we could make an exception here to merge the models into one?

Its probably not needed to get the teams from the db when not related to a db. We could merge these two classes / models, including the computed fields defined by the joins?

Just an idea - you may not think its appropriate 😃

spwoodcock (changes requested):
Looking good!

Last comments re the models!

src/backend/app/db/models.py
Anuj-Gupta4: I have shifted the pydantic model validation to routes from models file using response model and removed those imports.

src/backend/app/db/models.py
Anuj-Gupta4: That could work, but it's clearer to use separate models. Keeping them separate avoids unnecessary coupling and helps with maintainability.

spwoodcock:
Great PR - it's a big change - well done! 🎉

spwoodcock (approved):

PR #2238, opened 2025-02-28
Change user role to org admin in user list endpoint

Sujanadh:

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Previously, the project admin role was used to fetch the users list, thinking that one project admin can add another project admin but during project creation we won't have project id to check if the user is project admin or not since to create project they need to be org admin

## Describe this PR

- Update role dependency to `org_admin` in backend to fetch users list
- Passed org id as a query param from frontend

## Screenshots

N/A

## Review Guide

### DIFF
--- a/src/backend/app/users/user_routes.py
+++ b/src/backend/app/users/user_routes.py
@@ -23,7 +23,7 @@ from fastapi import APIRouter, Depends, Query, Response
 from loguru import logger as log
 from psycopg import Connection
 
-from app.auth.roles import mapper, project_manager, super_admin
+from app.auth.roles import mapper, org_admin, super_admin
 from app.db.database import db_conn
 from app.db.enums import HTTPStatus
 from app.db.enums import UserRole as UserRoleEnum
@@ -42,7 +42,7 @@ router = APIRouter(
 @router.get("", response_model=user_schemas.PaginatedUsers)
 async def get_users(
     db: Annotated[Connection, Depends(db_conn)],
-    current_user: Annotated[DbUser, Depends(project_manager)],
+    current_user: Annotated[DbUser, Depends(org_admin)],
     page: int = Query(1, ge=1),
     results_per_page: int = Query(13, le=100),
     search: str = None,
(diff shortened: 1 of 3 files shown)


spwoodcock (changes requested):
We need both the capability to access all users on FMTM (admin only), plus also users per project (project manager) I think?

So either we need a separate endpoint for these cases, or we need another way to handle this with one endpoint.

I don't think this change will work for both 😅

Sujanadh:
We also need to fetch all the users while assigning during project creation. So that would need org admin access. Like you said to fetch all the users by admin, our role management doesn't work even if I am super admin; I would need an org id or project id if the roles are org admin or project admin. I think we need to work around this if he is super admin; he shouldn't have any dependency on org and projects. That being said, it will check for all the users if they are super admin or not, which is not great. Need to rethink a way to allow that.

spwoodcock:
I would propose we have three endpoints:

- `/` the GET for the root of the users router makes sense being admin only. Return all user details.
- `/usernames` - org admin permission for project creation. Update the response model to only return user ids and usernames to display.
- `/roles` - project management, return user ids, usernames, and role (a different response model again)

Sujanadh:
- changed user role back to super admin in `/users` endpoint
- created new endpoint `users/usernames` to fetch user id and username to assign project admin : `role -> org_admin`
- created new endpoint `/users/{project-id}/project-users` to fetch users from the project and their roles

## Response:
/users/usernames

```json
[
  {
    "id": 1,
    "username": "localadmin"
  },
  {
    "id": 20386219,
    "username": "svcfmtm"
  }
]
```

src/backend/app/users/user_routes.py
```diff
@@ -64,7 +79,7 @@ async def get_user_roles(current_user: Annotated[DbUser, Depends(mapper)]):
 async def update_existing_user(
     user_id: int,
     new_user_data: user_schemas.UserUpdate,
-    current_user: Annotated[DbUser, Depends(super_admin)],
+    _: Annotated[DbUser, Depends(super_admin)],
```
spwoodcock, line 82: I like this approach for showing unused vars, similar to other languages

spwoodcock (approved):

Issue #2223, opened 2025-02-24
total_tasks field is not caluclated on the backend


spwoodcock:
**Describe the bug**

- During project creation we should calculate the total_tasks field.
- It should be present as project.total_tasks.
- It's not there for some reason - likely a failure or omission somewhere.

**To Reproduce**

- All projects on prod except two old ones, suggesting this never worked?

**Expected behavior**

- The value is populated during project creation.

Anuj-Gupta4:
Total tasks are populated as expected. Verified from dev server.

PR #2246, opened 2025-03-04
Add migration file to populate total tasks for old projects if null

Anuj-Gupta4:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related to: #2223

## Describe this PR

Added a python file that can be manually run to fix total tasks missing from some projects.

## Usage

```bash
docker compose build api
docker compose up -d
docker compose exec api bash
cd migrations
python populate_total_tasks.py
```

### DIFF
--- a/src/backend/migrations/populate_total_tasks.py
+++ b/src/backend/migrations/populate_total_tasks.py
@@ -1,0 +1,32 @@
+"""Populate total_tasks for old projects."""
+
+import asyncio
+
+from psycopg import AsyncConnection
+
+from app.config import settings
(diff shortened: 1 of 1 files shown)


spwoodcock (approved):
This might have been better as an SQL migration, to save it having to be run manually, but otherwise it looks good for what we need 🙌

PR #2263, opened 2025-03-11
Set osm_id property to negative value if the geom is not from OSM

spwoodcock:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes https://github.com/hotosm/osm-fieldwork/issues/344, as we no longer will create entities via XLSForm (moved to entity creation via API)

## Describe this PR

- Rob suggested this is a convention: negative numbers for geoms that come from outside OSM.
- Updated the mandatory fields check to reflect that.
- Hopefully doesn't break anywhere we use the `osm_id` in the code 🤞 

### DIFF
--- a/src/backend/app/db/postgis_utils.py
+++ b/src/backend/app/db/postgis_utils.py
@@ -325,13 +325,21 @@ def add_required_geojson_properties(
             properties["osm_id"] = feature_id
         else:
             osm_id = (
-                properties.get("osm_id")
+                properties.get(
+                    "osm_id"
+                )  # osm_id property takes priority (i.e. it's from OSM)
                 or properties.get("id")
-                or properties.get("fid")
+                or properties.get("fid")  # fid is typical from tools like QGIS
                 # Random id
                 # NOTE 32-bit int is max supported by standard postgres Integer
                 or getrandbits(30)
             )
+
+            # Ensure negative osm_id if it's derived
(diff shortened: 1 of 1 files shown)


src/backend/app/db/postgis_utils.py
```diff
@@ -325,13 +325,21 @@ def add_required_geojson_properties(
...
-                or properties.get("fid")
+                or properties.get("fid")  # fid is typical from tools like QGIS
                 # Random id
                 # NOTE 32-bit int is max supported by standard postgres Integer
                 or getrandbits(30)
             )
+
+            # Ensure negative osm_id if it's derived
+            # NOTE this is important to denote the geom is not from OSM
+            if not properties.get("osm_id"):
```
Sujanadh, line 340: Are we making osm id negative even though it is derived from id or fid?

spwoodcock: Yes to help distinguish it from actual OSM IDs - in the submission processing we could possibly revert to a positive into if the original IDs are required by the user 👍

Sujanadh: ok make sense :+1:

Sujanadh (approved):

PR #2268, opened 2025-03-12
DRAFT: web forms 2025 03 11 v2

DanielJDufour:


spwoodcock:
As it's only enabled via feature flag, should we merge in an iterate over progressive PRs?

spwoodcock:
Will do 👍

Note: uses URL based feature flag `webforms=true` to enable

PR #2278, opened 2025-03-14
Allow config to create projects with no existing geoms (new geom only)

### LINKED ISSUE #2270: Is it possible to only map new geometries for a project (no existing geoms)

- We have a use case for mapping when there is no data in the area at all.
- This is often the case when users don't want to map in OSM (generally OSM has at least **some** data).
- In this scenario we should probably provide an option to the user (as currently project creation would be blocked).

- Add an option to the 'Map Data' tab during project creation to 'start with no data (new geometries only)'.

**Additional context**

- We also need to consider how to handle splitting in this case.
- We probably just use the project area as a single task area by default?
  - It's not perfect to have to click the project area for the popups in that case, but it works for now.

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #2270

## Describe this PR

- Added extra option to create a project with no geometries.
- For projects when only new geom needs to be collected.
- It's a bit messy / not perfect - I did this in a rush - but project creation works.
- Needs to be tested thoroughly.

### DIFF
--- a/src/frontend/src/api/CreateProjectService.ts
+++ b/src/frontend/src/api/CreateProjectService.ts
@@ -60,34 +60,29 @@ const CreateProjectService = (
 
       // Upload data extract
       let extractResponse;
-      if (isOsmExtract) {
-        // Generated extract from raw-data-api
-        extractResponse = await API.get(
-          `${VITE_API_URL}/projects/data-extract-url?project_id=${projectId}&url=${projectData.data_extract_url}`,
-        );
-      } else if (dataExtractFile) {
-        // post custom data extract
-        const dataExtractFormData = new FormData();
-        dataExtractFormData.append('data_extract_file', dataExtractFile);
-        extractResponse = await API.post(
+      if (isEmptyDataExtract) {
+        // Manually set response as we don't call an API
+        extractResponse = { status: 200 };
+      }
+      else if (dataExtractFile) {
+      const dataExtractFormData = new FormData();
+      dataExtractFormData.append('data_extract_file', dataExtractFile);
+      extractResponse = await API.post(
         `${VITE_API_URL}/projects/upload-data-extract?project_id=${projectId}`,
         dataExtractFormData,
       );
-      } else if (isEmptyDataExtract) {
-        // Manually set response as we don't call an API
-        extractResponse = { status: 200 };
-      } else {
-        const msg = 'Neither isOsmExtract or dataExtractFile was set';
-        console.error(msg);
-        throw new Error(msg);
-      }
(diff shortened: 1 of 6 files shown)


Sujanadh:
For no data extracts, we can allow them to use split by square or choose project AOI as a task. This kind of project doesn't need to create entity; we can create dataset during project creation and avoid entity creation. I see your inline comment syaing status need to be set to avoid downstream error, what does that mean? failure of status api?

spwoodcock:
> For no data extracts, we can allow them to use split by square or choose project AOI as a task. This kind of project doesn't need to create an entity; we can create a dataset during project creation and avoid entity creation. I see your inline comment saying status needs to be set to avoid downstream errors; what does that mean? failure of status API?

Yeah I think so - having no status for an entity causes an error when its serialised as a JSON.

And yep I agree split by square or selecting the project AOI are the best options for this type of project. Interestingly, split by sql task splitting also worked, as it downloaded the osm data extract and split on that 😄

If avoiding the creation of an existing entities works and causes no errors, then that is a good option! 👍

Sujanadh:
Oh yes, splitting works :rofl:. I guess so; not creating entities is possible, but the dataset needs to be there.

spwoodcock:
@Sujanadh did you manage to check if creating a project without any existing entities works?

Let me know if this PR will be modified 🙏

Sujanadh:
yes i tested project creation without entity creation and it was working. But the mapper frontend wasn't working it seemed like some issue related to the svelte package?

Today i am on leave so can't be involved directly.

src/frontend/src/api/CreateProjectService.ts
```diff
@@ -41,7 +42,9 @@ const CreateProjectService = (
       await dispatch(CreateProjectActions.PostProjectDetails(projectCreateResp));
 
       if (!hasAPISuccess) {
-        throw new Error(`Request failed with status ${projectCreateResp.status}`);
+        const msg = `Request failed with status ${projectCreateResp.status}`;
+        console.error(msg);
```
Sujanadh, line 46: do we need consoles here?

Sujanadh:
@spwoodcock after you review, this PR is good to go

src/frontend/src/api/CreateProjectService.ts
spwoodcock: Yeah I think it makes sense, as it's quite a major error if the project creation fails. It can help debugging what went wrong

Issue #2266, opened 2025-03-11
Add hashtags to geojson submission output


adiatmad:
**Is your feature request related to a problem? Please describe.**  
When using Field Mapping Tasking Manager, users often forget to include the required hashtag when uploading data to OpenStreetMap. This makes it difficult to track contributions for specific projects, especially for monitoring and evaluation purposes.  

**Describe the solution you'd like**  
We propose an automatic hashtag inclusion feature in Field Mapping Tasking Manager. When users upload data to OpenStreetMap, the system should ensure that the predefined project-specific hashtag is included. This will streamline data tracking and improve consistency across all contributions.  

**Describe alternatives you've considered**  
- Providing manual reminders in the project description, but this is not reliable as users may still forget.  
- Using external scripts or plugins requires additional setup and is not user-friendly for all mappers.  

**Additional context**  
This feature is critical for an upcoming pilot project in April/May 2025, where proper contribution tracking is essential. Implementing this would enhance data accountability and streamline project managers' workflows.

Thank you for your hard work!

Sujanadh:
So when a user downloads data extracts or submissions, they should have hashtags included in them.

spwoodcock:
I think we need two things here:
- Add the hashtags to the final submission output on the **backend**, as Sujan says 👍 
- On the **frontend**:
  - Previously we had the hashtags present in the UI for the user hashtags input box.
  - Now instead we have an empty box (to make user input clearer), plus a comment below explaining we add the `#FMTM` tag automatically.
  ![Image](https://github.com/user-attachments/assets/665354c0-2d4f-4f13-aa2e-abf386038e6a)
  - We need one of two things here for a better user experience:
    - Add to the comment to say we also include the `#fmtm.hotosm.org-{project_id}` tag automatically too.
    - Display the `#FMTM` and `#fmtm.hotosm.org-123` tags in the box (not removable) so it's clear immediately to the user that these are included.

spwoodcock:
We just had our team call - the key actionable thing here:

- Append the project hashtags to **each feature** in the final submission geojson output (under the properties key).

The team can load the geojson into JOSM and the properties keys can be pulled out into the changeset data generated.

NSUWAL123:
There's an issue related to user-filled hashtags not being saved on the db. 

When creating a project with user-filled hashtags, the POST project creation API response includes the user-filled hastags. But the GET `/project/:id` returns only the default hashtags applied on the backend.

[...]

PR #2281, opened 2025-03-17
Add hashtags to entity properties, data extracts and submissions

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue: #2266 

## Describe this PR

- Hashtags provided during project creation are added in project features while converting them into fgb data
- Added hashtags are also used in dataset and entity creation
- Added calculated field to add hashtags of corresponding selected feature in submission data, it is also included in submission downloads

## Screenshots

**Entities**
image

**Submissions**
image

## Review Guide

### DIFF
--- a/src/backend/app/submissions/submission_crud.py
+++ b/src/backend/app/submissions/submission_crud.py
@@ -164,8 +164,16 @@ async def get_submission_by_project(
         ValueError: If the submission file cannot be found.
 
     """
+    hashtags = project.hashtags
     xform = get_odk_form(project.odk_credentials)
-    return xform.listSubmissions(project.odkid, project.odk_form_id, filters)
+    data = xform.listSubmissions(project.odkid, project.odk_form_id, filters)
+
+    def add_hashtags(item):
+        item["hashtags"] = hashtags
+        return item
+
+    data["value"] = list(map(add_hashtags, data["value"]))
+    return data
 
 
 async def get_submission_detail(


spwoodcock:
Haven't had time to consider this thoroughly. But we only need the hashtags for data output (submissions) and not during any of the mapping steps.

Wouldn't it make more sense to dynamically inject the hashtags into the submissions based on the projects.hashtags field?

This way, if the hashtags change (say one is forgotten at project create), then its easy to update the value and regenerate submission output (instead of having this set on each entity).

My other concern is that properties are expensive for entities. Sending all the data also takes bandwidth, so best to minimise

Sujanadh:
I thought we might need to have hashtags in features as well while downloading; what do you think? If we don't need that, then we can add hashtags only in submissions.

spwoodcock:
> I thought we might need to have hashtags in features as well while downloading; what do you think? If we don't need that, then we can add hashtags only in submissions.

Not a bad idea, but I don't think its essential 👍

The key thing is hashtags in submission output

Sujanadh:
Doing so we can't add hashtags in submission when they want to download submission in csv format as we directly fetch csv file media from the central.

spwoodcock:
Thats a good point 👍

I think its not a deal breaker as CSV format isn't really used for OSM upload, only for export by the user (e.g. for inspection)

Sujanadh:
In that case, I will push new changes.

Issue #2153, opened 2025-02-05
Option to login via Google account on mapper frontend


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- The mapper frontend allows login via OSM, and the option of no login at all (user not attributed for edits).
- We encountered a case when the manager wants to track the mapper progress (to pay them according to buildings mapped), but signing up 50+ mappers to OSM is clunky and prone to issues.

**Describe the solution you'd like**

- Add another auth option, likely Google login to start.
- The logic would be ported to https://github.com/hotosm/ui/issues/1 once developed.

**Describe alternatives you've considered**

- Microsoft, LinkedIn, Facebook, Github, etc...

**Additional context**

- OAM has some ancient code to do that, that is still functional.

manjitapandey:
when there is error in logging in through osm account on mapper frontend now. 
steps to reproduce:
1. login through google account.
2. Logout from google and try loging in from osm account
![Image](https://github.com/user-attachments/assets/e8e82739-b0e7-458b-b8ff-f55cd89f20e6)
@Sujanadh , @NSUWAL123

PR #2284, opened 2025-03-18
Add Google OAuth support for authentication

Anuj-Gupta4:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related to: #2153 

## Describe this PR

Add additional authentication option (Google)

## Screenshots

## Review Guide


Anuj-Gupta4:
I am considering converting user's IDs to varchar because user ID gained from Google are large numbers (This would require dropping every user foreign key constraint and recreating them as varchar). We cannot use auto incrementing IDs because it may conflict with new OSM user IDs in the future.

spwoodcock:
Good thinking 😃

Its not a simple migration.
We didn't really consider auth other than OSM in the original schema.

IMO the `id` field should be renamed `sub` and set to varchar as you say. Then the values should be like `osm|73844` or `google|738374748`.

Then to save updating all refs to the users.is field in the code, we should add a computed field to the pydantic model called `id`, which should split the sub field on the `|` character and return the integer id. We could also compute `login_provider` if needed too (google, osm).

### Mapper frontend only

Consider we only want other auth methods on the mapper frontend only. If the user calls the refresh/manager endpoint and is logged in via google, we want to log them out (and possibly provide an error message to say they should log in again via OSM).

### Migrations 

Check all refs to the users.id constraints

```sql
SELECT conrelid::regclass AS users, conname AS the_constraint_name_from_db
FROM pg_constraint
WHERE confrelid = 'users'::regclass;
```

1. Create new column sub as varchar
2. Populate new column with id field values, plus `osm|`
3. Drop foreign key constraints on referencing tables
4. Drop the users.id field
5. Repeat the process to create a new varchar column, perhaps rename `user_sub` and copy across values with new `osm|` values, then drop the original. Do this for each table referencing user_id.
6. Create new foreign key constraints
7. Verify (check for orphan records), e.g.

```sql
SELECT * FROM organisation_managers
LEFT JOIN users ON organisation_managers.user_sub = users.sub
WHERE organisation_managers.user_sub IS NULL;
```

8. Recreate indexes

src/backend/app/auth/auth_routes.py
```diff
@@ -49,27 +50,44 @@
...
     """
     login_url = osm_auth.login()
-    log.debug(f"Login URL returned: {login_url}")
+    log.debug(f"OSM Login URL returned: {login_url}")
     return JSONResponse(content=login_url, status_code=HTTPStatus.OK)
 
 
-@router.get("/callback")
-async def callback(
+@router.get("/google-login")
```
spwoodcock, line 71: Could this maybe be /login/google and /login/osm for better consistency?

.env.example
```diff
@@ -21,6 +21,12 @@ OSM_CLIENT_SECRET=${OSM_CLIENT_SECRET}
 OSM_URL=${OSM_URL:-"https://www.openstreetmap.org"}
 OSM_SCOPE=${OSM_SCOPE:-'["read_prefs","send_messages"]'}
 OSM_LOGIN_REDIRECT_URI=http${FMTM_DOMAIN:+s}://${FMTM_DOMAIN:-127.0.0.1:7051}/osmauth
+
+### Google ###
+GOOGLE_CLIENT_ID=${GOOGLE_CLIENT_ID}
+GOOGLE_CLIENT_SECRET=${GOOGLE_CLIENT_SECRET}
+GOOGLE_LOGIN_REDIRECT_URI=http${FMTM_DOMAIN:+s}://${FMTM_DOMAIN:-127.0.0.1:7051}/googleauth
```
spwoodcock, line 28: Google accepts localhost as a redirect URL, so we don't need a hacky workaround using 127.0.0.1 and port 7051.

The redirect should work with `${FMTM_DOMAIN}${FMTM_DEV_PORT+:}${FMTM_DEV_PORT}` or similar

src/backend/app/auth/auth_routes.py
```diff
@@ -173,37 +229,27 @@ async def get_or_create_user(
...
-                detail=f"User ID ({user_data.id}) could not be inserted in db",
+                detail=f"User ({user_data.sub}) could not be inserted in db",
             )
 
         return db_user_details
 
     except Exception as e:
         await db.rollback()
         log.exception(f"Exception occurred: {e}", stack_info=True)
-        if 'duplicate key value violates unique constraint "users_username_key"' in str(
```
spwoodcock, line 184: How come we removed this unique constraint for the username? Was it with the thought that the user could have the same username in OSM and Google? (If so, that does make sense)

src/backend/app/auth/auth_routes.py
```diff
@@ -220,10 +266,38 @@ async def refresh_management_cookies(
...
     This endpoint is specific to the management desktop frontend.
     Any temp auth cookies will be ignored and removed.
-    OSM login is required.
+    Authentication is required.
+    If signed in with login method other than OSM, the user will be logged out and
+    a forbidden status will be returned.
 
     NOTE this endpoint has no db calls and returns in ~2ms.
     """
+    if "google" in current_user.sub.lower():
```
spwoodcock, line 275: An inverse check would probably be best and more future proof.

If 'osm' not in ...

For if we add additional login methods in future

src/backend/app/auth/auth_routes.py
```diff
@@ -220,10 +266,38 @@ async def refresh_management_cookies(
...
+        )
+        fmtm_cookie_name = settings.cookie_name
+        refresh_cookie_name = f"{fmtm_cookie_name}_refresh"
+
+        for cookie_name in [
+            fmtm_cookie_name,
+            refresh_cookie_name,
+        ]:
+            log.debug(f"Resetting cookie in response named '{cookie_name}'")
+            response.set_cookie(
```
spwoodcock, line 288: Should this be in a separate function like 'expire_cookies', to reuse and avoid duplication?

src/backend/app/organisations/organisation_crud.py
```diff
@@ -53,7 +53,7 @@ async def init_admin_org(db: Connection) -> None:
 
     # Create service user
     svc_user = UserIn(
-        id=20386219,
+        sub="fmtm|20386219",
```
spwoodcock, line 56: Should this be osm|20386219 like the others?

spwoodcock (changes requested):
Huge refactor! Well done!

I can't test the migration - thats the key part here - but I just left some minor comments.

This should allow us to add any auth method we may want in future 🙌

src/backend/app/auth/auth_routes.py
Anuj-Gupta4: Yes. That was the reason.

src/backend/app/auth/auth_routes.py
Anuj-Gupta4: I didn't do that because it tries to expire the local admin user as well when debug is true.

spwoodcock (approved):
If the user doesn't configure the google env variables, what happens?

Can we provide an informative message to the user to say that Google authentication is not enabled?

spwoodcock (approved):

Anuj-Gupta4:
Merging since approved...

PR #2290, opened 2025-03-19
Prefill non-removable hashtag

NSUWAL123:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #2266

## Describe this PR
- Add to the comment to say we also include the #fmtm.hotosm.org-{project_id} tag automatically too.
- Display the #FMTM and #fmtm.hotosm.org-123 tags in the box (not removable) so it's clear immediately to the user that these are included.

## Screenshots
image


spwoodcock:
Could the input box be increased in height to show two rows of tags possibly?

With the default tags displayed below? And the user input tags above?

Or something similar to this to improve the user input. Its a bit messy having them as the first tags and then the user input after them (likely going off the screen and needing to scroll across the box)

spwoodcock:
Or alternatively just show the default tags at the end of the list somehow

NSUWAL123:
Updated the hashtags field with absolute positioning of the static default hashtag giving it a slight bg color to distinguish it from other hashtags.
image
image

spwoodcock:
Looks better!

Sorry to be a pain with another idea 😅

Do you think that having 'chips' would be possible for the hashtags instead?

I think chips would be much more user friendly and more obvious what is included (over a comma separated string)

NSUWAL123:
So you mean to say that there would be removable chips like this
image

where the two default tags on chips would be disabled, and there would be an input element with a plus (+) button to insert a new hashtag. where there should be validation that the hashtag should not contain spaces. 
@spwoodcock

spwoodcock:
> So you mean to say that there would be removable chips like this
> image
> 
> where the two default tags on chips would be disabled, and there would be an input element with a plus (+) button to insert a new hashtag. where there should be validation that the hashtag should not contain spaces. 
> @spwoodcock 

Sounds perfect!

I think that chips are just much more user friendly than a comma separated string 😃

spwoodcock:
Should we merge this, then possibly update via another PR? Or modify via this PR?

NSUWAL123:
Closing this as I'll make another PR to work on this instead of reverting the work!

Issue #2179, opened 2025-02-12
Comments posted by validators should be visible to enumerator in field.


manjitapandey:
**Is your feature request related to a problem? Please describe.**
Feedback provided by Tokha team,
Currently, during validation, Validators can post the comment. That is mostly used when user something is wrong with the submission and needs resurvey. But enumerator in field are not being able to know what is wrong and why the submission has been rejected. Communicating through phone call and social media is also difficult since its very hard to communicate the building. 

**Describe the solution you'd like**
A section on feature to show the comment if any from validator's side to the mapper. This way mapper will know what's wrong with their previous form and why has the submission marked has issues. 

**Describe alternatives you've considered**
If possible two way communication could be better but if we don't wanna over complicate the workflow. One way communication could also work. Meaning, Mapper can only see the comments if validators post any reason but can't reply back. 

**Additional context**
Further design on mapper's page need to be decided on how to visualise the comment.

spwoodcock:
Hmm this is a good idea, but we have no way to add comments to features in FMTM currently.

### Identify the building with issues

The following should be possible though:

- Mark feature as BAD.
- The colour of the feature is updated / flashing

Does this not help identify the building?

### Notify user with manager comments

The manager could also add a comment to the task for the user to view, but I agree this probably isn't very intuitive to the user.

As we have realtime notifications for task comments already, perhaps we could add the functionality of tagging the user? If the manager specifies `@username` in the comment, we can create a popup in the centre of the users screen instantly, to notify them?

(Of course this only works when they have FMTM open)

spwoodcock:
> >  but we have no way to add comments to features in FMTM currently.
> 
> Actually we can add feature level comment when validating a feature.
> 
> >  but we have no way to add comments to features in FMTM currently.
> 
> this helps a lot in identifying the building however, its also important to communicate with them what they are doing wrong through comments.
> 
> > As we have realtime notifications for task comments already, perhaps we could add the functionality of tagging the user? If the manager specifies @username in the comment, we can create a popup in the centre of the users screen instantly, to notify them?
> 
> this is also a good idea. But what I was thinking was we could visualise the comment when mapper clicks on pop-up of that feature. Something like this.
> 
> ![Image](https://github.com/user-attachments/assets/c51ffabb-7f46-453e-a672-992d24340bc9)
> 
> 

Oh I get you, let's definitely do that!

[...]

PR #2312, opened 2025-03-31
omment tagging

NSUWAL123:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #2179

## Describe this PR
- Submission comment username mention

## Screenshots

## Review Guide

### DIFF
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -90,6 +90,7 @@
     "pako": "^2.1.0",
     "pmtiles": "^3.0.6",
     "qrcode-generator": "^1.4.4",
+    "rc-mentions": "^2.19.1",
     "react": "^19.0.0",
     "react-datepicker": "^6.9.0",
     "react-dom": "^19.0.0",
--- a/src/frontend/src/store/slices/UserSlice.ts
+++ b/src/frontend/src/store/slices/UserSlice.ts
@@ -20,6 +20,8 @@ export const initialState: UserStateTypes = {
   updateUserRoleLoading: false,
   userListForSelect: [],
   userListForSelectLoading: false,
+  userNames: [],
+  getUserNamesLoading: false,
 };
 
 const UserSlice = createSlice({
@@ -45,6 +47,12 @@ const UserSlice = createSlice({
     },
     SetUserListForSelectLoading: (state, action: PayloadAction<boolean>) => {
       state.userListLoading = action.payload;
+    },
+    SetUserNames: (state, action: PayloadAction<UserStateTypes['userNames']>) => {
+      state.userNames = action.payload;
+    },
+    GetUserNamesLoading: (state, action: PayloadAction<boolean>) => {
+      state.getUserNamesLoading = action.payload;
     },
   },
 });
--- a/src/frontend/pnpm-lock.yaml
+++ b/src/frontend/pnpm-lock.yaml
@@ -143,6 +143,9 @@ importers:
       qrcode-generator:
         specifier: ^1.4.4
         version: 1.4.4
+      rc-mentions:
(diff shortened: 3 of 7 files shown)


spwoodcock:
Looks good!

src/frontend/package.json
```diff
@@ -90,6 +90,7 @@
     "pako": "^2.1.0",
     "pmtiles": "^3.0.6",
     "qrcode-generator": "^1.4.4",
+    "rc-mentions": "^2.19.1",
```
spwoodcock, line 93: The more react specific components we add, the more we will need to refactor in future.

I dont see the future of the management frontend being in React (in the long run).

As much as possible, could we try to use web components instead?

For example, what about this text expander web component from Github?
https://github.com/github/text-expander-element

It also picks up a specific tag (e.g. @ or #) from a text input, with a hook to do something.

I would suggest we only query the backend after 3 characters or more are typed. Hopefully this narrows it down a fair bit before searching the db

src/frontend/src/store/slices/UserSlice.ts
```diff
@@ -46,6 +48,12 @@ const UserSlice = createSlice({
     SetUserListForSelectLoading: (state, action: PayloadAction<boolean>) => {
       state.userListLoading = action.payload;
     },
+    SetUserNames: (state, action: PayloadAction<UserStateTypes['userNames']>) => {
```
spwoodcock, line 51: This will set all available usernames from the backend into the browser right?

If we have 10,000 users at some point, do you think this would cause an issue?

No need to over-optimise at the start (we dont have 10,000 users...), but its always good to keep scalability in mind when designing solutions 👍

spwoodcock (changes requested):
This is pretty cool 🙌

src/frontend/package.json
NSUWAL123: Sure! I'll have a look at text-expander-element. 
Also completely agree with you on querying after 3 characters

src/frontend/src/store/slices/UserSlice.ts
NSUWAL123: So currently instead of showing all user list, I am querying if there's 1 character, but will change to 3 min character as you mentioned!

spwoodcock:
As a tradeoff for performance and scalability, we could consider the following:
- On first 3+ character input (e.g. "@chr"), query the API.
- Store the result (["chris", "christina", "chrissy"]) in the store (redux or svelte).
- As the user continues to type (e.g. "@chri"), filter the local store list.
- Then finally, if the user selects an option, or deletes characters (< 3 chars), or deletes the `@`, then we clear the local store values.

NSUWAL123:
@spwoodcock tried implementing `text-expander-element` but doesn't seem to work. 
1. When using it with redux-store, the list doesn't update (the states are one state behind) due to the list populating after the change event trigger & list being built inside the change event.
2. Also when directly fetching the list under the change event, the list can be consoled but doesn't render. It's very weird as the static list would render.  
Also tried many other ways but doesn't work 😵‍💫

spwoodcock:
No worries @NSUWAL123 - it was maybe a bit ambitious to use that web component, as I haven't tested it myself!

It's not a huge deal, as the React frontend doesn't really use any web-components anyway - updating most things to use them would be a big undertaking.

The key thing is to ensure we are using web components everywhere possible in the **mapper** frontend.

Thanks for trying though!! 🙏

PR #2331, opened 2025-04-03
Google login add to mapper frontend

NSUWAL123:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #2153

## Describe this PR
- Google login add to mapper frontend

## Screenshots
[Screencast from 3-4-25 02:13:34 अपराह्न +0545.webm](https://github.com/user-attachments/assets/26963fa0-18d9-45ad-8276-32761e3e9845)


spwoodcock:
The backend API has a port range specified of
```
      - "7052-7055:8000"
```
so we probably can't use port 7055 for the mapper frontend

spwoodcock:
I just updated it to port `7053` and port binded it in compose.yaml 👍 

https://github.com/hotosm/fmtm/commit/3bb85448d7760a03a22d6e3db101258dbafa4195

Sujanadh:
oh okay I thought mapper runs on port 55
` mapper.fmtm.localhost_REVERSE_PROXY_HOST: http://ui-mapper:7055` in proxy service

Sujanadh:
> I just updated it to port `7053` and port binded it in compose.yaml 👍
> 
> [3bb8544](https://github.com/hotosm/fmtm/commit/3bb85448d7760a03a22d6e3db101258dbafa4195)

okay :+1:

NSUWAL123:
So basically we will have two endpoints like: 
- /auth/login/osm/mapper ----> that returns mapper frontend URI
- /auth/login/osm/management ----> that returns management frontend URI ??
@spwoodcock @Sujanadh

spwoodcock:
> oh okay I thought mapper runs on port 55 ` mapper.fmtm.localhost_REVERSE_PROXY_HOST: http://ui-mapper:7055` in proxy service

yeah it does, but we can't expose the port to the local machine, as the port is blocked by the API port range

> So basically we will have two endpoints like:
>
>    /auth/login/osm/mapper ----> that returns mapper frontend URI
>    /auth/login/osm/management ----> that returns management frontend URI ??
>    @spwoodcock @Sujanadh

yeah that could make sense 👍 (we would need to also update the redirect URLs set in the OSM OAuth applications on each instance after this too

spwoodcock:
Will also update the google auth one too!

```python
    @computed_field
    @property
    def google_login_redirect_uri(self) -> str:
        """The constructed Google redirect URL for mapper frontend.

        Must be set in the OAuth2 config for the Google profile.
        """
        if self.DEBUG:
            uri = "http://127.0.0.1:7055/googleauth"
        uri = f"https://mapper.{self.FMTM_DOMAIN}/googleauth"
        os.environ["GOOGLE_LOGIN_REDIRECT_URI"] = uri
        return uri
```

Although I'm surprised this works, as port 7055 goes to the API I believe?

Sujanadh:
Do we have to set env if we are computing redirect uri? I think that is unnecessary.
`os.environ["MANAGER_OSM_LOGIN_REDIRECT_URI"] = uri`

Sujanadh:
Can we use two redirect for same client ID and client secret? @spwoodcock

spwoodcock:
> Do we have to set env if we are computing redirect uri? I think that is unnecessary. `os.environ["MANAGER_OSM_LOGIN_REDIRECT_URI"] = uri`

yes unfortunately, as an all caps function name is invalid Python in most linters.

A lowercase function name is used, then the all caps env var is set via os.environ

Sujanadh:
I couldn't access the var `MANAGER_OSM_LOGIN_REDIRECT_URI` from settings when i try to test it.

spwoodcock:
> Can we use two redirect for same client ID and client secret? @spwoodcock

I'm not sure I understand - can you explain a bit more?

spwoodcock:
> I couldn't access the var `MANAGER_OSM_LOGIN_REDIRECT_URI` from settings when i try to test it.

Ah of course - apologies - I did that in too much of a rush.

You are right, in this case no need to set the env var at all. Just use the lowercase setting values 👍

Sujanadh:
So when we create an OAuth application in an OSM account, I have already created osm client variables from the redirect url `127.0.0.1:7051/osmauth` so if I edit it and add another redirect uri `127.0.0.1:7053/osmauth` will it work? so that we could use same client id and client secret for both management and mapper login

spwoodcock:
Oh right, yes that should be fine!

You can add multiple redirect_urls in openstreetmap:

image

It's just a list of **valid** redirect URLs, so can be many

Sujanadh:
Ok so we won't need extra vars for mapper login; I will update it.

spwoodcock:
> Ok so we won't need extra vars for mapper login; I will update it.

No extra secret vars etc, but we need separate redirect uri vars to direct to the correct application 👍

NSUWAL123:
The login is working now! Thanks @spwoodcock @Sujanadh @Anuj-Gupta4

Sujanadh:
There was a port binding issue with `7053`. I updated it to `7057`.

Sujanadh (approved):

spwoodcock (approved):

spwoodcock:
Nice work guys 🙌

Issue #2232, opened 2025-02-25
Increase test coverage on backend


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- We have some good backend tests for endpoints.
- They are invaluable for checking changes to the backend.
- It allows for fast changes as we have more assurances our code didn't break anything downstream.

**Describe the solution you'd like**

- More test cases for the backend APIs.
- Functionality is reasonably stable now, so we should add tests for most.
- Coverage of about 80% with good edge case coverage would be great going forward.
- This is key to avoid breaking changes when we add new features or modify existing code.
- The response for each endpoint should be inspected and checked for the essential fields it should have, plus to check if they are valid values.

PR #2346, opened 2025-04-04
Add some tests for project and task routes

Anuj-Gupta4:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [x] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related to: #2232 

## Describe this PR

Tests to:

- read project
- update project form
- download project form
- get contributors
- add new pm and get project user
- create entity
- download project boundary
- download task boundary
- create geometry log
- read geometry log
- delete geometry log
- read tasks
- get specific task
- more task events
- get task activity

### DIFF
--- a/src/backend/app/projects/project_routes.py
+++ b/src/backend/app/projects/project_routes.py
@@ -817,7 +817,7 @@ async def update_project_form(
                 "project_id": project.id,
             },
         )
-        db.commit()
+        await db.commit()
 
     return JSONResponse(
         status_code=HTTPStatus.OK,
--- a/src/backend/tests/test_projects_routes.py
+++ b/src/backend/tests/test_projects_routes.py
@@ -32,7 +32,7 @@ from loguru import logger as log
 
 from app.central.central_crud import create_odk_project
 from app.config import settings
-from app.db.enums import EntityState, HTTPStatus, MappingState
+from app.db.enums import EntityState, HTTPStatus, MappingState, ProjectRole
 from app.db.models import DbProject, slugify
 from app.db.postgis_utils import check_crs
 from app.projects import project_crud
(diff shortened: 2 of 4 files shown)


Sujanadh (approved):

src/backend/app/projects/project_routes.py
```diff
@@ -817,7 +817,7 @@ async def update_project_form(
                 "project_id": project.id,
             },
         )
-        db.commit()
```
spwoodcock, line 820: Good catch, but I'm not sure we really need db.commit() anywhere in the codebase to be honest:
- When an endpoint is called a transaction is started via a context manager
- Before the cursor / transaction is closed, the SQL is executed automatically

src/backend/tests/test_projects_routes.py
```diff
@@ -527,6 +527,233 @@ async def test_project_task_split(client):
...
+    assert data["organisation_logo"] is None
+
+
+async def test_update_and_download_project_form(client, project):
+    """Test updating and downloading the XLSForm for a project."""
+    updated_xls_content = b"updated xlsform data"
+    xls_file = BytesIO(updated_xls_content)
+    xls_file.name = "form.xlsx"
+
+    with patch(
```
spwoodcock, line 558: While this patch is good, we do run the entire stack for these tests (db, api, odk central), so you could get the actual response rather than doing a patch 👍

src/backend/tests/test_projects_routes.py
```diff
@@ -527,6 +527,233 @@ async def test_project_task_split(client):
...
+
+    assert isinstance(data, list)
+    assert len(data) > 0
+    assert all(
+        "user" in contributor and "contributions" in contributor for contributor in data
+    )
+
+    contributor = data[0]
+    assert contributor["user"] == admin_user.username
+    assert contributor["contributions"] == 2
```
spwoodcock, line 605: It might be good to have this as `>2` or something, as this could break in future if we added more tests that create task events

spwoodcock:
Test coverage increase 68% --> 72% 🙌

Issue #2337, opened 2025-04-03
Check that form media (question icons) can be accessed via backend API easily


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- Multiple choice question types with icons in ODK work via form media.
- The user uploads the icons attached to the form.
- We need a way to access these icons stored in ODK Central.

**Describe the solution you'd like**

- Does ODK Central provide a public link to the form media for display in a web page?
- Can we upload the form media to S3 instead?
- Consider the options.

## Updated Info (From Linked PR)

- We need a way to get the form attachments from ODK --> mapper frontend (display in web forms).
- There is a test implementation in this PR for osm-fieldwork, but what we probably actually need is a wrapper for these two endpoints:
  - https://docs.getodk.org/central-api-form-management/#listing-form-version-attachments
    - First, we need the version of the form.
    - Next we call this API to get a list of attachments.
    - We filter out the image attachments: JPG, PNG
  - https://docs.getodk.org/central-api-form-management/#downloading-a-form-version-attachment
    - Then for each attachment we are interested in, we get the data. HOWEVER, the docs state this is the blob data.
    - After the S3 update to ODK Central, if the blob is synced to S3, I actually think this will return a pre-signed URL.
    - We should forward a list of pre-signed URLs as an API response (the same as for photo attachments).

TODO:

- [x] Below OdkCentralAsync.listFormAttachments, add OdkCentralAsync.getFormAttachmentUrls. This method will be similar to getSubmissionAttachmentUrls, but for uploaded and synced **form attachments**.
- [x] Add a snippet like this to the `/central/upload-form-media` endpoint, to ensure the attachment blobs are synced db --> S3:
    ```python
        # Ensure S3 form attachments are uploaded to S3
        async with OdkCentral(
            url=project.odk_credentials.odk_central_url,
            user=project.odk_credentials.odk_central_user,
            passwd=project.odk_credentials.odk_central_password,
        ) as odk_central:
            await odk_central.s3_sync()
    ```
- [x] Then when the user calls the `/central/get-form-media` endpoint (when a project is loaded), we can simply call the OdkCentralAysnc.getFormAttachmentUrls method to get a list of URLs. The images will be downloaded by the frontend and embedded via CSS for usage.

## How to test this

- Use this test form to create a project: [mdp_xlsform.xlsx](https://github.com/user-attachments/files/19623877/mdp_xlsform.xlsx)
[...]

PR #2348, opened 2025-04-06
Allow for download of form media/attachments as list of URLs pt1

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Related #2337

## Describe this PR

Work in progress. @Anuj-Gupta4 could you look into this please 🙏 

- As described in the issue, we need a way to get the form attachments from ODK --> mapper frontend (display in web forms).
- There is a test implementation in this PR for osm-fieldwork, but what we probably actually need is a wrapper for these two endpoints:
  - https://docs.getodk.org/central-api-form-management/#listing-form-version-attachments
    - First, we need the version of the form.
    - Next we call this API to get a list of attachments.
    - We filter out the image attachments: JPG, PNG
  - https://docs.getodk.org/central-api-form-management/#downloading-a-form-version-attachment
    - Then for each attachment we are interested in, we get the data. HOWEVER, the docs state this is the blob data.
    - After the S3 update to ODK Central, if the blob is synced to S3, I actually think this will return a pre-signed URL.
    - We should forward a list of pre-signed URLs as an API response (the same as for photo attachments).

TODO:

- [ ] Below OdkCentralAsync.listFormAttachments, add OdkCentralAsync.getFormAttachmentUrls. This method will be similar to getSubmissionAttachmentUrls, but for uploaded and synced **form attachments**.
- [ ] Add a snippet like this to the `/central/upload-form-media` endpoint, to ensure the attachment blobs are synced db --> S3:
    ```python
        # Ensure S3 form attachments are uploaded to S3
        async with OdkCentral(
            url=project.odk_credentials.odk_central_url,
            user=project.odk_credentials.odk_central_user,
            passwd=project.odk_credentials.odk_central_password,
        ) as odk_central:
            await odk_central.s3_sync()
    ```
- [ ] Then when the user calls the `/central/get-form-media` endpoint (when a project is loaded), we can simply call the OdkCentralAysnc.getFormAttachmentUrls method to get a list of URLs. The images will be downloaded by the frontend and embedded via CSS for usage.

## Review Guide

- Use this test form to create a project: [mdp_xlsform.xlsx](https://github.com/user-attachments/files/19623877/mdp_xlsform.xlsx)
- Upload this image for the form via API: mdp
- The endpoints is `/central/upload-form-media`.
- Then continue to test using the `/central/get-form-media` endpoint.

### DIFF
--- a/src/backend/app/central/central_crud.py
+++ b/src/backend/app/central/central_crud.py
@@ -1004,3 +1004,4 @@ async def get_form_media(
             xform_id,
         )
         return form_attachments
+    # TODO continue from here
--- a/src/backend/app/central/central_crud.py
+++ b/src/backend/app/central/central_crud.py
@@ -990,3 +990,17 @@ async def upload_form_media(
                 form_id=xform_id,
                 attachments=attachment_filepaths,
             )
+
+
+async def get_form_media(
+    xform_id: str,
+    project_odk_id: int,
+    odk_credentials: central_schemas.ODKCentralDecrypted,
+):
+    """Get a list of form media attachments."""
+    async with central_deps.get_async_odk_form(odk_credentials) as async_odk_form:
+        form_attachments = await async_odk_form.listFormAttachments(
+            project_odk_id,
+            xform_id,
+        )
+        return form_attachments
--- a/src/backend/app/central/central_routes.py
+++ b/src/backend/app/central/central_routes.py
@@ -137,6 +137,8 @@ async def upload_form_media(
             project_odk_creds,
             media_attachments,
         )
+        # FIXME this doesn't publish the form after??
+        # FIXME is the form still in draft?
 
         return Response(status_code=HTTPStatus.OK)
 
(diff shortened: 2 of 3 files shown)


spwoodcock:
Merging this early with a hardcoded value, so @DanielJDufour can work with it.

To be continued in the coming days in a follow up PR @Anuj-Gupta4 

The endpoint `/central/get-form-media?project_id=2` returns a response like this:

```json
{
  "mdp.png": "https://s3.mdp.fmtm.hotosm.org/fmtm-data/frontend/mdp.png"
}
```

- In our call I said I would be a question to image mapping, but ODK doesn't return that info.
- Instead I think we simply use the filename to decide where it goes (see snippet from latest form):

  image

Issue #2311, opened 2025-03-31
Introduce different project type based on visibility


manjitapandey:
**Is your feature request related to a problem? Please describe.**
Field mapping have different types and different data security has to be enabled for different projects. Some project may have sensitive data and some can have totally private project necessity.

**Describe the solution you'd like**
Based on discussion earlier on discussion linked below, introduction to different project type based on visibility should be added. 
project creator can choose the type of project during project creation. 
For public projects all data collection, visualisation and download should be available for any users. 
Private: restricted contribution, restricted viewing
Invite only: restricted contribution, anyone view
Sensitive: anyone contribute, restricted viewing

**Describe alternatives you've considered**
This needs internal discussion and further refinement of requirement

**Additional context**
From discussion https://github.com/hotosm/fmtm/discussions/1083#discussioncomment-11447239

spwoodcock:
We actually need this in the MdP project 😃

By default all projects in MdP are `PRIVATE`, meaning only the users that have been invited to work on it are able to view the project and contribute to it.

@Sujanadh could you please look into this 🙏

We need to verify that for private projects:

- They are hidden from the project summaries API (so only users who have access can easily see it).
- Any endpoint with `mapper` level role first checks if the project is private. If yes, it blocks the user from editing the project.

I will let look into the potential for creating 'invite links', with a token where the user can open and easily be added to a project.

Sujanadh:
Okay, I will look into this issue.

manjitapandey:
Backend implementation is completed. Frontend needs to be implemented accordingly. During project creation, users should be provided options to choose the project visibility type.
By default, project's visibility is public but can be changed to other types via API documentation.

manjitapandey:
To do on Frontend: 

- [ ] Add a section during project creation to let project creator choose project's visibility type.
- [ ] Add a tag of project visibility type next to project title on project details page to make user clear. 
- [ ] Refactor visualisation of buttons accordingly and proper error message for not permitted sections

manjitapandey:
When I try to reach project details page of private project, it doesn't show proper error message. 

![Image](https://github.com/user-attachments/assets/01e6c4a4-0baa-4059-8933-5dc91710c60d)

spwoodcock:
Each user should only be seeing the projects they have access to on the home page.

Are you accessing the project page directly via URL, instead of clicking through the home page?

manjitapandey:
> are you accessing the project page directly via URL, instead of clicking through the home page?

[...]

PR #2353, opened 2025-04-09
Implement role wise access to project summaries

Sujanadh:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Issue: #2311
- osm login issue in mapper page

## Describe this PR

- Create a `public_view` dependency to allow public users without login to view the project summaries. The main goal of introducing it is to get the user information if they are logged in so that we can check if the user has access to the private projects or not. If `login_required` is used here user won't be able to view project cards without logging in.

- Check if the user is `superadmin`; if he is `superadmin`, he can view every project.

- Check if the user is `org_admin`; then he can view all the projects within the organisation, whether the projects are private or public, along with the other public projects from other organisations.

- Check if the user is a mapper and he has access to private projects; otherwise, he will only view public projects.

- Update osm callback endpoint in mapper page `callback/osm -> callback/osm/mapper`
## TODO

- Implement other project visibility categories such as sensitive, invite only.
- Implement access likewise, to view, to contribute and so on.

## Screenshots

N/A

## Review Guide

### DIFF
--- a/src/backend/app/auth/auth_deps.py
+++ b/src/backend/app/auth/auth_deps.py
@@ -268,6 +268,23 @@ async def expire_cookies(response: Response, cookie_names: list[str]) -> Respons
 
 
 ### Endpoint Dependencies ###
+async def public_view(
+    request: Request,
+    access_token: Annotated[Optional[str], Header()] = None,
+) -> Optional[AuthUser]:
+    """Optional login dependency: returns AuthUser if authenticated, else None."""
+    if settings.DEBUG:
+        return AuthUser(sub="osm|1", username="localadmin", role=UserRole.ADMIN)
+    
+    extracted_token = access_token or get_cookie_value(request, settings.cookie_name)
+
+    if not extracted_token:
+        return None
+
+    try:
+        return await _authenticate_user(extracted_token)
+    except Exception:
+        return None  # Fail silently to allow access to public content
 
 
 async def login_required(
--- a/src/backend/app/auth/auth_deps.py
+++ b/src/backend/app/auth/auth_deps.py
@@ -275,7 +275,7 @@ async def public_view(
     """Optional login dependency: returns AuthUser if authenticated, else None."""
     if settings.DEBUG:
         return AuthUser(sub="osm|1", username="localadmin", role=UserRole.ADMIN)
-    
+
     extracted_token = access_token or get_cookie_value(request, settings.cookie_name)
 
     if not extracted_token:
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -1481,6 +1481,7 @@ class DbProject(BaseModel):
         db: Connection,
         skip: Optional[int] = None,
         limit: Optional[int] = None,
+        current_user: Optional[str] = None,
         org_id: Optional[int] = None,
         user_sub: Optional[str] = None,
         hashtags: Optional[list[str]] = None,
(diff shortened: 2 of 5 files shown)


src/backend/app/auth/auth_deps.py
```diff
@@ -268,6 +268,23 @@ async def expire_cookies(response: Response, cookie_names: list[str]) -> Respons
 
 
 ### Endpoint Dependencies ###
+async def public_view(
```
spwoodcock, line 271: I like this idea, but do we definitely need this extra access level?

If we use `login_required` then the endpoint is truly public and we don't care about auth.

If we use `mapper` role, shouldn't we implement the logic to check for public/private here?

Maybe I haven't thought this through properly

spwoodcock (changes requested):

src/backend/app/auth/auth_deps.py
Sujanadh: So implementing login required won't allow users to view the projects without logging in to the FMTM; we need tokens set, the same for the mapper role. So, I would say `login_required` is only public if they have signed up in FMTM.

spwoodcock: Ah of course, my bad there - we do need the new dependency as you have created 👍 
- `login_required` is still needed to add authentication to endpoints that have no associated project, e.g. the helper endpoints.
- I would propose renaming `public_view` to `project_visibility`. The code would read like `Depends(project_visibility)`, which I think is quite intuitive.

spwoodcock: Ah sorry again, I actually looked through your code more thoroughly. I understand what you are trying to do.

We need a dependency to get the currently user, but also fallback to allowing no login!

spwoodcock (approved):
This is a really nice change - apologies for the rushed review - you documented the change really well in the description😅

src/backend/app/db/models.py
```diff
@@ -1509,47 +1510,95 @@ async def all(
...
+
+        if current_user:
+            db_user = await DbUser.one(db, current_user.sub)
+            is_superadmin = db_user.role == UserRole.ADMIN
+            managed_org_ids = (
+                db_user.orgs_managed if hasattr(db_user, "orgs_managed") else []
+            )
+            params["current_user_sub"] = current_user.sub
+
+        if current_user:
```
spwoodcock, line 1524: No need for second `if current_user` here

spwoodcock (commented):
Happy to merge this as it is. My only comment would be, as the `.all` method has grown so huge, we could probably make functions to make it easier to understand, e.g.

```python
def build_visibility_filter(is_superadmin, managed_org_ids, current_user_sub):
    if is_superadmin:
        return None, {}
    elif managed_org_ids:
        return (
            "(p.visibility = 'PUBLIC' OR p.organisation_id = ANY(%(managed_org_ids)s))",
            {"managed_org_ids": managed_org_ids}
        )
    else:
        return (
            "(p.visibility = 'PUBLIC' OR (p.visibility = 'PRIVATE' AND ur.user_sub = %(current_user_sub)s))",
            {"current_user_sub": current_user_sub}
        )

visibility_filter, visibility_params = build_visibility_filter(...)
if visibility_filter:
    filters.append(visibility_filter)
params.update(visibility_params)
```

spwoodcock:
Happy to merge this as it is. My only comment would be, as the `.all` method has grown so huge, we could probably make functions to make it easier to understand, e.g.

```python
def build_visibility_filter(is_superadmin, managed_org_ids, current_user_sub):
    if is_superadmin:
        return None, {}
    elif managed_org_ids:
        return (
            "(p.visibility = 'PUBLIC' OR p.organisation_id = ANY(%(managed_org_ids)s))",
            {"managed_org_ids": managed_org_ids}
        )
    else:
        return (
            "(p.visibility = 'PUBLIC' OR (p.visibility = 'PRIVATE' AND ur.user_sub = %(current_user_sub)s))",
            {"current_user_sub": current_user_sub}
        )

visibility_filter, visibility_params = build_visibility_filter(...)
if visibility_filter:
    filters.append(visibility_filter)
params.update(visibility_params)
```

Sujanadh:
## Updates:
- Added a `needs_user_roles_join` parameter to control when the `user_roles` join should be included
- Created separate function to get the user access level `_get_user_access_level`
- Refactored the visibility filter logic into separate function `_build_visibility_filter`
- Separated minimal and full query to separate functions :-  `_minimal_select_query` , `_full_select_query`

spwoodcock (approved):
This is really nice - great work 👏

PR #2365, opened 2025-04-10
Image selection override

### LINKED ISSUE #2075: Ensure select icon widget appearance question type works in web-forms

- The client asked for some unique question types.
- They need to be simple and intuitive: ideally icon based.
- ODK has support for this already in Collect: https://docs.getodk.org/form-question-types/#select-widget-with-no-buttons-appearance

- We need to make sure this question type works in web-forms.
- If it's not developed already:
  - Contribute to ODK
  - OR make a custom solution to just get this working quickly (custom rendering for the specific question node in the XML)

DanielJDufour:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Example: Fixes #2075

## Describe this PR

Here's how it works:
- fetch images attached to a form and "caches" them as blob urls (so repeated openings of the form won't trigger external fetches)
- injects form xml and the "dictionary" of forms into the iframe via url params
- inside iframe, parse formXML and create a stylesheet that adds a background button to a select option based on the image dictionary

this is more flexible than hardcoding, but is a bit hacky, so I'm not sure how well it will scale to all use cases, especially when people have forms with multiple select questions with a columns-packed appearance. I know ODK team is hard at work on a solution, so I'm thinking we'll get to migrate to a better solution soon :-)

## Review Guide

Submit a web form :-)


DanielJDufour:
I should also mention that I replaced `bind:this={iframeRef}` with `use:handleIframe` because although the successfully binds the iframeRef, what we really want is to run a callback function every time a new iframe mounts.  The $effect magic of svelte doesn't seem to apply to refs (only $state and $derived and maybe a few other runes).  This new approach makes sure that the handleIframe is called everytime there is a new iframe.

spwoodcock (approved):

spwoodcock:
Looks good 👍

I will test thoroughly tomorrow.

Based on our slack chat, it should be noted that its not as simple as getting the image node from the XML, as translations are involved. The image field must be replicated for each translation available, so it is rendered when the locale is selected. Hence why we must iterate through the translations here

spwoodcock:
Testing this works nicely during local dev:

image

- Using the newly updated get-form-media endpoint to actually get URLs from ODK - it works well.
- I made a workaround during local development to avoid using pre-signed URLs.

The only problem is the image isn't bound to the box via the CSS.
The mdp logo is far too large and is only partially displayed here.
We can't control the sizing of the images a user might upload, so can we handle this via CSS @DanielJDufour ?

DanielJDufour:
@spwoodcock , yes, we should be able to handle this in CSS maybe `background-size`?
https://www.w3schools.com/cssref/css3_pr_background-size.php

DanielJDufour:
I double checked and when I tested locally it seems that the image is resizing correctly.  I'm thinking it might be a browser support issue, so will have to dive in an assess fallback strategies.  I'll also try to test on more devices and browsers.

DanielJDufour:
I was able to replicate, will have fix up shortly. I think the issue was around aspect ratio.

Issue #2167, opened 2025-02-10
Add user role check to API key router, plus option to download submission data for a project


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- Recently we added API keys.
- We don't need granular API keys per project, but instead can have this based on user permissions.
- We use the existing auth checks for a user to determine if the user has access to the project, but this time via an API key.
- The user needs to be able to download submission data via API for projects they have access to.

**Describe the solution you'd like**

- [x] Update API login dependency to use check_access on roles for a given user.
  - Return response if they have access.
  - Provide error if they do not have access.
- [x] Add integrations router endpoint to download submission for a project.
  - This should wrap existing crud logic to keep things DRY.

manjitapandey:
@spwoodcock , I am not sure what should I check in this issue. 
Can you guide me so I can test and close this?

spwoodcock:
I think it's all good - as long as no existing functionality has broken then this works well 😄 

(the API keys are used in production to turn the buildings green, so I think this is working fine)

PR #2376, opened 2025-04-14
Add api-key option to all endpoints (respecting auth and user roles)

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes https://github.com/hotosm/fmtm/issues/2167

## Describe this PR

- Despite what the issue says, I solved this in a different way.
- Instead of a separate `integrations` router, I added the api key checking logic to `login_required`, meaning we can now use api keys for every endpoint in FMTM.
- This makes thing much more flexible and opens integration with other apps significantly.
- I refactored the auth logic too - removed `mapper_login_required` dependency and replaced with `public_endpoint` dep.
  - Now we have `login_required` OR `public_endpoint` as options. All endpoints using login_required, unless they are fully public and we can skip the auth check.
  - With `public_endpoint`, the logic that Sujan implemented still stands - if the user is logged in, then provide their details, else return the endpoint anyway.
- Also separated out some `auth_deps` logic into an `auth_logic` file for better separation.

## Review Guide

- Please set `DEBUG=False` for local dev, or test on dev server once merge.
- Test thoroughly to check I didn't miss something - auth is important 🙏 


Sujanadh (approved):

Sujanadh:
Good to go :+1:

PR #2378, opened 2025-04-14
Add user invites via URL

### LINKED ISSUE #2377: Invite URLs that (1) create user if not present (2) assign user a role for a project

- To fully implement **private** projects, we need a good method to invite users.
- We have OSM messages, which is good if the user is already registered in FMTM.
- However often we need to start a new mapping campaign with users that have not logged into the platform yet.
- To facilitate this, we need invite URLs that both create the users in the db AND assign them to a project as `mapper` (or other role).

- Endpoint to provide invite URLs - field admin or high level.
- These could be sent via OSM message, whatsapp, or other means.
- The user clicks on the invite URL, it should include a token that authenticates them.
- The page they visit will:
  - Take them to a login page.
  - Allow them to select the login they wish to use.
  - Finish oauth login flow.
  - Once complete, redirect them back to the original URL they attempted to access.
  - An API is called including the original token from the URL:
    - The token is provided by the manager, with manager level permissions.
    - The entry will be made for the `user_roles`: including current user ID and the assigned role.
- The endpoint to assign the role should not allow creating a role of higher privilege than the inviter had originally.
  - I.e. a project manager cannot create a new super admin user. Only field admin, validator, or mapper levels.
  - A project manager should also probably not be allowed to create a new project manager, as this could be abused (the invited user could modify the URL and grant themselves project manager).
  - How do we stop the user abusing the URL though, and granting themselves as field admin, when the URL specified mapper permission?
    - Either we create the user_role entry in advance with a dummy user, then rename the user? Not ideal.
    - Or we can just obfuscate the role being provided in the URL by encrypting the data or base64 encoding (most users won't know how to tamper with this).

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Related to #2377

## Describe this PR

- Add a new db table `user_invites`.
  - Records when a manager adds a user to a project and provides an invite URL to them.
  - Optionally OSM message or email the user.
- Adds a router for creating and consuming user invite tokens.
- Token is included in frontend URL:
  - Saved to sessionStorage
  - User prompted to log in
  - After log in and user created in db, the role from the invite is set for the project/user

@Anuj-Gupta4 it would be awesome if you get a chance to check this over and push it forward tomorrow 🙏 
Once the backend is fully in place, @NSUWAL123 we also need a small frontend page to do the token handling via sessionStorage, login redirect, then calling the invite API, as described above 😄 

## Review Guide

- Create private project
- Ensure standard user can't access project
- Create token as project manager
- Consume token as user without FMTM account
- Create FMTM account as new user
- Receive OSM message if set
- Be granted permission to access the project

### DIFF
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -81,7 +81,12 @@ if TYPE_CHECKING:
         ProjectUpdate,
     )
     from app.tasks.task_schemas import TaskEventIn
-    from app.users.user_schemas import UserIn, UserUpdate
+    from app.users.user_schemas import (
+        UserIn,
+        UserInviteIn,
+        UserInviteUpdate,
+        UserUpdate,
+    )
 
 
 def dump_and_check_model(db_model: BaseModel):
@@ -364,7 +369,7 @@ class DbUser(BaseModel):
     async def update(
         cls, db: Connection, user_sub: str, user_update: "UserUpdate"
     ) -> Self:
-        """Update the role of a specific user."""
+        """Update a specific user record."""
         model_dump = dump_and_check_model(user_update)
         placeholders = [f"{key} = %({key})s" for key in model_dump.keys()]
         sql = f"""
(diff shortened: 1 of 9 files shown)


Anuj-Gupta4 (approved):

spwoodcock (commented):
Looks good 👍 

Minor change for timestamp usage.

Regarding the expiry, I think the approach you chose is good.

I was thinking if we should delete the entry (but this isn't good if the admin wants to check an issue / see the log of the user accepting the invite - they can be cleaned up via a script after).

I was also thinking about removing the `token` when expired, but also see no point doing that now.

PR #2408, opened 2025-04-21
Improve performance of project status materialized view + other updates

### LINKED ISSUE #1630: Project export (archive) endpoint

- Currently the migration path of data from one instance into another is difficult.
- It would be great to have a standardised way to package up a project in a zip for storage or later import.

- Add an endpoint that downloads all data required to recreate a project from scratch.
- It should output a zip.
- The zip should contain `.json` files that could be posted to endpoints for to recreate the project creation flow.
- We would also need to store S3 data (data extract) and ODK data (submissions, entities).

**Additional context**

First attempt at this was made here: https://github.com/hotosm/fmtm/pull/1000

### LINKED ISSUE #1631: Project import endpoint

- Following on from https://github.com/hotosm/fmtm/issues/1630
- We need a way to get data back into FMTM from an archive format.

- An import endpoint that accepts a zip file.
- The zip is unpacked and relevant endpoint logic called to re-create the project as it was before.

### LINKED ISSUE #2407: Project stats materialized view has major performance issues
**Describe the bug**

- On starting the API, the project stats materialized view is created and/or refreshed.
- This works fine on local dev without much data.
- I just added a script to load all prod data into local dev (we don't have lots, but this is a good test).
- The create / refresh of the materalized view takes about an hour.
- Assuming there is some unoptimised indexing somewhere - running an `EXPLAIN ANALYZE` now to assess.
- When we push this through the prod, it's pretty likely we will have server downtime for the 1hr as the view is created, unless this is fixed.

**To Reproduce**

1. just test load-prod-data
2. docker compose logs api
3. wait... and check until the server starts

**Expected behavior**

- The MATERIALIZED VIEW refresh should be pretty quick.
- It also runs on a schedule every 10 mins, so we want to reduce server load.
- Analyse the query and improve perf.
- High priority, as this would break prod temporarily as mentioned.

spwoodcock:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #2407
Fixes #1630
Fixes #1631

### Loading prod data

- Added config for loading prod data into local development.
- The project import / export functionality in #1630 and #1631 serves two purposes:
  - Easier testing of prod data.
  - Export from one FieldTM instance to another.
- The first use case is covered by this PR. The second isn't high priority, unless we meet a user that actually needs this functionality... as it's not a simple one to implement (the main blocker being easy import / export of ODK data by project - they only support bulk export/import for now).

### Improve performance of project stats

- Instead of doing raw joins and aggregating afterward, we do the heavy aggregation in CTEs, and join only 1 row per project from each.

### Adding load testing config

- Added `oha` or `k6` - tbc

## Review Guide

### DIFF
--- a/src/backend/scheduler/project_stats.py
+++ b/src/backend/scheduler/project_stats.py
@@ -13,7 +13,7 @@ from app.db.postgis_utils import timestamp
 DB_URL = settings.FMTM_DB_URL.unicode_string()
 
 # Create materialized view to store project stats for faster query:
-# Here we retrieve:
+# Using a CTE to get lastest_task_events first, we then calculate:
 #  - Total number of contributors (task / entity events)
 #  - Total number of complete submissions
 #  - Total number of tasks mapped
@@ -25,10 +25,12 @@ DB_URL = settings.FMTM_DB_URL.unicode_string()
 #  - No big joins, no GROUP BY, no row explosions
 #  - Each subquery uses narrow filters (WHERE project_id = p.id)
 # NOTE as FieldTM scales and projects increase, we could swap the final
-# NOTE    WHERE p.id IN (SELECT p FROM projects);
+# NOTE    FROM projects p;
+# NOTE to
+# NOTE    WHERE p.id in (SELECT id FROM projects COUNT xxx);
 # NOTE to take a narrower range of IDs and do schedule partial updates to
-# NOTE avoid refreshing all in one go.
-# TODO performance of this takes ~15ms. Consider moving back to models.py logic.
+# NOTE avoid refreshing all in one go. Performance of this takes ~15ms 36 projects.
+# TODO We could consider moving back to models.py logic to generate via the API.
 CREATE_MATERIALIZED_VIEW_SQL = """
     CREATE MATERIALIZED VIEW IF NOT EXISTS mv_project_stats AS
     WITH latest_task_events AS (
@@ -69,7 +71,6 @@ CREATE_MATERIALIZED_VIEW_SQL = """
             WHERE project_id = p.id AND event = 'GOOD'
         ) AS tasks_validated
     FROM projects p;
-    -- where p.id in (SELECT id FROM projects);
 """
 
 CREATE_UNIQUE_INDEX_SQL = """


src/backend/scheduler/project_stats.py
```diff
@@ -8,45 +8,83 @@
...
+            SELECT COUNT(DISTINCT t.id)
+            FROM tasks t
+            WHERE t.project_id = p.id
+        ) AS total_tasks
     FROM projects p
-    LEFT JOIN task_events ev ON p.id = ev.project_id
-    LEFT JOIN odk_entities et ON p.id = et.project_id
-    LEFT JOIN latest_task_events lte ON p.id = lte.project_id
-    GROUP BY p.id;
+    WHERE p.id IN (SELECT id FROM projects);
```
Sujanadh, line 87: What if the project does not have any task events?

spwoodcock: Yes the final logic is redundant 👍
I put it there as an example so we could do a batching approach using project ids in future.

And good point about empty task events - probably need to tweak it!

spwoodcock:
> I think the approach to create a materialised view for the project stats repeatedly scans `task_events` for each project and for each stat, which might be non-performant if there are a large number of projects? I think the CTE approach well considers the performance. Creating CTE will allow us to scan each table only once, which avoids multiple scanning.
> 
> ```
> For example:
> If
> 100 projects
> 10,000 task events
> 
> With these changes
> 
> The `task_events` table might be scanned up to 300 times (3 stats × 100 projects)
> 
> With the CTE joins approach:
> 
> The `task_events` table is scanned just once, and the results are joined appropriately.
> ```
> 
> Latest PostgreSQL optimizers are very efficient with joins.

This is a good point for sure.

The approach is predictable and scales linearly with the number of projects we have. At 15ms for our current prod data, I think we have plenty of overhead. Its using the indexes nicely.

The best approach would probably be a hybrid, where we precomputed the latest_task_events via CTE first, then use the scalar sub queries scoped per project for the counts.

What do you think?

spwoodcock:
The problem with the previous implementation was chained joins

```sql
FROM projects p
LEFT JOIN task_events ev ON p.id = ev.project_id
LEFT JOIN odk_entities et ON p.id = et.project_id
LEFT JOIN tasks t ON p.id = t.project_id
```

Example:
- 1 project
- 50 tasks
- 100 task_events
- 50 odk_entities

If we don't isolate or dedup the rows, we get a row explosion where each row is joined to every other

50*100*50 = 250,000 rows for a single project

Sujanadh:
> Did you see my comment here: [#2407 (comment)](https://github.com/hotosm/fmtm/issues/2407#issuecomment-2818837493) ?
> 
> I was running the previous implementation on prod data and it was taking 30min - 1hr to compute. Now it takes 15ms 😅

Ahhh, I just noticed. That makes sense. We are only joining tables with `project_id`. When joining only on `project_id`, we get every possible combination of a project's task events with its ODK entities.

Sujanadh:
> > I think the approach to create a materialised view for the project stats repeatedly scans `task_events` for each project and for each stat, which might be non-performant if there are a large number of projects? I think the CTE approach well considers the performance. Creating CTE will allow us to scan each table only once, which avoids multiple scanning.
> > ```
> > For example:
> > If
> > 100 projects
> > 10,000 task events
> > 
> > With these changes
> > 
> > The `task_events` table might be scanned up to 300 times (3 stats × 100 projects)
> > 
> > With the CTE joins approach:
> > 
> > The `task_events` table is scanned just once, and the results are joined appropriately.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Latest PostgreSQL optimizers are very efficient with joins.
> 
> This is a good point for sure.
> 
> The approach is predictable and scales linearly with the number of projects we have. At 15ms for our current prod data, I think we have plenty of overhead. Its using the indexes nicely.
> 
> The best approach would probably be a hybrid, where we precomputed the latest_task_events via CTE first, then use the scalar sub queries scoped per project for the counts.
> 
> What do you think?

I think yes, that would be a way to go. Something like this:
```sql
CREATE MATERIALIZED VIEW IF NOT EXISTS mv_project_stats AS
WITH latest_task_events AS (
    SELECT DISTINCT ON (ev.project_id, ev.task_id)
        ev.project_id,
        ev.task_id,
        ev.event_id,
        ev.event
    FROM task_events ev
    ORDER BY ev.project_id, ev.task_id, ev.created_at DESC
)
SELECT
    p.id AS project_id,
    (
        SELECT COUNT(DISTINCT user_sub)
        FROM task_events
        WHERE project_id = p.id
    ) AS num_contributors,
    (
        SELECT COUNT(DISTINCT entity_id)
        FROM odk_entities
        WHERE project_id = p.id
        AND status = 'SURVEY_SUBMITTED'
    ) AS total_submissions,
    (
        SELECT COUNT(*)
        FROM latest_task_events
        WHERE project_id = p.id AND event = 'FINISH'
    ) AS tasks_mapped,
    (
        SELECT COUNT(*)
        FROM latest_task_events
        WHERE project_id = p.id AND event = 'BAD'
    ) AS tasks_bad,
    (
        SELECT COUNT(*)
        FROM latest_task_events
        WHERE project_id = p.id AND event = 'GOOD'
    ) AS tasks_validated,
FROM projects p;
```

And I think we don't need `total_tasks` as it is already computed in the project table?

spwoodcock:
Ah nice, yes we can remove total tasks!

The implementation looks good!

I'll be starting work in 1-2hrs (shifted times today), but feel free to modify and test if you want.

I can test on prod data too 👍

Sujanadh:
I have updated the SQL to use CTE for the latest task events.

Issue #1291, opened 2024-02-27
Make generate-project-files idempotent


spwoodcock:
**Is your feature request related to a problem? Please describe.**

Task creation may fail due to various reasons not related to incorrect input (server overload, db connection limit, etc).

*Describe the solution you'd like**

The project files generation should be idempotent. 

This means that there are checks to see which tasks have an odk token, form in odk central and uploaded geojson media. 

It would increase the time of the endpoint due to additional calls, but worth it to allow multiple sequential executions, i.e. resume.

**Describe alternatives you've considered**

An additional endpoint to resume, but it would be nice to have the same one.

spwoodcock:
To facilitate this I think we should update the project creation flow.

**Current**: org admin creates entire project.
**Proposal**:
- Org admin create a project stub with minimal details, assign a project admin
  - Requires a new endpoint `/project/stub/` (or perhaps reuse `/project` POST endpoint.
- Project admin fills out details and submits. This calls the project update endpoint, which is by default idempotent.
  - The project update will fill out the remaining fields.
  - For this to work, we probably need a flag in the db to determine when a project is full filled out, or only a stub.

Advantages of this approach:
- Idemponent: project creation can be repeated without issues.
- Reduce code duplication: the editing workflow can be the same as the project creation.
- Permissions: better to pass main project creation responsibility to project admin.
- Ordering: project is created prior to data extract, meaning we can include **project** id in the data extract file name.
  - This allows for data extract tracking by HOT in production.

spwoodcock:
Note: we have ProjectStatus enum with `DRAFT=2`, so should probably create the stub with DRAFT status.

Once complete, change to `PUBLISHED`, enum id = 1

Anuj-Gupta4:
The post project API can already be called with only a few parameters:

- Name
- Description
- Other instructions
- Project AOI
- ODK credentials

We could let organization admin input these as a draft project. It's probably ok to ask for ODK credentials here because sometimes, we use organization's ODK credentials. Maybe we can have project admin later update them if they would prefer to use their own?

There are separate APIs to do everything else. We can update the access to be project manager only. Once we make sure tasks are successfully created and related files are generated, we can update the project status to be published.

spwoodcock:
Yeah I think the project manager should be deciding on / adding ODK creds 👍 

I'm thinking it would be ideal if the org manager only needs to specify the project name as mandatory, but the project is created as `draft`.

[...]

PR #2411, opened 2025-04-22
Implement stub project creation and fix project deletion error

Anuj-Gupta4:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related to: #1291

## Describe this PR

- Added an endpoint for organization admins to create a stub project.
- Updated create project endpoint based on stub project.
- Updated project related endpoints for project manager's use.
- Rename data extract filename to have project ID in it.
- Added deletion of user invites associated with a project during project deletion.


src/backend/app/projects/project_routes.py
```diff
@@ -1288,6 +1263,50 @@ async def delete_project(
     return Response(status_code=HTTPStatus.NO_CONTENT)
 
 
+@router.post("", response_model=project_schemas.ProjectOut)
```
spwoodcock, line 1266: Should this possibly be a PATCH request?
As its updating an existing resource.

We could remove the existing PATCH endpoint is there is one

src/backend/tests/test_projects_routes.py
```diff
@@ -213,14 +213,23 @@ async def test_invalid_geojson_types(client, organisation, project_data, geojson
...
+    stub_project_data["outline"]["crs"] = crs
     with pytest.raises(HTTPException) as exc_info:
-        await check_crs(project_data["outline"])
+        await check_crs(stub_project_data["outline"])
     assert exc_info.value.status_code == 400
 
 
+async def create_project(client, organisation_id, stub_project_data):
+    """Create a new project."""
+    response = await client.post(
```
spwoodcock, line 226: Isn't this duplicating the other test a bit?

It would be great to have a combined test that does the full flow: POST sub PATCH remainder of project metadata

spwoodcock (changes requested):

spwoodcock:
Nice to see this finally happening!

I think it will be a big improvement to manager UX

spwoodcock:
This is great & we can merge it soon, but its embarking on a big change in project creation flow (involving frontend too).

I think we should merge it after the next release is done in a week or so 👍

spwoodcock:
I think this could probably be merged in soon.

Just checking - in order to partially migrate to this, we could:
- From the frontend, call the stub create endpoint, then call the patch one after the other.
- This should create a project in the database that is indistinguishable from what we have now.
- To do this, we need a small frontend refactor - currently there is lot of frontend work ongoing an this has been low priority.

In the long run, we could then use the stub project creation as part of an alternative workflow.

Anuj-Gupta4:
Yes, the post `create_stub_project` and patch `create_project` along with other already existing supporting project APIs will create a project that is indistinguishable from the projects that we already have.

PR #2422, opened 2025-04-25
fixing issue 2403

### LINKED ISSUE #2403: Use of URL.canParse fails on browsers < 2022 (polyfill required)
**Describe the bug**

- Testing on an iOS v16 device, using Safari from 2022.
- It's reasonable to expect many end users in developing countries will be using slightly outdated devices, particularly iOS (it varies country by country).
- I get an error `URL.canParse` is not a valid function. Support didn't come in until 2023 some time: https://caniuse.com/mdn-api_url_canparse_static
- We don't use the method directly, but it seems to be related to loading the blob XML for the XForm.

**To Reproduce**

1. Go to project
2. Click on point
3. 'Collect Data'
4. Opens web-forms, where we see the error.

Device: iPhone 8 plus
OS: iOS 16.7.2
Browser: Safari 16.6

**Expected behavior**

- This is the only thing that I found broken on the old device.
- It would be great to have a polyfill if possible.

Since canParse uses the same algorithm as the URL() constructor, it could be as simple as:

```js
URL.prototype.canParse = function(string) {
  try {
    new URL(string);
    return true;
  } catch (err) {
    return false;
  }
}
```
another example polyfill: https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/web.url.can-parse.js

**Screenshots**

Image

Image

**Smartphone (please complete the following information):**

**Additional context**

DanielJDufour:

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Example: Fixes #2403 

## Describe this PR

Added code that adds canParse to URL object.  This is contained within the iframe.  [popover] is also not supported in older browsers, so basically hide it.  Unfortunately odk web-forms doesn't seem to have any other attributes it updates when error popover is updated, so css rules rely on pesudo-classes.  What this means in practice is that certain errors might not appear; however, in my testing the relevant errors (like not filling in a required field) still appear correctly.

URL.prototype.canParse didn't work for me, so that's why I use URL.canParse

adding support directly into xforms-engine or web-forms.  However, I want to minimize merge conflicts with upstream.

## Review Guide

### DIFF
--- a/src/mapper/static/web-forms.html
+++ b/src/mapper/static/web-forms.html
@@ -46,14 +46,14 @@
 	</head>
 	<body style="height: 100vh; overflow-y: scroll; position: relative">
 		<script type="module">
-			URL.canParse = function(string) {
+			URL.canParse = function (string) {
 				try {
 					new URL(string);
 					return true;
 				} catch (err) {
 					return false;
 				}
-			}
+			};
 
 			const odkWebFormUrl = new URL(window.location.href).searchParams.get('odkWebFormUrl');
 			const OdkWebForm = (await import(odkWebFormUrl)).default;
--- a/src/mapper/static/web-forms.html
+++ b/src/mapper/static/web-forms.html
@@ -18,10 +18,43 @@
 			.columns-pack .label-text {
 				display: none;
 			}
+
+			/* css polyfilling for old browsers */
+			[popover] {
+				display: none;
+			}
+			[popover]:popover-open {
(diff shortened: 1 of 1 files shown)


spwoodcock:
nice: lgtm

Issue #2395, opened 2025-04-16
If a user sets a task area to 'complete' there is no way to revert this


spwoodcock:
**Describe the bug**

- We took the decision before to prevent users resetting tasks (in mapper frontend) after marked as complete.
- We have checks in place to see if all the features are mapped yet, before marking as complete.
- However, we didn't consider the case for 'new geometry' mapping only, as with MdP.
- If not existing geoms are present, and the user marks the area as complete accidentally, they can't revert and continue mapping new features.

**To Reproduce**

1. Project with no existing features in a task area
2. Start mapping --> mark complete
3. Clicking task area shows no popup

**Expected behavior**

- We need to reconsider how this is handled.
- Resetting the task is possible via management frontend, but we need a way to do this in the mapper frontend probably?
- Let's consider and discuss.

**Update: possible easy solution**

- I think an easy solution to this is to prevent locking the task in the first place when there is only one task in a project.
- Can we disable the action buttons like 'Start Mapping', 'Mark Complete', etc when there is a single task in a project?
- This circumvents the problem entirely, without having to change the logic we implemented for standard mapping projects (existing data).

manjitapandey:
> Resetting the task is possible via management frontend, but we need a way to do this in the mapper frontend probably?

We already have resetting via management frontend for either super admin or by the same mapper who marked the task as mapped. I think we can do the same on mapper frontend (only allow resetting for user who marked it as completed mapping). In case, if someone marks the task as mapped without logging in, they can contact their coordinator to reset the task. 

> Update: possible easy solution

I don't think its a good idea to only consider single task because we faced the same cases in Tokha as well where there were multiple tasks. Rather providing access to reset the task by limited users (mapper who marked the task as mapped and admin) can be good solution. wdyt?
@spwoodcock , @ramyaragupathy

NSUWAL123:
> We already have resetting via management frontend for either super admin or by the same mapper who marked the task as mapped. I think we can do the same on mapper frontend (only allow resetting for user who marked it as completed mapping). In case, if someone marks the task as mapped without logging in, they can contact their coordinator to reset the task.

Yeah, this should be an ideal solution.

Also, when we implement the concept of teams, we can check if the user is present on the assigned team, and then only allow those specific team members to update the task.

spwoodcock:
Sounds good to me! I think you covered all the bases between you 😄 

Anything to benefit 'single task only' projects can be addressed on another issue 👍 (as some improvements could be made for them)

[...]

PR #2425, opened 2025-04-25
Display confirmation dialog on task marked as completed

NSUWAL123:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #2395

## Describe this PR
- Display a confirmation dialog if the user tends to mark the task comprising submission counts less than feature counts

## Screenshots
image


NSUWAL123:
Since new geoms are drawn after selecting the task area and as it is posted to entities api, it behaves like a primary geometry containing the task id when entities/statuses api is called. 
So both cases will be handled by the confirmation dialog!

spwoodcock:
> Since new geoms are drawn after selecting the task area and as it is posted to entities api, it behaves like a primary geometry containing the task id when entities/statuses api is called. So both cases will be handled by the confirmation dialog! image

That's a really good point & works well 😄 

But what about if the user has drawn no geometries yet, but accidentally locks and completes the task?

I only mention this, as that's exactly what happened on the MdP demo server 😅 
A user was playing with functionality, completed the task (with no warning), then nobody else could map (until I reset the task as an admin).

NSUWAL123:
Updated the logic: 
1. If the entity submission is incomplete, show a confirmation modal including `submission`/`feature count`
2. If the task has no entities, show a confirmation modal
3. Otherwise, just mark the task as complete (since it's submission count is not less than feature count)

spwoodcock:
Encountered two possible issues:

### 1

- Project with no features
- A user1 maps one feature
- user2 clicks to lock the task --> 'Complete the task' (no dialog is shown)
- user1 can no longer map as user2 accidentally locked the task
- Expected: The only way to have this work with the `submission/feature count` check is to force showing a dialog every time. In this scenario it should say something like: '10/10 features are mapped. Completing the task will prevent further submissions.'.

### 2

- The `submission/feature count` is only correct based on the current users mapping.
- Say there are 5 features to map.
- user1 maps 2 features
- then user2 maps2 features
- then user1 maps the final feature
- The final `submission/feature count` for both users would be:
  - user1: 3/5
  - user2: 2/5
- This means the project cannot be completed by either user until they refresh the page and update their `submission/feature count`.
- Expected: the features mapped count is synced via electric - this should probably be in another issue to track it👍

spwoodcock (changes requested):
see comment 😄

NSUWAL123:
Updated: 
Screenshot 2025-04-28 170552
Screenshot 2025-04-28 170608
Screenshot 2025-04-28 170623

spwoodcock:
Nice, this is good 👍 

Don't forget that every new string needs to use translations now though 😅 
Sorry for the hassle!

I did a small update to include this - took longer than anticipated!

### Extra info (for interest)

#### Re: including 'only'

- We could do an approach that translates 'only' separately, then injects in the value as {only}, e.g.

    ```
    "confirm_task_complete": "You have {only} mapped {totalMapped} features in the task area. Are you sure you wish to mark this task as complete?"
    "confirm_task_complete": "Você mapeou {apenas} recursos {totalMapped} na área de tarefas. Tem certeza de que deseja marcar essa tarefa como concluída?"
    ```

- This works fine for English and Portuguese, as we can assume the fixed position of the word 'only' in the sentence.
- But language is complicated: we can't foresee the sentence structure of certain languages.
- For example in some languages the 'only' part might appear elsewhere in the sentence ordering and not make sense when injected in.

For example:

    ```
    "confirm_task_complete": "You have {only} mapped {totalMapped} features in the task area. Are you sure you wish to mark this task as complete?"
    ```

in another language, the equivalent sentence structure might actually be something like

    ```
    "confirm_task_complete": "You have mapped {only} {totalMapped} features in the task area. Are you sure you wish to mark this task as complete?"
    "confirm_task_complete": "You have mapped {totalMapped} features {only} in the task area. Are you sure you wish to mark this task as complete?"
    ```

The only easy option I see for now is to omit the 'only' part of the text, or have two separate translation strings!
I added separate strings 😄 

#### Re: having bold text in translations

- This one wasn't easy!
- Paraglide (and most libs) can't support formatting included for part of the translation string.
- One option is to cut up the string into parts, but you get problems like I mentioned above around natural sentence structuring.
- The easiest solution I found for now was to hardcode unicode bold text inside the translation!
- I made a util function that converts standard strings to bold strings, using unicode.

spwoodcock:
image

Issue #2338, opened 2025-04-03
Optimise usage of entities store in mapper frontend


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- We have a fair bit of code in `entities.svelte.ts` and `dialog-entities-actions.svelte` that does the same thing.
- I also have a feeling that code like this:

```ts
entitiesStore.updateEntityStatus(projectData.id, {
	entity_id: selectedEntity?.entity_id,
	status: 2,
	// NOTE here we don't translate the field as English values are always saved as the Entity label
	label: `Task ${selectedEntity?.task_id} Feature ${selectedEntity?.osmid}`,
});
```

can probably be optimised to be more like this:

```ts
entitiesStore.updateSelectedEntityStatus(projectData.id, 2);
```

and then simply use `entity_id`,  `task_id` and `osm_id` from the selected entity in the store to generate the entity_id and label fields.

**Describe the solution you'd like**

- Clean up the code usage of entities store a bit.
- Perhaps the creation of the Map in this code can be done once in the store only? To avoid having to do it in other files too.

```js
	let entityMap = $derived(new Map(entitiesStore.entitiesStatusList.map((entity) => [entity.entity_id, entity])));
```

- Most logic for the entities should be in the store, with access from other files.
- There is also logic to refactor in `lib/form/wrapper.svelte` that used the entities Map.

spwoodcock:
Part of future work, related to offline workflow.

Issue #1987, opened 2024-12-12
Store electric data in PGLite DB first, with selective synn


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- Originally we had this.
- Electric changed to become more modular.
- We now just use the Shapes to get data, then save in frontend store.

**Describe the solution you'd like**

- Usage of the store for many things should be replaced by PGLite DB storage (task events, ODK entity statuses, new and bad features).
Using a local-first approach, then then allow electric to sync from local db to remote db.

spwoodcock:
https://www.npmjs.com/package/@electric-sql/pglite-sync

spwoodcock:
Moving this to MdP tasks.

The entity creation is a bit flaky for users.

I would propose we move:
- tasks
- task_events
- geometry_log
to PGLite local-first storage, then sync to electric.

The issue is that electric sometimes doesn't immediately sync the data to the user, so nothing is displayed.
Electric should be a sync later, not a data fetcher.
Data should be fetched from local PGLite first, then updated via electric.

PR #2427, opened 2025-04-26
Implement PGLite for local-first database in mapper frontend

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Related to https://github.com/hotosm/fmtm/issues/1987
Will probably also address #2338
Should also address #2457

## Describe this PR

- Init PGLite db on startup.
  - Attempts loading db first.
  - If fails, initialises a new db with the schema.
- Data loads from local db **first**.
- Data syncs to the local db when updated (live).
- More resilient to connectivity drops.
- Can work entirely offline with good caching.

    - [x] task_events
    - [x] geometrylog
    - [x] odk_entities

Note for the backend:
- As part of this PR I made the database initialisation script modular.
- SQL statements are imported using `\i './2-tables.sql'`.
- This allows reuse of part of the schema in PGLite in the frontend.
- We actually run the same init SQL for both & could add migrations in the frontend in theory too.


manjitapandey:
from #2475  conversation, what I understand this PR is for is: 
1. We would still be needing internet connection to create new entity. 
2. We would need internet for opening the form. 
3. We won't need internet during filling up form but will need to upload the form. At the same time we can trigger sync functionality. 
If my understanding is correct, we should also have save as draft option to deal with cases like they were filling up the form and if internet is unstable, the filled form won't get lost.

spwoodcock:
This PR addresses the **read path** only. This means it we load the task events, odk entities, and new/ bad geoms into a local db on first load.

If the user goes offline they can still read data in the app.

There is another issue open about solving the **write path**: saving new new features and submissions to a temp db table, then syncing when online.

spwoodcock:
Ready for review.

- This implements reading the data and storing in PGLite.
- Data is synced via Electric into PGLite.
- Entities or geometrylog records that are written are sent to both API and to local PGLite (+svelte store).

The future goal will be to remove the immediate API calls on the write path, meaning we write to PGLite first, then make the API calls in sequence when connectivity is available.

@NSUWAL123 could you please test this out tomorrow 🙏 
I managed to mess up the map layer selection a bit - it flickers on click, suggesting the layers are re-added on each click 😅 
It doesn't break things, but it would be great if we can find the problem!

NSUWAL123:
@spwoodcock Yeah, the flickering of the data extract happens, but in order to test it, I'm having some problems.
Tried on my local setup but had problems. 
At first, it works fine, but whenever I change & save the file, even if I add a small console, the entities are not being selected. Nothing is being console (might be some problem with the electric SQL?). 
For this, when I down & up the containers, the entity modal is shown and everything works fine, even the consoles. But when I change & save the file, the issue repeats

Also, on switching between projects, sometimes the layer of the opened project is not visible, so I have to hard reload that page.

I've built the containers multiple times, but the issue persists 🥲

spwoodcock:
I no!

I think its because onUnmount we unsubscribe from electric possibly?

But I think we subscribe onMount, so this should probably work.

I was hard refreshing to test, as doing a hot reload doesn't reflect the user experience. But perhaps there is a way to solve it

NSUWAL123:
After `docker compose down -v`, the entity selection issue is solved!
I'll have a look on the flickering issue now!

spwoodcock:
Some debug info from @NSUWAL123 on Slack 🙏 :

- Hard-coding the extent with the task boundary geojson in this PR, the map click doesn't triggered the flickering. Possibly because it looks like there was no call to the flatgeobuf URL for some reason (on S3).
- However, on the dev server (where this code is not merged yet), whenever there's click within the task boundary, the flatgeobuf URL is called but there is no flickering.

PR #2428, opened 2025-04-26
send correct entity status when submitting

### LINKED ISSUE #2412: Send the correct Entity status to API after web form completion
**Describe the bug**

- I made a workaround to the entity status not updating in ODK after form submit (assume this was the intention in web-forms, else not implemented yet:
https://github.com/hotosm/fmtm/blob/4e69d05751ec581b0e3626c79e7d8ffbdaae0f39/src/mapper/src/lib/components/forms/wrapper.svelte#L124-L130
- The problem is that we always send status `2`, meaning success.

We have options:

```python
    READY = 0
    OPENED_IN_ODK = 1
    SURVEY_SUBMITTED = 2
    NEW_GEOM = 3
    VALIDATED = 5
    MARKED_BAD = 6
```

- We need to also handle cases where `NEW_GEOM` and `MARKED_BAD` and send them to the API to get the correct feature colours on the map (and status for statistics).

**Expected behavior**

- If the building_exists or digitisation_problem questions exist, and there is an issue we should mark the status as `6`. Determine this from the form.
- If the geom is a new feature we injected in, then we should mark the status as `3` (we may need to determine this from store, as the form won't tell us if it's a new geom anymore for webforms).

Use this logic

```python
    calculate_status = (
        "if(${feature} != '', 2,"
        "if(${building_exists} = 'no', 6,"
        "if(${digitisation_correct} = 'no', 6,"
        "if(${new_feature} != '', 3, ${status}))" if use_odk_collect
        else "${status})))"
    )
```
translation:
- if feature included, set to 2
- if building_exists='no', then 6
- if digitisation_correct = 'no', then 6
- if new_feature included, set to 3
- else set to 6, as there was something wrong in the form

DanielJDufour:

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #2412

## Describe this PR

Basically sends an entity status of 6 if there's a problem, 3 if it's a new geom, and 2 otherwise.

I don't check if there's a missing `<feature>` tag, because the form wouldn't be open nor submitted if that is missing.

## Screenshots

I had originally checked if a `<feature>UUID here</feature>` existed, but that would always return true because we add that in earlier in the same function call.

## Review Guide

- Create a new feature and submit that (status sent should be 3)
- Create a submission where the building or feature doesn't exist (status sent should be 6)
- Create a submission where the feature isn't digitized correctly(status sent should be 6)
- Create a submission where everything is good (status sent should be 2)

### DIFF
--- a/src/mapper/src/lib/components/forms/wrapper.svelte
+++ b/src/mapper/src/lib/components/forms/wrapper.svelte
@@ -39,6 +39,7 @@
 	function handleSubmit(payload: any) {
 		(async () => {
 			if (!payload.detail) return;
+			if (!projectId) return;
 
 			const { instanceFile, attachments = [] } = await payload.detail[0].data[0];
 			let submission_xml = await instanceFile.text();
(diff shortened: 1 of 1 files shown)


spwoodcock (approved):
Nice and simple impl (thankfully) - love it 🎉

PR #2431, opened 2025-04-27
Feat/form language warning

### LINKED ISSUE #2358: The language of the web form should reflect the language set in the users browser
- I have my language set to 'en' in the browser.
- The web-form when opened is in 'pt-BR'.
- This is most likely because the `default_language` in the form is being respected, which is good.
- But I would propose we change this behaviour if possible:
  1. User has locale set in browser:
     - Attempt to update the form language to locale.
     - If the language doesn't exist in the form, fallback to `default_language`.
     - If there is no `default_language`, then default to `en`.
  2. User **does not** have locale set in browser:
     - Attempt to use to `default_language`.
     - If there is no `default_language`, then default to `en`.
     - This is probably the default behaviour of web-forms, so there may be no actions points on this one.

Does this sound reasonable?

DanielJDufour:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #2358

## Describe this PR

Added warning message div inside of the drawer that contains the iframe for web forms.  Add a message to this div if there isn't a version of the form for the app's language.

Technically speaking, we calculate the number of translations provided in the form instance for each language.  Unfortunately, we have to do that because even if a user doesn't provide translations for some languages, it appears the system (maybe on the backend?) will inject translations for some questions.  This can lead to a situation where the odkForm reports that there are 6 languages available, but in reality only 2 of them have translations of more than 50% of the questions.

## Screenshots

Screenshot 2025-04-27 at 5 23 49 PM

I thought of displaying the warning inside of the iframe, but I think passing this to the iframe or injecting it would lead to more complexity.  I also worked on an approach of iterating over relevant form questions and seeing if translations existed, but this quickly got very complicated.  This was additionally complicated because some questions become relevant based on what the user selected beforehand.

## Review Guide


DanielJDufour:
re-opening on clean branch

Issue #2426, opened 2025-04-25
Allow default organisation to be configurable via dotenv overrides


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- In `organisation_crud` we have:

```python
    org_in = OrganisationIn(
        name="HOTOSM",
        description="Humanitarian OpenStreetMap Team.",
        url="https://hotosm.org",
        associated_email="sysadmin@hotosm.org",
        odk_central_url=settings.ODK_CENTRAL_URL,
        odk_central_user=settings.ODK_CENTRAL_USER,
        odk_central_password=settings.ODK_CENTRAL_PASSWD.get_secret_value()
        if settings.ODK_CENTRAL_PASSWD
        else "",
        approved=True,
    )
    with open("/opt/app/images/hot-org-logo.png", "rb") as logo_file:
        org_logo = UploadFile(
            file=logo_file,
            filename="hot-org-logo.png",
            headers={"Content-Type": "image/png"},
        )
```

- We hardcode HOTOSM and the logo.
- Ideally we have this configurable by the user from dotenv **overrides**.
- The default without env vars remain the same.

**Describe the solution you'd like**

- If the variables are configured, they should be picked up by `config.py`
- `DEFAULT_ORG_NAME`, `DEFAULT_ORG_URL`, `DEFAULT_ORG_LOGO_URL`
- Description should probably just be something generic like 'Default organisation' to make it simple.
- The logo is the hard part. DEFAULT_ORG_LOGO_URL should probably be a remote link to make this easier. If present, we attempt download. If download success, we should pass wrapped in UploadFile.
- I'm pretty sure the filename doesn't matter, as it's written to a specific S3 key, but worth checking.

PR #2441, opened 2025-04-28
Add default organization settings and logo handling from env

Anuj-Gupta4:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related to: #2426 

## Describe this PR

Make the default organization generated during setup be configurable through the env file.


src/backend/app/organisations/organisation_crud.py
```diff
@@ -67,31 +69,38 @@ async def init_admin_org(db: Connection) -> None:
...
-            db,
-            org_in,
-            admin_user.sub,
-            org_logo,
-            ignore_conflict=True,
-        )
+
+    org_logo = None
+    if settings.DEFAULT_ORG_LOGO_URL:
+        async with httpx.AsyncClient() as client:
```
spwoodcock, line 86: `aiohttp` is probably more appropriate here. It's a much smaller lib than `httpx`, specifically for making requests.

We use `httpx` to stand up a testing server in pytest, but it probably shouldn't be used in the backend code, as it's only in the `test` dependencies in pyproject.toml.
By coincidence it is actually bundled with FastAPI though.

I see httpx is used in total_mapped.py, but this should be refactored at some point.
All the other code in the backend uses aiohttp 👍

src/backend/app/config.py
```diff
@@ -159,6 +159,13 @@ class Settings(BaseSettings):
     FMTM_DOMAIN: str
     FMTM_DEV_PORT: Optional[str] = "7050"
 
+    DEFAULT_ORG_NAME: Optional[str] = "HOTOSM"
+    DEFAULT_ORG_URL: Optional[str] = "https://hotosm.org"
+    DEFAULT_ORG_EMAIL: Optional[str] = "syadmin@hotosm.org"
+    DEFAULT_ORG_LOGO_URL: Optional[str] = (
+        "https://img.freepik.com/free-vector/colorful-bird-illustration-gradient_343694-1741.jpg"
```
spwoodcock, line 166: In https://github.com/hotosm/fmtm/commit/9f26a6aee619ec795773ed03152ae35fec07c34f I moved the `hot-org-logo.png` from the backend to frontend.

It should be served under FMTM_DOMAIN/hot-org-logo.png, meaning we could set that as the default now 👍 

If the backend started on it's own, the logo wouldn't be available. But if the frontend is started, then the backend started again, it would update the DbOrganisation entry with the logo.

Could we make this work?

Alternatively we just reference the git repo image as the default here:
`https://raw.githubusercontent.com/hotosm/fmtm/refs/heads/development/src/frontend/public/hot-org-logo.png`

spwoodcock (changes requested):

Anuj-Gupta4:
Yes, I have used the org logo URL based on the development branch. I meant to say that since we are using development based URL, the URL will work only after the PR has been merged since the image does not exist in development branch yet.

spwoodcock:
Got you, thanks!

Issue #2401, opened 2025-04-17
Add option to lock / mark complete project after mapping is complete (manager blocks additional contributions)


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- If a project is entirely mapped, the manager may wish to mark it as 'complete' and block additional contribution.

**Describe the solution you'd like**

Project status has:

```python
    DRAFT = "DRAFT"
    PUBLISHED = "PUBLISHED"
    ARCHIVED = "ARCHIVED"
```

Presumably:
- Draft = being created
- Published = ready for mapping
- Archived = hidden

So:
- We could use 'Archived', but perhaps we want archived projects to not display on the UI?
- Instead should we add 'Complete' as an option too?
- Any complete project should be visible, but no mapping can take place.
- To simplify this, we could mostly do it via the API, and just return 'This project has been locked as it is complete'.
- Then the user will be informed, without having to modify the frontend too much.

spwoodcock:
This replaces #565

manjitapandey:
Yes, you make total sense. 
For admins only, we show draft projects. The complete status can also be made accessible by applying completed filter. 
For other users we should only show the list of published projects by default and it shouldn't require the badge on project cards. 
But in project details page, its still a good idea to add project status.

PR #2478, opened 2025-05-05
Add project complete status to set it to read only

Anuj-Gupta4:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- Related to: #2401 

## Describe this PR

Added `COMPLETED` project status.
Updated access check dependency to easily control access to endpoints when project is set as completed.
The completed projects will be read only.

### DIFF
--- a/src/backend/app/auth/roles.py
+++ b/src/backend/app/auth/roles.py
@@ -46,6 +46,7 @@ async def check_access(
     org_id: Optional[int] = None,
     project_id: Optional[int] = None,
     role: Optional[ProjectRole] = None,
+    check_completed: bool = False,
 ) -> Optional[DbUser]:
     """Check if the user has access to a project or organisation.
 
@@ -64,6 +65,7 @@ async def check_access(
         org_id (Optional[int]): Org ID for organisation-specific access.
         project_id (Optional[int]): Project ID for project-specific access.
         role (Optional[ProjectRole]): Role to check for project-specific access.
+        check_completed (bool): If True, check if project is completed.
 
     Returns:
         Optional[DbUser]: The user details if access is granted,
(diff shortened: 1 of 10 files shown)


src/backend/app/auth/roles.py
```diff
@@ -89,42 +91,54 @@ async def check_access(
...
-                        OR (
-                            %(org_id)s IS NULL
+                        -- Check if project is completed and check_completed is true
+                        NOT (
+                            %(check_completed)s
                             AND EXISTS (
+                                SELECT 1
+                                FROM projects
+                                WHERE id = %(project_id)s
+                                AND status = 'COMPLETED'::public.projectstatus
```
spwoodcock, line 101: As we also have `ARCHIVED` as an option, could we also check for that?

We should probably block modification of COMPLETED or ARCHIVED projects, with the distinction being that archived projects are hidden for viewing, while completed projects can be viewed but not edited.

spwoodcock (changes requested):

src/backend/app/auth/roles.py
```diff
@@ -272,21 +289,27 @@ async def wrap_check_access(
...
-    # NOTE here we get the project manually to avoid warnings before the project
-    # if fully created yet (about odk_token not existing)
-    project = await DbProject.one(db, project_id, warn_on_missing_token=False)
-    return await wrap_check_access(
-        project,
-        db,
-        current_user,
-        ProjectRole.PROJECT_MANAGER,
-    )
+def project_manager(check_completed: bool = False):
```
spwoodcock, line 292: Can you explain the wrapper usage rationale a bit?

Wrapping an async def in a standard def probably doesn't make sense, as the async will have no benefit

src/backend/app/projects/project_deps.py
```diff
@@ -49,6 +49,15 @@ async def get_project_by_id(db: Connection, project_id: int, minimal: bool = Fal
         raise HTTPException(status_code=HTTPStatus.NOT_FOUND, detail=str(e)) from e
 
 
+async def get_active_project(project: Annotated[DbProject, Depends(get_project)]):
+    """Check if the project is not completed."""
+    if project.status == "COMPLETED":
```
spwoodcock, line 54: Likewise, it would be good if this was generic. If COMPLETED or ARCHIVED, then:
`f"Project is locked as it is {status.lower}."`

src/backend/app/projects/project_routes.py
```diff
@@ -194,7 +194,7 @@ async def get_odk_entities_geojson(
     response_model=list[central_schemas.EntityMappingStatus],
 )
 async def get_odk_entities_mapping_statuses(
-    project_user: Annotated[ProjectUserDict, Depends(mapper)],
+    project_user: Annotated[ProjectUserDict, Depends(mapper())],
```
spwoodcock, line 197: A callable probably shouldn't be used in Depends.
`Depends(mapper)` should work just fine

spwoodcock (changes requested):
see comments

src/backend/app/auth/roles.py
Anuj-Gupta4: I wanted to use the `check_access` dependency to restrict access whenever the project was completed, since it already existed and used raw sql rather than python code. But only some apis need to be restricted while others didnt. So, I decided to use parameterized dependencies to pass a project completion check. There are docs for it that uses class based dependency.

https://fastapi.tiangolo.com/advanced/advanced-dependencies/

I went with a simple wrapper because the code looked simpler. As far as I can tell, the wrapper only configures the boolean variable to set up the dependency and the dependency still runs asynchronously.

But maybe, class based dependency is more viable? If so, the code would look something like this:

```
class ProjectManagerDependency:
    def __init__(self, check_completed: bool = False):
        self.check_completed = check_completed

    async def __call__(
        self,
        project_id: int,
        db: Annotated[Connection, Depends(db_conn)],
        current_user: Annotated[AuthUser, Depends(login_required)],
    ) -> ProjectUserDict:
        project = await DbProject.one(db, project_id, warn_on_missing_token=False)
        return await wrap_check_access(
            project,
            db,
            current_user,
            ProjectRole.PROJECT_MANAGER,
            check_completed=self.check_completed,
        )


# Initialize with default params (can be overridden per route)
project_manager = ProjectManagerDependency(check_completed=False)
```

src/backend/app/projects/project_routes.py
Anuj-Gupta4: `project_user.get("project")` does not return project without calling the wrapper function

spwoodcock: If that's the case then there is something wrong with the `mapper` function I would say then 😅 

You shouldn't call the function when passing it to Depends(), but pass the function object itself.
`Depends(mapper())` is wrong, because it would call `mapper()` immediately at app startup, not per-request.

src/backend/app/auth/roles.py
spwoodcock: This is a nice idea!

spwoodcock (changes requested):
This looks great, my only concern is around the usage of `mapper()` in the Depends:

https://fastapi.tiangolo.com/tutorial/dependencies/#create-a-dependency-or-dependable

Correct me if I'm wrong, but I believe endpoints are loaded at API startup, meaning using `mapper()` would actually evaluate the function immediately. We need to use `Depends(mapper)` to the function is used with variable injection correctly when the endpoint is called

spwoodcock:
Sorry I modified the base schema usage significantly to be modular - this will need a rebase too, so it's built on top of the modular SQL 😅

spwoodcock (approved):
Nice change!

PR #2491, opened 2025-05-08
Mapper frontend layer related bug fixes

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR
- Fix data extract layer flickering on map click within task area (Reason for this issue is that whenever the map is clicked, the selected task & it's outline was being set to null first and then only updated to the curr selected task i.e. say task 3 is selected and state is set to task 3 ---> when the task is clicked again, the task first set to null then only updated to 3)

- Fix selected entities not being highlighted
- Fix point entity not visible
- Fix layer not rendering issue on project switch

## Screenshots
https://github.com/user-attachments/assets/947db45e-cb1d-4292-8299-2637f8ef398e


spwoodcock:
Thanks so much for picking this up!

I refactored a lot to try and simplify, but missed a lot of referenced to selectedEntity vars & made a few mistakes 😅

Issue #2488, opened 2025-05-08
The option to choose between ODK Collect and Webform is not functional


manjitapandey:
**Describe the bug**
During project creation, there is an option for project creator to use odk collect instead of webforms. But it is not functional, rather it is server based. I mean, whatever option you choose it always odk collect based form submission available on dev server.

**To Reproduce**

1. Go to 'create project step 1 on dev server'
2. Skip click on 'use odk collect instead of webform'
3. Create project and go to mapper's page
4. See odk collect is available instead of webform option
5. See error

**Expected behavior**
The choice between odk collect and webform is not functional

**Screenshots**

![Image](https://github.com/user-attachments/assets/7f960ebc-6041-4ce4-aa15-f854276d8899)

**Additional context**
The issue seems to only be on dev server and not in staging and production.

NSUWAL123:
The issue is not related to dev server. It's related to the config file on the corresponding server.
In the dev server, enableWebForms is set to false, whereas in staging, enableWebForms is set to true.

After the frontend implementation of the project-specific option to enable/disable webforms, this issue will be solved.

spwoodcock:
Yeah Nishit is right here - closing this issue as i think we are good 👍

Feel free to open again if needed.

manjitapandey:
Thanks both, It clears me up.

Issue #2352, opened 2025-04-08
Newly created polygons are stored below tasks and existing features layers


manjitapandey:
**Describe the bug**
Newly created polygons are stored below tasks and existing features layers making it hard to select. 

**To Reproduce**

1. Go to 'mapper view of any project'
2. Click on 'map new feature' and draw new feature  just above existing building features
3. Proceed with incomplete form submission
4. Try getting back to FMTM and select same feature
5. See error

**Expected behavior**
Maybe a building layer toggle on/off may solve the issue. If mapper encountered this type of issue on field they can toggle off building layer temporarily and select layers below. 

**Screenshots**

![Image](https://github.com/user-attachments/assets/9e828dd7-5d69-4fb6-ad7d-d7eae2eecc97)

**Additional context**
Maybe it can fix this too: [Remove multipolygons / relations from OSM data extract during project creation](https://github.com/hotosm/fmtm/issues/1801)

NSUWAL123:
I think there are two possible solutions: 
1. Have a checkbox for toggling on/off the layer (osm/new feature layers). This way, if a new feat overlaps an osm feat, the new feat layer could be toggled off, and vice versa.
2. If the clicked coordinate contains more than 1 feature, display a modal (like task/entities actions), where the user would be asked, `Which feature would you like to map?`, then have buttons `Select this feature` ---> then after btn click highlight the entity (with red border) and show `Cancel `& `Confirm` btn. On cancel, show the previous step of selecting the feature, if confirm, show the entity action modal. I think this one would be an ideal solution. 

@spwoodcock @manjitapandey

spwoodcock:
Love option 2! Its very flexible 😃

You may not even need the 'which feature would you like to map' text, as its quite intuitive that the user is selecting between geometries.

manjitapandey:
Everything is working fine. But I feel we are going for one extra step we don't need. 
When I select a feature, then it prompt me to map or cancel that feature. When clicked on map the feature, it again ask for map feature in odk, which felt like extra step for me. 

https://github.com/user-attachments/assets/5a24267f-f247-4933-8493-4947b9a7be5b

spwoodcock:
Yeah you are probably right - it could be:
- Click features
- 'Select This Feature'
- (remove the 'Cancel' and 'Map This Feature' dialog)
- Go straight to the standard popup for entity selection to 'Map This Feature In ODK'

manjitapandey:
[...]

PR #2492, opened 2025-05-09
Project specific ODK form

NSUWAL123:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #2488

## Describe this PR
- Display webforms or allow users to use Othe DK app based on project settings
- Remove enableWebforms from config.json with implementation of project-specific ODK

## Screenshots
image
image

### DIFF
--- a/src/mapper/src/routes/[projectId]/+page.svelte
+++ b/src/mapper/src/routes/[projectId]/+page.svelte
@@ -64,11 +64,6 @@
 	const latestEvent = $derived(taskStore.latestEvent);
 	const commentMention = $derived(taskStore.commentMention);
 
-	// Set useOdkCollect override to disable webforms in app
-	if (project.use_odk_collect) {
-		commonStore.setUseOdkCollectOverride(true);
-	}
-
 	// Update the geojson task states when a new event is added
 	$effect(() => {
 		if (latestEvent) {
@@ -120,6 +115,7 @@
 	}
 
 	onMount(async () => {
+		commonStore.setEnableWebforms(!project.use_odk_collect);
 		await entitiesStore.subscribeToNewBadGeom(newBadGeomStream);
 
 		// In store/tasks.svelte.ts
--- a/src/mapper/src/routes/[projectId]/+page.svelte
+++ b/src/mapper/src/routes/[projectId]/+page.svelte
@@ -48,35 +48,38 @@
 	let latestEventTime: string = $state('');
 	let isGeometryCreationLoading: boolean = $state(false);
 
+	const projectId = $derived(data.projectId);
+	const project = $derived(data.project);
+
 	const taskStore = getTaskStore();
 	const entitiesStore = getEntitiesStatusStore();
 	const commonStore = getCommonStore();
 	const alertStore = getAlertStore();
 
-	const taskEventStream = getTaskEventStream(data.projectId);
-	const entityStatusStream = getEntityStatusStream(data.projectId);
-	const newBadGeomStream = getNewBadGeomStream(data.projectId);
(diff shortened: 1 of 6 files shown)


src/mapper/static/config.json
```diff
@@ -3,7 +3,6 @@
 	"logoText": "Humanitarian OpenStreetMap Team",
 	"cssFile": "https://cdn.jsdelivr.net/npm/@hotosm/ui@0.2.0-b6/dist/style.css",
 	"cssFileWebformsOverride": "",
-	"enableWebforms": false,
```
spwoodcock, line 6: The values are here as examples, plus this is the default file if no config in S3.

I think we will need
enableWebforms: true

src/mapper/src/routes/[projectId]/+page.svelte
```diff
@@ -48,35 +48,33 @@
 	let latestEventTime: string = $state('');
 	let isGeometryCreationLoading: boolean = $state(false);
 
+	const projectId = $derived(data.projectId);
+	const project = $derived(data.project);
```
Comment (spwoodcock) [inline src/mapper/src/routes/[projectId]/+page.svelte:52]: data doesn't need derived I don't think, but I think the recommended Svelte way is prop destructuring

`const { project } = data`

src/mapper/src/routes/[projectId]/+page.svelte
```diff
@@ -48,35 +48,33 @@
 	let latestEventTime: string = $state('');
 	let isGeometryCreationLoading: boolean = $state(false);
 
+	const projectId = $derived(data.projectId);
```
Comment (spwoodcock) [inline src/mapper/src/routes/[projectId]/+page.svelte:51]: Not 100% on this, but I think all reactive svelte vars should use `let` and not `const`

spwoodcock (changes requested):

src/mapper/static/config.json
NSUWAL123: Since we are using the value of use_odk_collect from the project API, why do we need this?

spwoodcock: I think there is a misunderstanding on the use of both vars:
- enableWebforms is per instance (its a feature flag to enable web forms functionality, else the instance uses ODK Collect for every project (ignore use_odk_collect)
- The use_odk_collect param is per project and is an override. Its anticipated that most uses use web forms, but its good to have an override if necessary (web forms has some missing form functionality)

NSUWAL123:
Forgot to switch the branch. This PR is also related to: 
- #2352 (new-geom, bad-geom etc layers vanish on baselayer switch)
- #2493 (UI crash when try to validate the submission from project-details page)

src/mapper/static/config.json
NSUWAL123: I think I finally got your point: 
You mean that if enableWebforms is false, though setting the use_odk_collect to false during project creation, we don't show webforms because we have disabled the feature of webforms in that particular instance.

And if enableWebforms is true, the user can override the behaviour of using webforms by setting the user_odk_collect to true, and the user can use ODK Collect instead of webforms.

NSUWAL123: Also, there seems to be a bug with this condition:
```
if (project.use_odk_collect) {
    commonStore.setUseOdkCollectOverride(project.use_odk_collect);
}
```

What's happening here is, the initial state of `useOdkCollectOverride` is false, when a project with use_odk_collect set to true is opened, then the useOdkCollectOverride is set to true. Now if a user then opens a project where use_odk_collect set to false, the state of useOdkCollectOverride is not updated to false, due to which only mapping through ODK app is set.

Issue #2341, opened 2025-04-03
Consider clusting on map for projects that contain only point data


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- Say we have a project that maps 1000 points in an area.
- Or a project that only maps new points.
- Perhaps we should consider some form of clustering for performance.

**Describe the solution you'd like**

- As clustering to maplibre.
- Adding it to openlayers is a bit of a waste of time, if we end up migrating.

**Additional context**

- This would cause an issue knowing what colour to have the clusters though.
- Having all the points available on the screen with their mapping colours gives a good indication of progress.
- If the points are hidden within clusters, it's hard to tell what is remaining to be mapped (as we can't see the colours).

manjitapandey:
Related #2340

Issue #2553, opened 2025-05-28
Point entities not being saved on odk_entities table


NSUWAL123:
**Describe the bug**
- When a project is created with the primary geom type of point, the features are not visible on the map (in management) & nothing happens when point feature is clicked on mapper frontend.
- Also the `/status` throws 500 error if primary geom type is point.
- The response from electric shape from odk_entities is also returning empty array
(All the above behaviour is only shown when project is created with primary geom of point)

The main reason here seems to be that the point entities are not saved in the `odk_entities` table while creating project with primary geom type of point.

**To Reproduce**

**Expected behavior**

**Screenshots**
/status api throwing 500 error
![Image](https://github.com/user-attachments/assets/efaeb6e2-697d-40b1-b3a5-be91f461a461)

manjitapandey:
This should be fixed for linestring as well.

PR #2555, opened 2025-05-28
Clustering if primary geoms are point & update point icon

NSUWAL123:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to https://github.com/hotosm/fmtm/issues/2341

## Describe this PR
- Cluster point features on both mapper & management frontend

## Screenshots
https://github.com/user-attachments/assets/656e0156-22d4-4849-ae5f-1a2b5f993da3
https://github.com/user-attachments/assets/ae8b531e-2308-4fcd-bdb0-ead522fd32cc


NSUWAL123:
Blocked by #2553

NSUWAL123:
There's one problem where I'm stuck: 
When using clustering on mapper frontend, the clustering works absolutely fine, the extracts are clipped within the task area and everything works fine. 
But when the baselayer is changed, the clustering starts to break and functions weirdly. When a different task is selected, the geometries for that task are not visible; rather, the cluster of the previous task is shown, and the clustering also breaks on zooming in and zooming out. 

The findings I found are: 
- When I used the demo data of svelte-maplibre, and tried to change the base layer, it worked absolutely fine,
- When using the FGB data to show points directly (showing data in the circle layer), it works fine
So, there's no problem with the clustering that svelte-maplibre does, and it's not a problem with the FGB layer alone as well. 
In conclusion, it seems that when clustering FGB data & switching the baselayer, the issue arises.

The first half of the video shows the clustering of the data extracts, where the second half shows the demo data from svelte-maplibre used for clustering. So clustering is not working as expected when using it with svelte-maplibre.
https://github.com/user-attachments/assets/ae76f40c-2e9c-40b1-b4a1-6e7f1554a9e2

Any inputs from you would be very helpful, @spwoodcock

NSUWAL123:
Yeah, the problem arises when switching the base layer when there's clustering (specifically when using fgb with clustering)

spwoodcock:
If the fgb is the problem, should we start with only clustering the new geometries layer?

Then we can look into the fgb layer after

spwoodcock (approved):

Issue #2561, opened 2025-05-29
Issues while trying to map road features on Field-TM


manjitapandey:
**Describe the bug**
Field TM used to support road mapping exercise previously but we haven't been using those features and maintaining them with recent code changes. Causing that few things seems to be broken while creating project or mapping a project having road (linear features) as primary features. 

**Expected behavior**
The following broken steps should be made functional for road mapping via field-tm. 

- [ ] Download highway data as map features from osm should be functional. 
- [ ] After the project is created, the data extract is not visible on project details page

**Screenshots**

![Image](https://github.com/user-attachments/assets/663f9424-b84c-4ba5-92d1-8b8166d15c4c)

**Additional context**

PR #2571, opened 2025-05-30
Feat roadways issue fix

varun2948:

- [X] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Trying to work on multiple part of this issue #2561 
## Describe this PR

Worked on the provided geojson of Tommy Charles in the hotosm slack which now helps the user to map roadways aswell
- Fix project creation based on outside geojson for road / Linestring layers.
- Fix data-extract not showing up on both frontend because of the lack in styling of "Linestring"
- Added one migration to add enum for Linestring on geomtype which was missing.

NOTE: Still need some code to work on osm data extract for road, which i can do later on.

## Review Guide

- Create a new project based on roadways line string with custom data extract not generating data extract 

### DIFF
--- a/src/backend/app/db/postgis_utils.py
+++ b/src/backend/app/db/postgis_utils.py
@@ -339,7 +339,7 @@ def add_required_geojson_properties(
             # Ensure negative osm_id if it's derived
             # NOTE this is important to denote the geom is not from OSM
             if not properties.get("osm_id"):
-                if("/" in osm_id ):
+                if(osm_id is type(str) and "/" in osm_id ):
                     osm_id = osm_id.split('/')[-1]
                 osm_id = -abs(int(osm_id))
 
--- a/src/backend/app/db/postgis_utils.py
+++ b/src/backend/app/db/postgis_utils.py
@@ -715,18 +715,14 @@ def merge_polygons(
             properties = feature["properties"]
             geom = shape(feature["geometry"])
             
-            # Handle LineString geometries by converting to Polygon
             if isinstance(geom, (LineString, MultiLineString)):
                 if isinstance(geom, MultiLineString):
-                    # Convert each LineString to a Polygon using buffer
                     for line in geom.geoms:
-                        # Create a small buffer around the line to form a polygon
-                        buffered = line.buffer(0.0001)  # Approximately 11 meters at equator
+                        buffered = line.buffer(0.0001)  
                         if buffered.is_valid:
                             geom_list.append(buffered)
                 else:
-                    # Create a small buffer around the line to form a polygon
-                    buffered = geom.buffer(0.0001)  # Approximately 11 meters at equator
+                    buffered = geom.buffer(0.0001)  
                     if buffered.is_valid:
                         geom_list.append(buffered)
                 continue
--- a/src/backend/app/db/postgis_utils.py
+++ b/src/backend/app/db/postgis_utils.py
@@ -32,7 +32,7 @@ from osm_rawdata.postgres import PostgresClient
 from psycopg import Connection, ProgrammingError
 from psycopg.rows import class_row, dict_row
 from psycopg.types.json import Json
-from shapely.geometry import MultiPolygon, Point, Polygon, mapping, shape
+from shapely.geometry import MultiPolygon, Point, Polygon, mapping, shape, LineString, MultiLineString
 from shapely.geometry.base import BaseGeometry
 from shapely.ops import unary_union
 
@@ -339,6 +339,8 @@ def add_required_geojson_properties(
             # Ensure negative osm_id if it's derived
             # NOTE this is important to denote the geom is not from OSM
             if not properties.get("osm_id"):
+                if("/" in osm_id ):
+                    osm_id = osm_id.split('/')[-1]
                 osm_id = -abs(int(osm_id))
 
             feature["id"] = str(osm_id)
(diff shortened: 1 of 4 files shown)


varun2948:
@Anuj-Gupta4 i am having an issue on the PR. It seems to be from the previous merge, as suggested by @Sujanadh. Can you have a look and try to fix this check fail? Let me and @manjitapandey know if it gets fixed, and please merge it if you can so that manjita can see the changes.

Anuj-Gupta4:
I have fixed the test cases on backend side. However, I noticed that the entities are not being saved on odk central. I will look into this first.

src/backend/app/db/postgis_utils.py
```diff
@@ -711,19 +722,42 @@ def merge_polygons(
...
+            geom = shape(feature["geometry"])
+
+            if isinstance(geom, (LineString, MultiLineString)):
+                if isinstance(geom, MultiLineString):
+                    for line in geom.geoms:
+                        buffered = line.buffer(0.0001)
+                        if buffered.is_valid:
+                            geom_list.append(buffered)
+                else:
+                    buffered = geom.buffer(0.0001)
```
spwoodcock, line 734: This code will convert linestrings to polygons right? I'm not sure thats what we want.

Buffering could be done in the frontend for easier display, but the underlying data should probably remain linestring no?

spwoodcock: Perhaps I'm wrong tho, as you added frontend logic for styling linestrings and I assume you are using that?

src/mapper/src/lib/components/map/main.svelte
```diff
@@ -592,30 +592,56 @@
...
 							['get', 'status'],
 							'READY',
-							'MAP_PIN_GREY',
+							cssValue('--entity-ready'),
 							'OPENED_IN_ODK',
-							'MAP_PIN_YELLOW',
+							cssValue('--entity-opened-in-odk'),
 							'SURVEY_SUBMITTED',
-							'MAP_PIN_GREEN',
+							cssValue('--entity-survey-submitted'),
```
spwoodcock, line 633: Nice 🙌

src/migrations/002-update-geomtype-enum.sql
```diff
@@ -0,0 +1,10 @@
+-- ## Migration reset the _migrations table.
+
+-- Start a transaction
+BEGIN;
+-- Create new enum type with all values
+
+ALTER TYPE public.geomtype ADD VALUE IF NOT EXISTS 'LINESTRING';
```
spwoodcock, line 7: Migration is good, but we also need this set in the base schema, as migrations get archived on new release 👍

Anuj-Gupta4:
This works now on backend, but the project is no longer visible on mapper frontend. I used Cockle Bay Roads and Bridges1.geojson that Tommy Charles shared on slack.

src/backend/app/db/postgis_utils.py
Sujanadh: Perhaps the choice of this function `merge_polygons` is wrong here. I created this function to convert multipolygon AOI to single AOI. Instead we could use `multigeom_to_singlegeom` to convert multilinestring to linestring if that is necessary.

Anuj-Gupta4:
A couple of things to address:

1. **Task Splitter & Geometry**  
   Task splitting does not currently work for `LINESTRING`. The available options in the project area split section should dynamically change based on the selected geometry type.

2. **Project Visibility**  
   Projects should be:
   - **Mappable** on the [mapper page](https://mapper.dev.fmtm.hotosm.org/)
   - **Manageable** on the [management page](https://dev.fmtm.hotosm.org/)

src/migrations/init/shared/1-enums.sql
```diff
@@ -32,7 +32,9 @@ ALTER TYPE public.mappinglevel OWNER TO fmtm;
 CREATE TYPE public.geomtype AS ENUM (
     'POINT',
     'POLYLINE',
```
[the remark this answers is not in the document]
Anuj-Gupta4: I have added MultiPolyline and removed Linestring and MultiLinestring.

NSUWAL123 (approved):

Sujanadh:
> A couple of things to address:
> 
> 1. **Task Splitter & Geometry**
>    Task splitting does not currently work for `LINESTRING`. The available options in the project area split section should dynamically change based on the selected geometry type.
> 2. **Project**
>    Projects should be:
>    
>    * **Mappable** on the [mapper page](https://mapper.dev.fmtm.hotosm.org/)
>      This is not working.
>    * **Manageable** on the [management page](https://dev.fmtm.hotosm.org/)

1. If the primary geom type is linestring, we can retrieve all the polygon features from OSM while task splitting and split the area of interest based on the number of features.
2. But option 1 doesn't seem feasible in the case of a linestring, as we split the area based on the linear features present. So splitting while using linestring as the primary geom type doesn't make sense at all. For this we can simply avoid the options to split during project creation.

src/backend/app/db/postgis_utils.py
Sujanadh: I think buffering a linestring and using it as a polygon in a data extract doesn't raise problems for now, as we can select, collect and validate data that comes from submissions, but when it comes to uploading all the validated features into the OSM, it will be an issue. So we need to address it later or sooner before or after release based on priority. 
cc. @manjitapandey

Issue #2526, opened 2025-05-15
Ensure features with `SURVEY_SUBMITTED` status are not reset accidentally when creating a second submission


spwoodcock:
**Describe the bug**

- After submitting the survey, features go green with `SURVEY_SUBMITTED`.
- If the user opens to webform / survey again (say accidentally), then the feature status will be updated to `OPENED_IN_ODK` again.
- We need to avoid this with logic to block that.
- If `SURVEY_SUBMITTED`, do not update the entity status upon opening the survey (simply skip the API call / PGLite update)

spwoodcock:
Downgraded - this is definitely nice to have, but not a blocker for now

varun2948:
@spwoodcock was looking over the pglite whole day but surely can switch it to this one for now. I cannot apparenly test the issue out but was trying different things out - #2532

spwoodcock:
> @spwoodcock was looking over the pglite whole day but surely can switch it to this one for now. I cannot apparenly test the issue out but was trying different things out  
> 
> - #2532 

No sorry I'm just going through issues and assigning between people - no urgency to swap!

The PGlite issue is the most urgent 👍

PR #2601, opened 2025-06-11
Fix/mapper entity bugs

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #2526

## Describe this PR
- Optimize entity selection by using the map object to find selected entities details instead of using array find, since sometimes selecting entity was laggy
- Fix Geo-restrict functionality not working when using webforms
- Only update status to `opened_in_odk` when the entity is in the ready state
- Code refactors

## Screenshots

### DIFF
--- a/src/mapper/src/lib/components/dialog-entities-actions.svelte
+++ b/src/mapper/src/lib/components/dialog-entities-actions.svelte
@@ -274,56 +274,54 @@
 						</div>
 					{/if}
 				</div>
-				{#if selectedEntity?.status !== 'SURVEY_SUBMITTED' && selectedEntity?.status !== 'VALIDATED'}
-					<div class="entity">
+				<div class="entity">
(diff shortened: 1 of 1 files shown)


NSUWAL123:
> I will prefer to Allow multiple submissions for a single feature but can we not downgrade the status ?? Like once the submission is sent, the feature colour changes to green and then to red or blue based on either its approved or rejected. In case of approved or just submission sent, we shouldn't change colour to yellow(opened in odk status), when clicking for second submission.

Yeah, we can do that. Will update and push the changes accordingly!

spwoodcock (approved):

Issue #2619, opened 2025-06-16
Unable to invite users by osm username with space in username


manjitapandey:
**Describe the bug**
On manage users page, when I invite user via osm. Username having space is treated as a multiple username instead of single. 

**To Reproduce**

1. Go to 'manage users page'
2. Write an osm username  with space.
3. Click on add

**Expected behavior**
OSM username with space in between should be accepted as a single user instead of multiple. 

**Screenshots**

![Image](https://github.com/user-attachments/assets/17a412e5-ffe8-46da-886f-d7321f95f0f1)

**Additional context**

PR #2621, opened 2025-06-16
Invite multiple OSM user by comma separated value

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #2619

## Describe this PR
Invite multiple OSM users by inserting comma separated OSM usernames

## Screenshots
https://github.com/user-attachments/assets/cc62eb3f-b105-400c-97a5-7542bf151e2e

### DIFF
--- a/src/frontend/src/components/ManageProject/User/Invite.tsx
+++ b/src/frontend/src/components/ManageProject/User/Invite.tsx
@@ -117,10 +117,16 @@ const InviteTab = ({ roleList }: propType) => {
                 name="name"
                 label="Invite User"
                 value={user}
-                onChange={(e) => setUser(e.target.value.replace(/[\r\n]+/g, ' ').replace(/\t/g, ' '))}
+                onChange={(e) => {
+                  if (values.inviteVia === 'osm') {
+                    setUser(e.target.value.replace(/[\r\n]+/g, ', '));
+                  } else {
+                    setUser(e.target.value.replace(/[\r\n]+/g, ' ').replace(/\t/g, ' '));
+                  }
+                }}
                 placeholder={
                   values.inviteVia === 'osm'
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/ManageProject/User/Invite.tsx
```diff
@@ -117,10 +117,16 @@ const InviteTab = ({ roleList }: propType) => {
...
+                    setUser(e.target.value.replace(/[\r\n]+/g, ', '));
+                  } else {
+                    setUser(e.target.value.replace(/[\r\n]+/g, ' ').replace(/\t/g, ' '));
+                  }
+                }}
                 placeholder={
                   values.inviteVia === 'osm'
-                    ? 'Enter Username (To assign multiple users, separate osm usernames with space)'
+                    ? 'Enter Username (To assign multiple users, separate osm usernames with commas)'
                     : 'Enter Gmail (To assign multiple users, separate gmail addresses with space)'
```
spwoodcock, line 130: The code is updated for email too, but I think just this prompt text was missed 😃

NSUWAL123: The reason behind using space in emails is for bulk email assignment. There seemed to be many cases where the PMs maintained an excel record of emails and simply copy paste the email column in the field (similar to Gmail)
And since it is not feasible to use spaces in OSM, so had to use commas instead.

spwoodcock: Ah interesting, thanks!

Typically in email clients, they accept either comma or semicolon separation. Let's keep this as it is for now though!

Issue #2620, opened 2025-06-16
Single validate this feature option on project details page even for multiple submissions


manjitapandey:
**Describe the bug**
When I send multiple submissions in a single feature. Clicking on the feature on project details page shows only a single 'validate feature' option.

**To Reproduce**

1. Send multiple submission on single feature.
2. Go to project details page
3. Click on the feature
4. See error (only single validate this feature option is shown)

**Expected behavior**
Multiple validate this feature option should be visible if there are multiple submissions sent

**Screenshots**

![Image](https://github.com/user-attachments/assets/13d1a49b-c35d-491a-841e-7c03ee075c56)

![Image](https://github.com/user-attachments/assets/d875dd6e-6c7f-450d-ac38-6db64eb151ed)

![Image](https://github.com/user-attachments/assets/c5467c48-ceb8-4747-b4ea-412cc98a3b52)

**Additional context**

PR #2634, opened 2025-06-19
Sync multiple submission id's

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #2620

## Describe this PR
- add submission_ids to OdkEntitiesUpdate for webhook processing
- update submission_ids also when updating local db
- remove redundant status update on backend (status update is already handled during webhook processing)
- implement entity processing queue for efficient updates (the reason for implementing this is: 
```
for (const entity of rows) {
    const updatedEntity = {
        ...entityMapClone.get(entity.entity_id), // fallback to existing data
        ...entity, // overwrite with new status
    };
    entityMapClone.set(entity.entity_id, updatedEntity);
    // Store value in local db for persistence across restarts
    await DbEntity.update(db, entity);
    }
)
```
The await was here was making stale submission_ids data to be concatenated with the new submission_id, causing the final submission_ids value to be incorrect. And since we want the asynchronous behaviour while making updates in the DB, `processQueue `workflow is used so that it would asynchronously update on each electric shape and its rows.

## Screenshots
image

### DIFF
--- a/src/backend/app/central/central_schemas.py
+++ b/src/backend/app/central/central_schemas.py
@@ -329,3 +329,4 @@ class OdkEntitiesUpdate(BaseModel):
     """A small base model to update the OdkEntity status field only."""
 
     status: str  # this must be the str representation of the db enum
+    submission_ids: Optional[str] = None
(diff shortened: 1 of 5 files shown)


spwoodcock (approved):
Great job!

Issue #2470, opened 2025-05-02
Organization based submission download and infographics


manjitapandey:
**Is your feature request related to a problem? Please describe.**
Sometimes the actual projects may be divided into multiple sub projects within the same organisation and project managers may be interested in downloading the data for analysis or viewing the analytics of whole project instead of merging the parts to get actual stats. 

**Describe the solution you'd like**
Organization based statistics page where we can allow user to view analytics like number of submissions, number of task mapped and validated and more. This page can be made accessible to only organization admin or higher user roles. 
We can also provide an option for organization managers to download submissions of overall organization. 

**Additional context**
This is not the finalised UI but I have drafted my understanding in the figma. We can discuss further and polish requirements before we move towards development of the feature. 
This is also the requirement from Tokha Field surveying activity.
![Image](https://github.com/user-attachments/assets/d41fd022-09f8-471d-87af-be9124dd177f)

Sujan167:
I'm currently looking into this. Will share thoughts or questions soon, especially around the data aggregation and download flow for organization-level submissions and stats.

Sujan167:
At the moment, the /organisation/{org_id}/stats endpoint only includes the finalized data you see in the response and does not provide counts for draft, completed, or published projects. That information is currently available through the /projects/summaries API

However, if you’d prefer to have those project counts included directly in the /organisation/{org_id}/stats response, I’d be happy to add that as well

Sujan167:
Hi again 👋

Just an update: I’ve implemented the project status breakdown directly in the /organisation/{org_id}/stats response under the overview.project_status_counts field.

This now includes counts for each project status (e.g. DRAFT, PUBLISHED, COMPLETED), so you won’t need to rely on /projects/summaries for that data anymore.

Let me know if you have any feedback or need adjustments!
Now the final response looks like this:

```{
  "overview": {
   ...

    "project_status_counts": {
      "PUBLISHED": 1,
      "DRAFT": 3,
      "COMPLETED": 1,
      "ARCHIVED":0
      "
    }

...
}
```

PR #2689, opened 2025-07-08
Organization based submission download and infographics

Sujan167:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
Issues:
- #2470

## Describe this PR
- Implemented `get_organisation_stats` to retrieve aggregated statistics for an organisation, including task counts and daily activity.

- Added `download_organisation_submissions` to export submissions in CSV or GeoJSON format.

- Introduced new Pydantic models for organisation statistics in `organisation_schemas.py`.

- Updated routes to include endpoints for fetching stats and downloading submissions.

- Created utility functions for handling submission data and formatting outputs.

## Screenshots
- Organisation infographics
image
image

### DIFF
--- a/src/backend/app/organisations/organisation_crud.py
+++ b/src/backend/app/organisations/organisation_crud.py
@@ -392,37 +392,22 @@ async def get_organisation_stats(
 
     try:
         async with db.cursor() as cur:
+            params = {
+                "org_id": org_id,
+                "start_date": start_date,
+                "end_date": end_date,
+            }
+
             # Fetch org-wide stats
-            await cur.execute(
-                main_sql,
-                {
-                    "org_id": org_id,
-                    "start_date": start_date,
-                    "end_date": end_date,
-                },
-            )
+            await cur.execute(main_sql, params)
             row = await cur.fetchone()
 
             # Fetch daily stats
-            await cur.execute(
-                daily_sql,
-                {
-                    "org_id": org_id,
-                    "start_date": start_date,
-                    "end_date": end_date,
-                },
-            )
+            await cur.execute(daily_sql, params)
             daily_rows = await cur.fetchall()
 
             # Fetch per-project stats
-            await cur.execute(
-                project_task_sql,
-                {
-                    "org_id": org_id,
-                    "start_date": start_date,
-                    "end_date": end_date,
-                },
-            )
+            await cur.execute(project_task_sql, params)
             project_rows = await cur.fetchall()
 
         daily_stats = [


src/backend/app/organisations/organisation_utils.py
```diff
@@ -0,0 +1,130 @@
...
+"""Utils for download organisation submissions."""
+
+import csv
+import io
+
+from app.organisations.organisation_deps import get_org_odk_creds
+from app.submissions import submission_crud
+
+
+async def populate_odk_credentials_for_projects(projects, org):
```
spwoodcock, line 27: Is there a need for this logic?

When the project has empty credentials, then we default to use the organisation credentials.
The nice thing about doing that, is organisation creds could be changed in one place.

Could you possibly do an sql concat similar to the get projects endpoints?

src/backend/app/organisations/organisation_deps.py
```diff
@@ -127,3 +129,51 @@ async def org_from_project(
 ) -> DbOrganisation:
     """Get an organisation from a project id."""
     return await get_organisation(db, project.organisation_id)
+
+
+async def org_or_project_manager(
```
spwoodcock, line 134: I haven't dug too deeply into the logic here, but it seems a bit against what we have built to far:
- Generally any role that is higher in the hierarchy should automatically have the permissions of the lower roles

admin > org manager > proj manager > users / validators

I think most other things work that way - is it possible to do the same here?

If there a good justification why, then we could consider adding this approach 👍

src/backend/app/organisations/organisation_utils.py
```diff
@@ -0,0 +1,130 @@
...
+            filters["$filter"] = (
+                f"__system/submissionDate ge {start_date}T00:00:00+00:00 and "
+                f"__system/submissionDate le {end_date}T23:59:59.999+00:00"
+            )
+        except ValueError:
+            pass  # Gracefully skip if format is invalid
+    return filters
+
+
+async def collect_all_submissions(projects, filters):
```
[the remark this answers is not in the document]
Sujan167: Yeah, you're totally right — performance is definitely something to think about here.

Since each project might have its own ODK credentials, and ODK Central doesn’t support querying submissions across multiple projects in one go, we have to make one API call per project. So if there are N projects, that means N calls.

To help with performance, I’ve used asyncio.gather to run those API calls concurrently. And to keep things from getting too overwhelming or hitting rate limits, I’ve added a semaphore to limit concurrency to 10 projects at a time.

Happy to tweak the limit or explore other ideas if you think there’s a better way!😀

src/backend/app/organisations/organisation_utils.py
Sujan167: Each project may have own ODK credentials, so fallback organisation ODK credential will be injected to those projects which have empty credentials.

src/backend/app/organisations/organisation_deps.py
Sujan167: Ah, I wasn’t aware of that — thanks for pointing it out!

I’ve updated the logic so that this endpoint now allows access for both admin and org_admin, following the role hierarchy convention we’ve used elsewhere.

Let me know if that looks better now 👍

src/backend/app/organisations/organisation_utils.py
spwoodcock: This could be wrapped in a method on DbProject if it's useful in many places

Sujan167:
Since the submission payloads are large and include data from multiple projects, the response size becomes too big, which creates a bottleneck for OKD Central on long run. To avoid this, submissions should be downloaded directly from S3 instead.

There is currently a cron job that fetches submissions every 6 hours (if I’m not mistaken). This job should be adjusted to run more frequently — possibly every hour or even shorter, depending on feasibility.

Would love to hear your thoughts on this approach @spwoodcock

Anuj-Gupta4 (changes requested):

Anuj-Gupta4 (approved):

spwoodcock (approved):

PR #2696, opened 2025-07-11
Remove custom web-forms caching by key, defer to workbox pwa

spwoodcock:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Looking into this during https://github.com/hotosm/field-tm/issues/2674

## Describe this PR

- We implemented a custom caching mechanism for webforms, to avoid loading every time a feature is clicked.
- Since then, we added workbox runtime caching.
- Ideally the webforms can be cached as part of the PWA workbox config.
- If you think it's feasible, could you please continue this @DanielJDufour 🙏?
- If not, we can close and keep caching manually - but we need a way to avoid the cache conflicts that keep occurring (perhaps we need a rule to omit this path in workbox too).

### DIFF
--- a/src/mapper/src/lib/api/fetch.ts
+++ b/src/mapper/src/lib/api/fetch.ts
@@ -4,42 +4,6 @@ import { DbApiSubmission } from '$lib/db/api-submissions.ts';
 
 const API_URL = import.meta.env.VITE_API_URL;
 
-const DEFAULT_CACHE_NAME = 'c488ea01-8c52-4a18-a93e-934bc77f1eb8';
-
-async function fetchCachedBlobUrl(url: string, cacheName: string, clean: boolean): Promise<string> {
-	// delete old cache entries
-	if (clean) await clearCacheStorage(url, cacheName);
-
-	const cacheStorage = await caches.open(cacheName || DEFAULT_CACHE_NAME);
-	const response = await cacheStorage.match(url);
-	if (response) {
-		const blob = await response.blob();
-		return URL.createObjectURL(blob);
-	} else {
-		const response = await fetch(url);
-		// clone the response stream as it can only be consumed again
-		cacheStorage.put(url, response.clone());
-		const blob = await response.blob();
-		return URL.createObjectURL(blob);
-	}
-}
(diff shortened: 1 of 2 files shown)


DanielJDufour:
Sure! I can take a look! I think the crux of it will be trying to find a way for the workbox cache to apply to the iframe's context.  I'll try to think creatively, including maybe loading workbox inside the iframe.  I'll know more once I dive in! Thanks for getting this started!

DanielJDufour:
I'm working on this currently.  I think we want to look into pre-caching 'https://hotosm.github.io/web-forms/odk-web-form.js' and '/web-forms.html'.

For the things that depend on the config.json, which is dynamically loaded, I don't think we can technically do pre-caching in the service worker using a manifest.  Instead, I'll look at trying to do "preemptive" (is that the right word?) runtime caching where we cache right when the project is loaded (and not waiting for a map feature to be selected).

spwoodcock:
> I'm working on this currently.  I think we want to look into pre-caching 'https://hotosm.github.io/web-forms/odk-web-form.js' and '/web-forms.html'.
> 
> For the things that depend on the config.json, which is dynamically loaded, I don't think we can technically do pre-caching in the service worker using a manifest.  Instead, I'll look at trying to do "preemptive" (is that the right word?) runtime caching where we cache right when the project is loaded (and not waiting for a map feature to be selected).   In other words, try to load the resources as soon as we can (after config.json loaded).

Nice, sounds like a plan! Its likely the cache config in the vite config for the precaching 😃

DanielJDufour:
@spwoodcock , yes.  I think it's solvable.  Worse case scenario is if I can't get the elegant vite.config.json approach to work to precache the files in the service worker.  I can just force an early runtime cache, by calling fetch on the resources when the app loads.  I'll give myself to Wednesday to get the more elegant approach to work.  If it doesn't work then, I can take the more hacky approach.

Issue #2724, opened 2025-07-23
False error message when submission status is updated


manjitapandey:
**Which server**

- [x] Dev
- [x] Stage
- [ ] Prod
- [ ] Other (please specify)

**Describe the bug**
In submission details page, When a validator updates submission status to approved or has issues, there is error message saying "error in updating submission status". But the status is updated when you get back on submission table page. This wrong information confuses users. 

**To Reproduce**

1. Go to 'submission details page'
2. Click on 'Review submission status'
3. Mark submission as approved or has issues
4. See error (see failed to update submission status message and 
https://api.stage.fmtm.hotosm.org/submission/update-review-state?project_id) failing with 500 error. 

**Expected behavior**
If the submission status is being updated, the false error message as well as failing endpoint should be removed to ensure clear workflow. 

**Screenshots**

<img width="1918" height="922" alt="Image" src="https://github.com/user-attachments/assets/afad3e7b-4b07-49d4-b0e8-2b5925efbb14" />

**Additional context**

PR #2761, opened 2025-08-04
Fixed false error on submission status update.

Sujan167:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- #2724 

## Describe this PR

- Updated the ReviewStateOut Pydantic response model to allow deviceId: Optional[str] = None.

- The ODK submission may not include deviceid metadata if it wasn't declared in the XLSForm

## Review Guide

### DIFF
--- a/src/backend/app/submissions/submission_crud.py
+++ b/src/backend/app/submissions/submission_crud.py
@@ -20,13 +20,13 @@
 import asyncio
 import json
 import uuid
-import xml.etree.ElementTree as ET
 from datetime import datetime, timedelta, timezone
 from io import BytesIO
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Optional
 
+from defusedxml import ElementTree
 from fastapi import HTTPException, Response
 from loguru import logger as log
 from psycopg import Connection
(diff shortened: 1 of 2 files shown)


Sujan167:
> Did you check why deviceid is none in some cases and is it expected to be none?

Yes, I investigated this. Here are the reasons why deviceID can be null, and this is expected behavior in some cases:

ODK Central stores deviceId as null if it's not included in the submission metadata. This is a normal fallback behavior.

If the XLSForm doesn't explicitly include deviceiD as a metadata field, ODK Collect won't capture or submit it.

And the schema doesnot expect deviceID as input.
`
class ReviewStateIn(BaseModel):
    """Update to the review state of a submission."""

    instance_id: str
    review_state: ReviewStateEnum
`

src/backend/app/submissions/submission_crud.py
```diff
@@ -251,6 +255,16 @@ async def create_new_submission(
             )
 
 
+def extract_device_id_from_xml(xml_str: str) -> Optional[str]:
+    """Extract device ID from the XML string."""
+    try:
+        root = ET.fromstring(xml_str)
```
spwoodcock, line 261: Its always a bad idea to directly load an XML sent to a web API - we would first need to run this through a package like defusedxml to avoid XXE attacks or similar.

We have tried to avoid manipulating XML as much as possible in the backend.

But in this case, ODK expects the deviceId in the XML and this is the best option I think.

So we probably need to keep the code the same, but add the defusedxml package and import ElementTree from defusedxml instead of standard lib XML 👍

spwoodcock (changes requested):

spwoodcock:
Good debugging to find this info! 🙌

Anuj-Gupta4 (approved):

PR #2786, opened 2025-08-19
Only add translation for the set language in mandatory fields

Anuj-Gupta4:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- https://github.com/hotosm/field-tm/issues/2627
- https://github.com/hotosm/field-tm/issues/2374

## Describe this PR

- Only add relevant translations using label field for mandatory fields.
- In case of bare label, don't add translations of other languages and use English label for injected fields.
- Update translation check test case.
- Check if we have the translation for all labels present and raise an error if not present.
- Update the xlsform documentation.

### DIFF
--- a/docs/manuals/xlsform-design.md
+++ b/docs/manuals/xlsform-design.md
@@ -176,8 +176,11 @@ overview of the injected fields and their purposes:
 4. If no feature is selected, the user would be prompted to take a GPS coordinate
    of new feature.
    Note: One of these two options must be filled up to proceed.
-5. We also dedicate few rows for calculating OSM ID,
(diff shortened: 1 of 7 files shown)


spwoodcock:
> In that case, should we rely on label fields instead of default language to only add relevant translations for mandatory fields?

Yeah I think we might have to scan the columns and extract all the languages the user provided 👍

Sujan167 (approved):

Issue #2717, opened 2025-07-22
Add Default Fonts to Repo


DanielJDufour:
**Which server**

- [X] Dev
- [ ] Stage
- [ ] Prod
- [ ] Other (please specify)

**Describe the bug**
Add HOTOSM to mapper code's assets, so they are always available for offline development.

**To Reproduce**
Go to mapper.  No default font is set.

**Expected behavior**
Go to mapper project and should see font files loaded locally and used in app by default.

**Screenshots**
N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**
From previous conversation:
Font files are here: https://github.com/hotosm/ui-design/tree/fa06e8b31151d3b8e792b93dd660424349436af2/assets/fonts
[8:13](https://hotosm.slack.com/archives/C0738H93AKT/p1753143199270279?thread_ts=1753025903.942949&cid=C0738H93AKT)
An example on how to load them is here:
https://github.com/hotosm/ui-design/blob/fa06e8b31151d3b8e792b93dd660424349436af2/css/hot-font-face.css (edited) 
[8:15](https://hotosm.slack.com/archives/C0738H93AKT/p1753143326533359?thread_ts=1753025903.942949&cid=C0738H93AKT)
You should use fonts as variables, like it’s defined here:
https://github.com/hotosm/ui-design/blob/fa06e8b31151d3b8e792b93dd660424349436af2/css/hot.css

PR #2794, opened 2025-08-21
Font add to mapper frontend

NSUWAL123:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI
- [x] ❓ Other (style)

## Related Issue
- Related to #2717

## Describe this PR
- Barlow font add to mapper frontend

## Screenshots
image


NSUWAL123:
@spwoodcock @DanielJDufour is this the intended approach?

spwoodcock:
Thanks Varun!

For this to work @NSUWAL123 we might need to upgrade the version of hotosm/ui to the latest 👍

spwoodcock:
Ah this makes sense! Sorry I forgot to think that we removed the re-exported components.

Replacing any `hot-xxx` prefixes with `sl-xxx` should fix this.

For the UI lib we are actually using Web Awesome, which is basically Shoelace v3 (a rebrand).

If there are still problems after changing the prefixes, then keeping the banner placement where it is would be fine for now 👍 (Varun was working on integrating the UI lib & we can address it a bit later on)

varun2948:
@NSUWAL123  we can sit on a call in your office time and figure out things if it is not working for u

PR #2828, opened 2025-09-16
QField Cloud integration

### LINKED ISSUE #2318: Optimise new geometry collection experience based on research into other apps

- There are countless other apps that do the collection of new geometries in the field better: Mergin Maps, QField to name two that are essential 'mobile QGIS'.
- We should install these apps and gather information about the user experience of collecting new geometries.
- No doubt our simple UI to do this could be optimised.
- We also need to look into how we can improve the mobile experience for drawing polygons and polylines via TerraDraw.

- A much nicer user experience for directly drawing new geometries: points, polygons, polylines.
- The ability to collect a geopoint based on the users current GPS location.

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Closes #2318

## Describe this PR

- Updated the project creation page to select app: QField, ODK, FieldTM
- Integration of QFieldCloud SDK
- Project creation from FieldTM --> QFieldCloud
- Configuration of QField project ready for mapping

required fixes
- [ ] reset fieldMappingApp if project deleted / project create component is unmounted

## Screenshots

image

## Review Guide


spwoodcock:
Thanks for the feedback!

Mathieu from QField really pushed me in this direction and luckily there is some great inspiration to draw from QFieldCloud to help this implementation (its great that they will maintain the QGIS container instead of us, as they are very good at that).

For your questions:

- You are exactly right regarding the mapping app selection reset. Also currently if the user creates a project, then goes to create a second, they will still have the mapping app selected. It just needs to be unset when the project creation is navigated away from (unmounted).

- I think the creation flow can be basically the same as for ODK. I was thinking for the 'FieldTM' flow we should significantly simplify it to provide basically everything as default. The ODK Collect flow can allow customisation. Then the QField flow is essentially the same as ODK Collect perhaps with some minor tweaks.

- I'll probably make a PR with the follow up configuration to improve the mapping experience (there are some obvious things like task splitting and font size etc). Once we have the basics in place, I agree we should make issues for future refinements and from user feedback 😃

-

Anuj-Gupta4:
Could you share the local configurations required to test these changes, or add it as default values if none are passed?

NSUWAL123:
image (1)

spwoodcock:
> Could you share the local configurations required to test these changes, or add it as default values if none are passed?

I think they are all in the example dotenv file, hopefully enough not get it working 😃

I'll be back from the field in 15 mins or so and can send via slack otherwise

spwoodcock:
> image (1)
> 

I think that means the QGIS project created fine, but the QFieldCloud connection couldn't be made to upload. It will be dotenv config related surely

spwoodcock:
> @spwoodcock, I was thinking of slightly modifying the Field mapping app selections step. Instead of setting the value on the redux state, it would be good if we store the value on the form state itself. 
> 
> I am on midway making the changes, I won't able to complete it by today, so I'll will push the changes tomorrow asap. 
> 
> Also, the PR is not breaking any project creation workflow, but one thing I noticed is that `field_mapping_app` value is passed during stub project creation and final project patch, is it necessary to pass `field_mapping_app` during the project patch at the last step? Currently, if we refresh the page, `null` is being passed 

You're right in both cases! The frontend part was a bit rushed and lazy 😅

Form state 👍
Stub creation with only with the extra var (not needed on patch) 👍

Sujan167:
I'm still running into the same issue.
Is there anything else I need to configure when running this?

It looks like the error might be happening because the backend is trying to fetch ODK Central entities for a QField project, but QField projects don't have associated datasets in ODK Central.

Also, just to clarify, are we creating a corresponding project in ODK as well when we create one in QField?

One more thing: I'm also facing an issue with the `QFIELDCLOUD_TOKEN`.
Does the system generate this automatically, or do we need to create it manually?

I tried creating one through the admin panel, but ran into this error:

RelatedObjectDoesNotExist at /admin/authentication/authtoken/add/
AuthToken has no user.

spwoodcock:
If you have a valid username and password set, the token should be set automatically on server start =)

Regarding 'issues' - can you share any more details?
If you are at the point of attempting to load ODK Entities, then you are past the project creation step right?

**Just to clarify, the workflow isn't**: create QField project --> collect data via field tm --> view project in field tm
**It should be**: create QField project --> view project in QFieldCloud --> Collect data in QField --> view data synced to QGIS (once the project is made, FieldTM is essentially not used)

nirvn:
@spwoodcock , I'm loving what I see here, we should catch up over a video chat, I'd love to get a demo of this :smile_cat:

spwoodcock:
> @spwoodcock , I'm loving what I see here, we should catch up over a video chat, I'd love to get a demo of this :smile_cat: 

That would be fantastic! I'm getting back from the field next week & we could arrange something 😃

Issue #2866, opened 2025-10-18
Unable to edit the project boundary, issue with project area


Sujanadh:
**Which server**

- [ ] Dev
- [ ] Stage
- [x] Prod
- [ ] Other (please specify)

**Describe the bug**
There are few issues with the project boundary during project creation :-
- When user uploads a boundary larger than 100 sq km, for example 300 sq km, there comes a warning message to upload area within 100 sq km but still I am able to proceed to next step
- When I move to further steps and try to generate data extracts, there is an error which says I can't extract the data extracts with boundary area greater than 200 sq km
- Now, I want to edit the boundary but there is not any option to edit the boundary

**To Reproduce**

1. Go to project creation
2. Upload boundary greater than 200 sq km
3. See the warning
4. Go to next steps and try generating data extracts
5. See error that says to upload area less than 200 sq km

**Expected behavior**
If data extracts is allowed with the boundary area within 200 sq km, we should allow users to upload the area unless it is greater than 200 sq km. Having a warning message to upload an area under 100 sq km is good, but if the area is greater than 200 sq km, we should avoid uploading that area in the very first place.

**Screenshots**

<img width="1915" height="880" alt="Image" src="https://github.com/user-attachments/assets/ff53ca36-a5b2-441d-9116-03ab12c26a18" />

<img width="1917" height="879" alt="Image" src="https://github.com/user-attachments/assets/49f2bb04-932a-4b2d-a4c9-fd71878b7963" />

Slight error in the error text (should be "is not supported")

PR #2876, opened 2025-10-28
feat(createProject): set maximum aoi limit to 200 sq.km.

NSUWAL123:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
- Related to #2866

## Describe this PR
Set maximum limit of AOI area to 200 sq.km. instead of 1000 sq.km.

## Screenshots
image
image

### DIFF
--- a/src/frontend/src/components/CreateProject/01-ProjectOverview.tsx
+++ b/src/frontend/src/components/CreateProject/01-ProjectOverview.tsx
@@ -124,10 +124,10 @@ const ProjectOverview = () => {
 
     if (unit !== 'km²') return;
 
-    if (area > 1000) {
+    if (area > 200) {
       dispatch(
         CommonActions.SetSnackBar({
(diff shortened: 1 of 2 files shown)


spwoodcock:
This would fix the problem, but the limits were agreed upon a long time ago based on discussion, documented here: https://github.com/hotosm/field-tm/discussions/1732

I think the only way to solve this (within the previous constraints we set) is to provide a warning / dialog for the user:

"
The mapping area is very large >200km2. Could you consider subdividing into smaller projects?

Note that getting data from OSM will not work for areas >200km2, so you will have to provide your own data.

Confirm: Yes / No
"

If they continue, block OSM extraction.

PR #2918, opened 2025-11-13
Change entity color based on review.

Anuj-Gupta4:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

- https://github.com/hotosm/field-tm/issues/2651#issuecomment-3521599062

## Describe this PR

In ODK Collect workflow, change entity status based on submission review from Field-TM.
Also change the entity color on ready state from gray to darker gray and change border color to black.

## Screenshots

image

### DIFF
--- a/src/backend/app/central/central_crud.py
+++ b/src/backend/app/central/central_crud.py
@@ -50,11 +50,6 @@ from app.projects import project_schemas
 from app.s3 import strip_presigned_url_for_local_dev
 
 STATUS_VISUALS = {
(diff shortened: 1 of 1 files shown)


spwoodcock (commented):
This is great, but perhaps the stroke could be removed entirely imo.

Also, what about points and lines? I think they use different params

Issue #2990, opened 2025-12-16
Update area-splitter convex hull approach to StraightSkeleton algorithm


spwoodcock:
**Is your feature request related to a problem? Please describe.**

- We previously used voronois, then convex hulls: https://hotosm.github.io/fmtm-splitter/splitting-algorithm/#4-enclose-the-buildings
- The [StraightSkeleton](https://postgis.net/docs/CG_StraightSkeleton.html) algorithm looks very promising on testing, for dividing up the buildings cleanly, without jaggy edges:

<img width="1650" height="820" alt="Image" src="https://github.com/user-attachments/assets/7ceec67a-2d04-4a7e-b8bf-d8cd8a8ccc21" />

**Describe the solution you'd like**

- Test out replacing the current implementation in area-splitter with a StraightSkeleton approach.
- This can be tested and visualised easily using QGIS, by connecting to a PostGIS database and iterating on a layer.
- Criteria for success:
  - Buildings successfully divided into clusters based on average number of buildings parameter.
  - The clusters are clearly split into task areas using an algorithm, with straight edges, clear demarcation, and no overlap with geometry areas.

PR #2999, opened 2025-12-22
feat(area-splitter): Use Straight Skeleton algorithm to split tasks.

amitdkhan-pg:
This replaces the previous Voronoi algorithm.

- [x]   🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes : Update area-splitter convex hull approach to StraightSkeleton algorithm #2990 

## Brief description

The Voronoi algorithm creates edges that intersect buildings especially when
they are dense. The Straight skeleton algorithm is specifically for getting rid
of this problem, because it inherently has the property that the buildings won't
be intersected.

The idea is to run Straight skeleton algorithm on the empty space between the
buildings. The end result is that each building is surrounded by a polygon
which extends to an adjacent polygon which is similarly generated for an
adjacent building. This way, we get exactly one polygon for each building. So
all such polygons of a given clusterid merged together form a single polygon
that contains only buildings of that cluster.

## Details

There are two issues with this algorithm:

1. The PostGIS function for this algorithm has a limitation: It aborts if there
are any entities that touch each other (point or line touch). This has
something to do with a fundamental aspect with the way the lines are moved. So
it's not like a problem that can be trivially fixed in PostGIS or the
underlying SFCGAL or CGAL library.

So we fix this by removing minute patches of regions surrounding these touching
parts, making space for the splitting lines to pass through them.

2. It is quite slow compared to Voronoi.  It is observed that running straight
skeleton on (the empty space between) all the buildings together is quite slow.

So the idea is to reduce the entities on which to run the algorithm. We do this
by running it on hulls each containing a cluster of buildings rather than on
all the individual buildings. This speeds it up by roughly 7 times for clusters
containing 20 buildings.

Regardless of whether the hulls are convex or concave, there are chances that
they would intersect each other and also with the buildings of the neighbouring
cluster's buildings. So first clip the convex hull by its intersections and
then merge back any protruding buildings of it own. This will result in a
disjoint hull containing its and only its buildings. This hull itself would
usually serve as a task polygon and so in most of the cases we would be done
with splitting work; just that the task polygons won't share their border which
should be fine. But, in some cases there might be one or more buildings that
are completely taken out due to the clipping so they are not merged back into
the clipped hulls. Hence, we anyway require a straight skeleton algorithm to be
run in order to merge back the separated buildings with their original hulls.

So now the touching issue needs to be taken care of for these hulls rather than
on the buildings. This may happen if there are buildings at the hull borders
touching each other.

-------------------

Miscellaneous changes:

Moved the contents of 5-simplify.sql into avg_building_voronoi.sql because it
was meant to simpify the jagged lines produced by the Voronoi algorithm and is
not needed for Straight Skeleton. So this file was anyways specific to Voronoi.

Moved the index creation of taskpolygons from 5-simplify.sql to 6-alignment
because it is independent of the algorithms.

## Screenshots

Screenshot

## Test files

[Kathmandu_aois.tar.gz](https://github.com/user-attachments/files/24292051/Kathmandu_aois.tar.gz)

As mentioned above, initially the Straight Skeleton algo was attempted directly
on the buildings without creating the convex hulls. But that slowed down the execution.

## Review Guide

### DIFF
--- a/src/backend/packages/area-splitter/area_splitter/splitter.py
+++ b/src/backend/packages/area-splitter/area_splitter/splitter.py
@@ -452,7 +452,7 @@ class FMTMSplitter:
             "common/2-group-buildings.sql",
             "common/3-cluster-buildings.sql",
             "straight_skeleton.sql",
-            "common/6-alignment.sql", # TODO: Fix the file numbers
+            "common/6-alignment.sql",  # TODO: Fix the file numbers
             "common/7-extract.sql",
         ]
 


amitdkhan-pg:
Thanks for reviewing and merging this @spwoodcock 🙏. Do let me know if you find any issues.

amitdkhan-pg:
It seems that the PostgreSQL table ways_poly does not have these particular buildings, so those two buildings are not "visible" to the area splitter. See the attached screenshot. Note that the building polygons are not present for those particular two buildings, although the OSM layer has them. In fact, at the bottom right many buildings that are seen on OSM do not have the building polygons generated.
Screenshot from 2026-01-14 11-10-23


Sam, the blue-bordered building polygons in your screenshot are generated by what way?

amitdkhan-pg:
Screenshot from 2026-01-15 09-40-01

Attached screenshot has two layers of QGIS:
- Green layer is your kathmandu_extract json file.
- On top of it, is the Orange layer which is the 'ways_poly' table.

Most of the green layer is masked by the orange layer. But there are some green buildings you see because there are no orange buildings on top of them. So some of the green buildings do not have a corresponding ways_poly table row, including the particular two buildings through which the task borders cross. This means, we have a bug in generating the ways_poly table. Some buildings from geojson file didn't go into the ways_poly table.

PR #3000, opened 2025-12-23
Replace FastAPI with LiteStar & JSON responses with HTMX

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR

- Litestar isn't so different to FastAPI.
- It has a lot more built in and some really nice patterns.
- It's a bit stricted on return types on endpoints etc.
- But one very nice thing is the DTO pattern where we no longer need so many copies of the same pydantic models. Instead we extend from the primary DbXXX model and use `excludes=` or `includes=` to determine which fields to included. These DTOs are only using for model serialisation on input and output from end points. All db queries use the primary db model.

For now I only did the migration, keeping all the same JSON endpoints for the existing frontend.

In future, we can start to add HTMX endpoints for HTMX responses instead of JSON, making the whole app a bit simpler into the future 🤞 

## Review Guide

### DIFF
--- a/src/backend/app/db/models.py
+++ b/src/backend/app/db/models.py
@@ -325,6 +325,8 @@ class DbProject:
     odk_token: Optional[str] = None
     # GeoJSON data extract stored directly in database (replaces S3 URL approach)
     data_extract_geojson: Optional[dict] = None
+    # GeoJSON task areas/boundaries stored directly in database
+    task_areas_geojson: Optional[dict] = None
 
     @classmethod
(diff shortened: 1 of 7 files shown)


spwoodcock:
https://github.com/hotosm/field-tm/discussions/1685

PR #3028, opened 2026-03-05
Replace custom auth with hotosm_auth pkg + web component from hotosm/login

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Following on from PR https://github.com/hotosm/login/pull/7

## Describe this PR

- Replace custom auth implementation with hotosm_auth package + web component (from hotosm/login).
- The web component needs to be integrated into hotosm/ui ideally, but it also works fine with this setup adding a button in the top right header slot & then the `hotosm-auth` component in a dialog / modal.
- I added a local docker compose config for bundling hanko login during local development.

Would appreciate if someone working on this could give me a hand - thank you!! 🙏 

## AI Tool Usage

- [ ] No AI tools were used

**If AI-assisted:**

- Tool(s) used: Opus 4.6 + Codex 5.3
- What was generated: Refactored existing auth --> hotosm_auth pkg.
- What you reviewed and changed: Reviewed the code myself.

## Review Guide

To run FieldTM with Hanko login:

```bash
# install `just`
just config generate-dotenv
# you only need the backend - it's HTMX
just start login
```

I get past CORS and receive 200 responses, but then something seems to fail with the hanko config - I stopped debugging once reaching this point.

`http://mail.field.localhost:7050` is a mailpit instance that receives the sent emails.

## Checklist

- [x] PR is focused and small
- [x] Tests are included or updated
- [x] Commit messages are descriptive
- [x] Related docs and screenshots are updated


emi420:
@spwoodcock I suggest that you implement a Hanko standalone auth for FieldTM like we did with ChatMap, so this project can be installed as an instance separated from HOT if someone wants to. Then we can easily add support for shared login, once we finish the implementation that we're currently doing on other HOT tools.

One question I have for FieldTM is if you will offer users the possibility of recovering existing account/data from the old FieldTM? we are doing that for uMap for example, so when people login with the new HOT Accounts for first time, they can choose to connect to OSM once and recover their existing account and data.

spwoodcock:
Thanks 👍

No there won't be account recovery, so we can start from scratch

spwoodcock:
I implemented standalone hanko for local dev debugging, as requested!

image

emi420:
Great! I would not use "HOT Account" in this case, because it will be confusing for people. This is standalone authentication, not part of the HOT shared accounts yet. Same with the message "Access to all HOT tools and services" , I would change that.

emi420:
In ChatMap we're currently using standalone auth mostly for testing, but this is how it looks so you can get an idea:

Screenshot 2026-03-08 at 5 13 24 PM

spwoodcock:
Based on feedback in slack, this isn't the way the lib is supposed to work.
We need to sit down and discuss the architecture of the combined login to better understand.

But the key takeaway way:
- The web component `hotosm-auth` is used by the login.hotosm.org service.
- FieldTM actually creates a redirect to login.hotosm.org, where the login is complete.
- Then the login service redirects back to FieldTM, plus cookie for API auth.

spwoodcock:
https://github.com/user-attachments/assets/78439af7-cf03-4307-8167-ab718bf6a619

emi420:
The web component to be used by the tools is `@hotosm/hanko-auth`

https://github.com/hotosm/login/tree/main/auth-libs/web-component

spwoodcock:
Ah right! I saw that component, but searched npm and assumed it was the login component! Thanks 🙏

spwoodcock:
It works!

image

At least it succeeds for `/session/validate`.

Then when it calls `/api/profile/me` it gets a CORS error (although the response is correct with user profile data).
Solved by https://github.com/hotosm/login/pull/15

spwoodcock:
Going to merge this early, continue to some on dev, then come back to this when there is a chance to finalise the auth together (but we are most the way there!)

PR #3046, opened 2026-03-31
feat(monitoring): filter robots.txt noise from sentry traces and events

### LINKED ISSUE #3040: Also add robots.txt to sentry exclusions
- And fix the readme coverage badge + app image in readme

Sjorza:
  
  - [x] 🍕 Feature
  - [ ] 🐛 Bug Fix
  - [ ] 📝 Documentation
  - [ ] 🧑‍💻 Refactor
  - [ ] ✅ Test
  - [ ] 🤖 Build or CI
  
  ## Related Issue
  
  Fixes #3040
  
  ## Describe this PR
  
 This PR addresses the Sentry noise issue by filtering out requests to `/robots.txt`. 

### Changes
- Added `traces_sampler` to `/monitoring.py` to ignore `/robots.txt` transactions.
- Added `before_send` hook to drop any error events triggered by `/robots.txt`.

### Note on README tasks
I investigated the broken links mentioned in the original issue:
- The **splash image** (`screenshot-mapper.jpeg`) appears to be missing from the current repository structure.
- The **coverage badge** link is returning a 404 on the documentation server.

  
  ## AI Tool Usage
  
  - [ ] No AI tools were used
  
  **If AI-assisted:**
  
  - Tool(s) used: Gemini / ChatGPT
  - What was generated: They helped me write clean code and correctly implement the Sentry filtering logic for `monitoring.py`.
  
  ## Screenshots
  
  If applicable.
  
I considered only using `traces_sampler`, but adding `before_send` ensures that both performance traces and potential error events from `/robots.txt` are suppressed, providing a cleaner Sentry dashboard.
  
  ## Review Guide
  
The changes are isolated to `/monitoring.py`. You can test this by triggering a request to `/robots.txt` and verifying no new transactions or errors appear in Sentry.
  
  ## Checklist
  
  - [x] PR is focused and small
  - [ ] Tests are included or updated
  - [ ] Commit messages are descriptive
  - [ ] Related docs and screenshots are updated
  

Sjorza:

spwoodcock (changes requested):
Interesting approach!

It's quite verbose for something that should be simple though.

There is a `OTEL_PYTHON_EXCLUDED_URLS` env var that is configured here: 
https://github.com/hotosm/field-tm/blob/a77877e95896ee7d5cd54f1775edb67506956695/src/backend/app/config.py#L103-L114

I think it should be read by sentry, but we should double check.
Either way, it probably does the same thing configuring that env var as the code in this PR 😄 

Sorry for the hassle!

Sjorza:
Thanks for the suggestion! I completely overlooked that environment variable in `config.py`, but it’s definitely a much cleaner way to handle this.

I've refactored the PR to use `OTEL_PYTHON_EXCLUDED_URLS` and reverted the changes in `monitoring.py` as you suggested. Let me know if this works for you!

src/backend/app/monitoring.py
```diff
@@ -99,69 +99,24 @@ async def send_wrapper(message: dict[str, Any]) -> None:
...
-
-    if url:
-        path = urlparse(url).path
-        if path in IGNORED_PATHS:
-            return None
-
-    return event
-
-
-def set_sentry_otel_tracer(dsn: str) -> None:
```
spwoodcock: The updated version seems like a regression from this, removing the graceful fallback during local development - otherwise I think we might get an error

spwoodcock (changes requested):
I think only the small change in config.py is needed actually!

Thanks so much for taking this on 🙏 And sorry for the back and forth / hassle implementing - really appreciate it =)

src/backend/app/monitoring.py
Sjorza: I’ve reverted `monitoring.py` to its original state. The PR should now only contain the `config.py` change.

To be honest, the specific regression isn't entirely clear to me yet, but I've restored the file to ensure the original logic is back in place.

spwoodcock: Thank you =)

During local development we don't want Sentry to run, so there is logic to ensure that it should only run in production.

spwoodcock (approved):

spwoodcock:
@all-contributors add @Sjorza for code

Issue #3095, opened 2026-06-09
We need a task assignment UI


spwoodcock:
- It's clear from field testing that most field campaigns actually operate on a 'assign in advance' basis, rather than an 'assign in the field' approach.
- We need a UI to display task areas on a map.
- The task can be assigned to specific users / groups via comment.
- Task can be grouped for assignment in bulk to a group (i.e. higher level task division, say 4 groups, one NE one SW one SE one NW section, with many assigned tasks within each quadrant).
- The UI would load the tasks geojson from the fieldtm db, then all edits would be made in-browser (localstorage perhaps?). Once complete, the user 'confirms' the assignments, then this will pull the qfieldcloud geopackage, add task assignment fields, then push back to qfieldcloud. For ODK this would be updating the entity properties for the task.
- This approach also facilitates field mapping via tools like CoMaps / Organic Maps, as we can do the assignment - similar to field papers, but tasks that are more field-appropriate - then send the tasks into the app alongside the data.

spwoodcock:
Related https://github.com/hotosm/field-tm/issues/3064

kartikey54:
hey @spwoodcock 
thanks for promoting this good cause. hopefully this helps. cheers.

Issue #3064, opened 2026-04-14
Add CoMaps as a downstream mapping tool


spwoodcock:
Follow on from https://github.com/hotosm/field-tm/issues/3057, which wasn't viable.
This one could be.

**Is your feature request related to a problem? Please describe.**

- CoMaps is a fork of OrganicMaps, created as a community-led app that has fully transparent governance (unlike OrganicMaps, owned by a company without clear intentions).
- CoMaps can **only map POIs**, not complex features. The data goes to OpenStreetMap directly (directly via their linked OSM account and changeset on device).
- CoMaps is simple to use though - many people like maps.me and OrganicMaps, from which it is based on. It's familiar to many (although they might need to download the app if they use one of the other two mentioned tools currently).
- It could become a downstream mapping tool, as long as we can:
  1. Load a task geometry into CoMaps, so the user has indication of where they need to map.
  2. Have a thin coordination layer built into field-tm, allowing users to download a `.kmz` bookmark file for CoMaps, that contains their given task area to map.

> [!NOTE]
> Key conditions to use CoMaps:
> - If you don't have drone imagery or high-res satellite imagery for digitisation.
> - If you need a simple mapping workflow, that doesn't involve complex surveys, or beefier tools like QField.
> - If the user base is already familiar and comfortable using maps.me / OrganicMaps / CoMaps.
(otherwise, it may be more suitable to use QField / ODK).

**Describe the solution you'd like**

1. Add CoMaps as a mapping app:
  - During project creation, all of the steps continue as normal, until the 'Finalization' step. At which we don't really need to do much here, as there is no external app packaging.
  - By default, we don't select existing geometries (force mapping from scratch), as the geom layers are stored directly in CoMaps instead (the user has to download them).
  - On the project details page, show the task geometries on a map.
  - Users can select one (or multiple) task area(s) to download the CoMaps file for (this will be a `.kmz` bookmark file, containing the task area(s) that are rendered on the screen in CoMaps (simply giving a visual indication of where they need to map).
  - We could **consider** adding task locking in FieldTM, but this complicates things. Perhaps we can have users tag their username to a specific task area, or an informal team name.

2. When users map:
  - They download the kmz file from the FieldTM project page.
  - The task area is loaded up on their map.
  - (we must ensure the user has downloaded the most up to date map data for their region in CoMaps).
  - The user can then either (1) update data on existing POIs, within their task (2) add new POIs, within their task (there is no hard locking preventing them mapping outside).
[...]

PR #3097, opened 2026-06-14
feat: task assignment UI for project managers (#3095)

kartikey54:
## What

A first, **"assign in advance"** task assignment UI for project managers, addressing #3095. Managers get a panel on the project details page that shows task areas on a map, lets them select tasks and apply an assignee + group label (with one-click quadrant auto-grouping for the NE/SW/SE/NW bulk case), and persists assignments as feature properties on the existing `task_areas_geojson`.

This is opened as a **draft for discussion** — it implements the durable "edit in the browser, confirm to save" half of the issue. The field-device sync (QFieldCloud geopackage / ODK entity push) and the CoMaps `.kmz` export (#3064) are deliberately deferred to follow-up PRs (see _Deferred_ below). I'd value a steer on the open questions before going further.

## How it fits the existing architecture

No SPA, no build step, no new dependencies, no schema change — entirely server-rendered LiteStar + HTMX + Jinja2 with a flat static JS file, per `AGENTS.md` and the existing decisions.

Two commits:

1. **`fix(backend): stabilize task_id across save and export paths`** — a small prerequisite refactor. Task ids were assigned positionally and divergently in the QField (`_build_tasks_geojson`) and ODK (`_build_task_entities`) export paths, so the two ecosystems could disagree on task identity. This stamps a persistent, collision-safe integer `task_id` into `task_areas_geojson` features on save and makes both export builders read the stored id. This is the keystone the assignment feature keys on, and is independently useful — happy to split it into its own PR if you'd prefer.

2. **`feat(htmx): task assignment map panel for project managers`** — the panel itself:
   - New `app/htmx/assignment/` module with three `project_manager`-gated routes: a lazy-loaded panel fragment, a JSON `assignments/geojson` endpoint (tasks merged with `assigned_to` / `assigned_group`), and a `POST` save.
   - `save_task_assignments` service alongside `save_task_areas`: validates the group label (`^[a-zA-Z0-9_-]*$`), caps/sanitises the assignee, rejects unknown task ids, merges into feature properties under a `FOR UPDATE` row lock, and **never writes `status`** — task state stays owned by the field apps.
   - `task_assignment.js` (flat file, reuses the global Leaflet): click / shift-click selection shared between map and table, quadrant auto-grouping, draft edits staged in `sessionStorage` (matching the existing `ftm-draft-aoi-outline` precedent) and replayed over fresh server data, all user values rendered via `textContent` only.
   - Confirm warnings on the data-extract / re-split steps when stored assignments would be discarded.

## Screenshots

**Assignment panel** — task areas on a map, quadrant-coloured by group, with the summary table below:

Assignment panel

**Selecting tasks** — click / shift-click on the map and table, then apply an assignee + group to the selection:

Selecting tasks

**After saving** — assignments persisted to feature properties and reflected in the table:

After saving

## Tests

34 new tests (`tests/htmx/test_assignment_routes.py`, plus task-id stability coverage in `test_finalize_odk.py` / `test_qfield_routes.py`): route registration and auth, geojson shape and defaults, the save happy path (incl. `HX-Trigger` payload and that `status` is never written), XSS escaping, validation rejects (bad group label, overlong assignee, unknown task id, missing payload), the empty/`{}` no-split sentinel, and cross-builder task-id agreement. Local backend suite: **305 passed, 0 failed** (the two ODK-Central / QGIS-container suites were not run locally).

## Deliberately deferred (happy to sequence as follow-ups)

- **ODK entity sync** — push `assigned_to` / `assigned_group` to the `tasks` dataset entities (most machinery already exists in `central_crud`).
- **QFieldCloud sync** — pull `tasks.gpkg`, merge assignment fields by `task_id`, push back; plus stopping `_add_plugin_task_fields` from clobbering assignments on regeneration.
- **CoMaps / Organic Maps `.kmz` export** (#3064).
- **Field status read-back** (showing live task progress on the manager's map) — depends on the sync work above.

## Open questions

1. **Storage shape:** assignments as `task_areas_geojson` feature properties (this PR) vs a relational `task_assignments` table (audit trail, row-level locking). I went properties-first to keep this PR schema-free; would you want the table from the start?
2. **Permissions:** nothing in the codebase appears to write `user_roles`, so `project_manager` only resolves for global admins in deployments today. Should project creators get a project-admin role at creation?
3. **Assignee identity:** free-text `assigned_to` (covers account-less mappers / shared QFC accounts) with opportunistic QFC-collaborator suggestions — acceptable, or should assignment drive per-mapper QFC user provisioning?
4. **Groups:** is the free-text `assigned_group` label the right model for the quadrant use case, or did you envision groups of *users*?
5. **Single-task projects** (`task_areas_geojson == {}`): this PR shows a "split first" callout rather than materialising a synthetic single task — OK?

---

Refs #3095. Related #3064.

### DIFF
--- a/src/backend/app/helpers/geometry_utils.py
+++ b/src/backend/app/helpers/geometry_utils.py
@@ -160,6 +160,54 @@ def get_featcol_dominant_geom_type(featcol: dict) -> str:
     return max(geometry_counts, key=lambda key: geometry_counts[key])
 
 
+def _valid_task_id(task_id) -> bool:
+    """A valid task id is a positive integer (bools excluded)."""
+    return isinstance(task_id, int) and not isinstance(task_id, bool) and task_id > 0
+
+
+def stamp_missing_task_ids(featcol: dict) -> dict:
+    """Stamp a stable integer ``task_id`` onto features lacking a valid one.
+
+    A valid id is a unique positive integer; digit-strings are normalized to
+    int in place. Features whose id is missing, malformed (non-int, zero,
+    negative, bool), or a duplicate of an earlier feature's id are re-stamped
+    with the smallest unused positive integers (starting from 1), so the
+    collection always ends up with unique integer ids.
+    """
+    features = featcol.get("features", []) if isinstance(featcol, dict) else []
+
+    # First pass: normalize digit-strings and reserve valid first-seen ids
+    existing_ids = set()
+    for feature in features:
+        properties = feature.get("properties")
+        if not isinstance(properties, dict):
+            properties = {}
+            feature["properties"] = properties
+        task_id = properties.get("task_id")
+        if isinstance(task_id, str) and task_id.isdigit():
+            task_id = int(task_id)
+            properties["task_id"] = task_id
+        if _valid_task_id(task_id):
+            existing_ids.add(task_id)
+
+    # Second pass: re-stamp anything missing, malformed, or duplicated
+    next_id = 1
+    seen_ids = set()
+    for feature in features:
+        properties = feature["properties"]
+        task_id = properties.get("task_id")
+        if _valid_task_id(task_id) and task_id not in seen_ids:
+            seen_ids.add(task_id)
+            continue
+        while next_id in existing_ids:
+            next_id += 1
+        properties["task_id"] = next_id
+        existing_ids.add(next_id)
+        seen_ids.add(next_id)
+
+    return featcol
+
+
 async def check_crs(  # noqa: C901
     input_geojson: dict,
 ):
--- a/src/backend/app/htmx/assignment/__init__.py
+++ b/src/backend/app/htmx/assignment/__init__.py
@@ -1,0 +1,1 @@
+"""HTMX task assignment route modules."""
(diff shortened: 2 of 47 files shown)


kartikey54:

spwoodcock:
Cool! Thanks @kartikey54!

Sorry for the slow review, I have been on holiday. I will check this ASAP and give feedback =)

caiotheodoro:
Read through the task_id stability fix, `save_task_assignments`, and `assignment_routes.py` - the `FOR UPDATE` locking is correct (psycopg's async connections default to `autocommit=False`, so the lock holds until the later `commit()`), and the `<option value="{{ name }}">` QFC-suggestion rendering is safe since `JinjaTemplateEngine` sets `autoescape=True`. No blocking issues in the core files.

On open question #2: confirmed by grep, `user_roles` has no writer anywhere in the codebase (only read in `roles.py`/`user_crud.py`). `project_manager` requires either `db_user.is_admin` or a `PROJECT_ADMIN` row in that table, so as things stand a project's own creator can't reach this panel (or any other `project_manager`-gated route) unless someone is manually inserting rows. That's not new to this PR, but worth flagging loudly since it means the feature can't actually be exercised by its target user in a normal deployment yet - seems like the real blocker to sequence before or alongside this.

PR #3103, opened 2026-07-02
docs: document new QField project creation workflow

saminadamn:
Closes #2873

## Summary
- Adds a **comparison table** (Quick OSM Buildings vs Custom) at the top of the project creation section so managers can choose the right path at a glance
- Documents the **Quick Setup / OSM Buildings** path end-to-end: one-step AOI draw, what Field-TM auto-configures, and how mappers join via QField Cloud QR code
- Restructures the existing **Custom 5-step workflow** with the new field app choice (QField vs ODK Collect) and adds post-creation steps for each app (QField Cloud sharing, ODK Collect QR code)
- Replaces the old single-path ODK-centric instructions with the current two-path reality

## Test plan
- [ ] Read through Quick Setup section for accuracy against the live `/new/simple` page
- [ ] Read through Custom Setup steps 1–5 and verify they match the current UI
- [ ] Check QField Cloud sign-up link is correct
- [ ] Confirm markdown table renders correctly

Updated by @saminadamn

PR #3106, opened 2026-07-08
fix(splitter): handle polygons with connected rings

### LINKED ISSUE #3081: Handle polygons with connected rings in straight skeleton splitter
```
backend-1  |   File "/opt/python/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 967, in run
backend-1  |     result = context.run(func, *args)
backend-1  |   File "/opt/python/lib/python3.13/site-packages/area_splitter/splitter.py", line 1039, in split_by_sql
backend-1  |     splitter.splitBySQL(
backend-1  |     ~~~~~~~~~~~~~~~~~~~^
backend-1  |         db, algorithm, algorithm_params, osm_extract=extract_geojson
backend-1  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
backend-1  |     )
backend-1  |     ^
backend-1  |   File "/opt/python/lib/python3.13/site-packages/area_splitter/splitter.py", line 803, in splitBySQL
backend-1  |     features = self._run_split_sql_files(splitter_cursor, algorithm, params)
backend-1  |   File "/opt/python/lib/python3.13/site-packages/area_splitter/splitter.py", line 685, in _run_split_sql_files
backend-1  |     splitter_cursor.execute(sql_content)
backend-1  |     ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
backend-1  |   File "/opt/python/lib/python3.13/site-packages/psycopg/cursor.py", line 117, in execute
backend-1  |     raise ex.with_traceback(None)
backend-1  | psycopg.errors.InternalError_: Intersection does not support polygon with connected rings
```

saminadamn:
Closes #3081

## Problem
`ST_GeomFromGeoJSON` accepts polygons where an interior ring touches the exterior ring. These pass validation but cause PostGIS/GEOS to throw `Intersection does not support polygon with connected rings` when the straight skeleton SQL runs `ST_Intersection` on them.

## Fix
Wrap geometries in `ST_MakeValid()` at insert time in `db.py`:
- `insert_geoms_batch()` — covers OSM building polygons and linear features
- `aoi_to_postgis()` — covers the project AOI itself

`ST_MakeValid` converts connected-ring polygons into valid multi-polygons, which downstream SQL already handles via `ST_Dump()`.

## Test
Added a regression assertion to the existing `test_insert_geoms_batch_uses_executemany_and_json_tags` test confirming `ST_MakeValid` is present in the generated SQL.

(Replaces the incorrectly-linked #3103, which is an unrelated docs PR.)

### DIFF
--- a/src/backend/packages/area-splitter/area_splitter/db.py
+++ b/src/backend/packages/area-splitter/area_splitter/db.py
@@ -194,7 +194,7 @@ def aoi_to_postgis(conn: psycopg.Connection, geom: dict) -> None:
 
     sql_insert = """
         INSERT INTO project_aoi (geom)
-        VALUES (ST_SetSRID(ST_GeomFromGeoJSON(%s), 4326))
+        VALUES (ST_MakeValid(ST_SetSRID(ST_GeomFromGeoJSON(%s), 4326)))
         RETURNING id, geom;
     """
 
@@ -235,7 +235,7 @@ def insert_geoms_batch(
 
     query = (
         f"INSERT INTO {table_name} (geom, osm_id, tags) "
-        "VALUES (ST_SetSRID(ST_GeomFromGeoJSON(%(geom)s), 4326), %(osm_id)s, %(tags)s)"
+        "VALUES (ST_MakeValid(ST_SetSRID(ST_GeomFromGeoJSON(%(geom)s), 4326)), %(osm_id)s, %(tags)s)"
     )
     prepared = [{**row, "tags": Json(row.get("tags"))} for row in rows]
     cur.executemany(query, prepared)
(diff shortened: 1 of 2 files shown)


saminadamn:
Added scenario test coverage for the straight skeleton algorithm (`AVG_BUILDING_SKELETON`, the default `split_by_sql` algorithm) covering:

- The exact connected-ring regression from this issue
- A real (non-touching) interior hole, to confirm `ST_MakeValid` doesn't change already-valid geometry
- A single building
- Sparse/far-apart buildings
- A concave (L-shaped) AOI
- An invariant check against the real Kathmandu fixture

Assertions check geometry validity, that tasks don't overlap, and that tasks stay within the AOI, rather than exact feature counts — the algorithm only builds tasks around building clusters so it doesn't always reconstruct the full AOI (confirmed empirically, even the real Kathmandu fixture only covers ~93%).

All 44 tests in the package pass locally against the compose test stack.

spwoodcock (approved):
Really nice - thanks for all the tests too - I'll merge and try it out on local 👍

PR #3116, opened 2026-08-01
feat(splitter): add split-by-roads task splitting algorithm

### LINKED ISSUE #3107: Splitting algorithm: simply split by every road present in AOI

- Related to: https://github.com/hotosm/field-tm/issues/3018
- If the user is collecting street imagery, it makes sense to prioritise splitting along roads only (as they will walk / cycle roads to capture imagery)
- We could ideally have an algorithm that simply splits by all roads in the AOI.

- Grab all the major and minor roads in OAM.
- Do a simple `ST_Polygonize` to generate polygons inside the intersecting areas of roads?

- Areas could be simple buffered polygons around the road linestrings, split at intersections?
- Users would select a 'road' to map, instead of a polygon area, which might make more sense for projects like this?

saminadamn:
Closes hotosm/field-tm#3107

## Summary
- Adds a new `SPLIT_BY_ROADS` splitting algorithm for projects collecting street-level data (e.g. imagery), where mappers walk/cycle roads rather than visiting individual buildings.
- Reuses the existing linear-features polygonize step (`common/1-linear-features.sql`, which already unions roads/rivers/railways/aeroways/non-traversable barriers with the AOI boundary and runs `ST_Polygonize`) directly as the final output, skipping the building-clustering/alignment steps (2-5) entirely.
- New `algorithms/split_by_roads.sql` extract query, mirroring `common/6-extract.sql` but sourcing `polygonsnocount` directly, with `building_count` computed against `ways_poly` for informational purposes.
- Refactors `SplittingAlgorithm` to expose a `sql_files` property (ordered SQL pipeline per algorithm) instead of hardcoding the 6-file list inline in `splitter.py`.
- Fixes a latent bug in `_resolve_algorithm_params`: it previously raised "requires the following parameters" even when an algorithm's `required_params` was empty, if no legacy `num_buildings`/`num_enumerators` args were passed. Didn't affect existing algorithms (all have non-empty required params) but is needed for `SPLIT_BY_ROADS` (zero required params) to work correctly.
- Wires the new algorithm into `project_services.py::split_aoi`, the step4 splitting UI dropdown, and the JS param-visibility logic.
- Adds docs following the existing per-algorithm convention: `docs/algorithms/roads.md`, a short new section in `docs/pipeline.md`, a bullet in the package `README.md`, and a new section + table row in the user-facing `docs/manuals/task-splitting.md`.

## Test plan
- [x] `docker compose -f compose.test.yaml run --rm backend pytest package_tests/test_area_splitter` — 46/46 passed, including two new tests for `SPLIT_BY_ROADS` (one exercising the `_resolve_algorithm_params` fix via a monkeypatched call, one running the real SQL pipeline against the `kathmandu_extract.geojson` fixture and asserting valid Polygon/MultiPolygon output).

### DIFF
--- a/docs/manuals/task-splitting.md
+++ b/docs/manuals/task-splitting.md
@@ -65,15 +65,31 @@ Needs: AOI polygon, OSM extract, target number of tasks.
 
 [More details about how this algorithm works](https://github.com/hotosm/field-tm/blob/dev/src/backend/packages/area-splitter/docs/algorithms/straight-skeleton.md).
 
+### Split by Roads
+
+Task boundaries are exactly the polygons enclosed by roads, rivers,
+railways, aeroways, and non-traversable barriers — no building
+clustering. Each task is bounded by the streets around it.
+
+Best for street-level data collection (e.g. imagery capture) where
+mappers walk or cycle roads rather than visiting individual buildings.
+Task size isn't controllable — it follows however dense the street
+network is.
+
+Needs: AOI polygon, OSM extract with roads.
+
+[More details about how this algorithm works](https://github.com/hotosm/field-tm/blob/dev/src/backend/packages/area-splitter/docs/algorithms/roads.md).
+
 ## Which one to pick
 
-| Situation                          | Use                      |
-| ---------------------------------- | ------------------------ |
-| Very small AOI, one mapper         | No Splitting             |
-| No OSM data, or a sparse area      | Split by Square          |
-| You know how many mappers you have | Specific Number of Tasks |
-| Standard case, building-dense area | Average Buildings v2     |
-| Reproducing an older project       | Average Buildings v1     |
+| Situation                               | Use                      |
+| ---------------------------------------- | ------------------------ |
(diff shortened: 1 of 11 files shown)


spwoodcock:
Thank you!

Do you have any screenshots / tests of this working?

The docs say this type of split is suitable for street level mapping, but I would imagine that the end result it a bit confusing for that?

I think people travelling down roads to collect photos probably want to select actual roads, instead of task areas overlapping roads? You can imagine that two tasks will touch each road (as we split by roads), so which task does the user select to map a given road?

Based on that:
- Perhaps this algorithm is best suited for cities divided into blocks. People can work on a block at a time.
- The original linear features splitter only uses *major* non-traversible roads, such as highways. Perhaps for this algo we need to widen the criteria so that we capture minor city roads too?
- A street level mapping splitting algorithm would probably look somewhat different. Perhaps we simply segment all the roads, add a buffer around each road, and allow roads to be selectable as "tasks"?

PR #3118, opened 2026-08-10
feat(auth): use the shared Hanko login component in the header

### LINKED ISSUE #3038: Use a locale switcher cookie based on unified HOT login cookie
- Once global HOT login is working, the users have a preferred language defined.
- Ideally we can just pull from this on first load of FieldTM, saving the user having to set the locale manually.
- Order
  1. Manual language override cookie
  2. HOT profile preferred language
  3. Browser language
  4. English

hg1g:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [x] 🤖 Build or CI

## Related Issue

Fixes #3038

## Describe this PR

Wires Field-TM into the shared HOT login (Hanko SSO), the same way the other
HOT apps do it.

Until now the header had its own `<wa-button id="login-btn">` that redirected to
the login page by hand, and the `<hotosm-auth>` web component was hidden away in
a `display: none` div purely to validate the session after the redirect back.
This PR drops the custom button and puts `<hotosm-auth>` in the header slot
directly, so the shared component renders both states: the login button when
logged out, and the user profile dropdown with logout when logged in.

Two things run off the component's `hanko-login` event, both guarded so
navigation doesn't re-trigger them:

1. **User sync** — registers/syncs the authenticated user into the Field-TM
   database via `/api/v1/auth/profile/me`, once per browser session
   (`sessionStorage`). Same behaviour as before, just moved.
2. **Language adoption (#3038)** — reads the user's language from their HOT
   login profile and applies it via `?lang=`, which the existing locale
   middleware persists into the `ftm_locale` cookie. This is one-way
   (login → Field-TM) and only fires when the profile language *changes*: we
   remember the last value we saw in `localStorage`, so a profile change shows
   up on the next load, while a manual choice in the language switcher is
   respected as long as the profile language stays put.

On logout both flags are cleared, so a different user logging in on the same
browser gets a fresh sync and their own language.

Also included: a `deploy-testlogin.yml` workflow and `compose.testlogin.yaml`
that stand up the test environment described below, pointing at
`dev.login.hotosm.org`.

## AI Tool Usage

- [ ] No AI tools were used

**If AI-assisted:**

- Tool(s) used: Claude Code (Claude Opus 5)

## Screenshots

N/A — the header login button and profile dropdown are the shared component's
own UI, unchanged apart from the brand colour noted below.

**Theming the login button.** The obvious way to colour it was
`--login-btn-bg-color: var(--ftm-brand)`, but `--ftm-brand` is itself
`var(--hot-color-primary-600)`, and that nested indirection resolves
inconsistently once inherited into the component's shadow DOM — it fell back to
gray-600. So the CSS sets the HOT red as a literal `#d73f3f`. Only the login
button is affected; the profile dropdown keeps the component's defaults.

**Syncing language both ways.** Deliberately not done. Pushing a Field-TM
switcher choice back up to the login profile would make the central profile
mutable from every app, and two apps disagreeing would fight over it. One-way
adoption with change-detection gives the useful behaviour (your profile language
follows you into Field-TM) without that.

## Review Guide

**A live test environment is deployed from this branch:**
👉 **https://fieldtm.testlogin.hotosm.org**

It runs against `dev.login.hotosm.org` for auth, so you can log in with a dev
Hanko account and exercise the whole flow without touching production. It
redeploys automatically on every push to this branch.

Worth checking there:

- Logged out → the header shows the red **Login** button; clicking it takes you
  to the central login page and returns you to where you were.
- Logged in → the button is replaced by the profile dropdown, and logout works
  from it.
- First login → the user appears in the Field-TM database (the profile sync).
- Set a language on your HOT login profile, then reload Field-TM → the UI
  switches to it.
- Then change the language with Field-TM's own switcher and navigate around →
  your manual choice sticks, it doesn't get overwritten on every page load.
- Log out and back in as a different user with a different profile language →
  you get theirs, not the previous user's.

## Checklist

- [x] PR is focused and small
- [ ] Tests are included or updated
- [x] Commit messages are descriptive
- [ ] Related docs and screenshots are updated

### DIFF
--- a/src/backend/app/templates/layout.html
+++ b/src/backend/app/templates/layout.html
@@ -194,7 +194,15 @@
         </div>
       </wa-dropdown>
       {% if auth_enabled %}
-      <wa-button slot="auth" variant="brand" id="login-btn">{{ _("Login") }}</wa-button>
+      <!-- Shared Hanko web component: renders the login button when logged out
+           (redirects to the central login page) and the user profile dropdown
+           with logout when logged in. Same pattern as the other HOT apps. -->
+      <hotosm-auth
+        slot="auth"
+        hanko-url="{{ hanko_public_url }}"
+        login-url="{{ login_url }}"
+        lang="{{ current_locale() }}"
+      ></hotosm-auth>
       {% endif %}
     </hot-header>
 
@@ -204,13 +212,7 @@
       domain="field.hotosm.org"
     ></hot-tracking>
 
-    {% if auth_enabled %}
-    <!-- Hidden session verifier: validates the Hanko session after redirect back
-         from the login page and emits hanko-login. -->
-    <div style="display: none">
-      <hotosm-auth hanko-url="{{ hanko_public_url }}"></hotosm-auth>
-    </div>
-    {% endif %} {% block content %}{% endblock %}
+    {% block content %}{% endblock %}
 
     <script>
       // Strip ?lang= from the URL after the locale cookie middleware has set
@@ -232,15 +234,6 @@
       });
 
       {% if auth_enabled %}
-      const loginButton = document.getElementById("login-btn");
-      if (loginButton) {
-        loginButton.addEventListener("click", () => {
-          const returnTo = encodeURIComponent(window.location.href);
-          const separator = "{{ login_url }}".includes("?") ? "&" : "?";
-          window.location.href = `{{ login_url }}${separator}return_to=${returnTo}`;
-        });
-      }
-
       // Sync the authenticated user into the Field-TM database once per browser
       // session. hotosm-auth validates the session and emits 'hanko-login' on
       // every page load when a session is found; sessionStorage prevents us
--- a/src/backend/app/templates/layout.html
+++ b/src/backend/app/templates/layout.html
@@ -202,6 +202,8 @@
         hanko-url="{{ hanko_public_url }}"
         login-url="{{ login_url }}"
         lang="{{ current_locale() }}"
+        button-variant="filled"
+        button-color="danger"
       ></hotosm-auth>
       {% endif %}
     </hot-header>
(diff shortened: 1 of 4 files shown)


spwoodcock:
Nice, thanks! This will get reviewed & finalised in a week or two. Appreciate it 🙏

Could you reply to sign the CLA please too

hg1g:

spwoodcock:
recheck

src/backend/app/templates/layout.html
```diff
@@ -232,36 +236,61 @@
...
+          // we saw (localStorage) and only re-apply when it *changes*. This way
+          // a profile language change is reflected on the next reload, while a
+          // manual switcher choice is respected as long as the profile language
+          // stays the same. Applied via ?lang=, which the locale middleware
+          // persists into the ftm_locale cookie.
           try {
-            const resp = await fetch("/api/v1/auth/profile/me", { credentials: "include" });
-            if (resp.ok) {
-              sessionStorage.setItem("ftm_profile_synced", "1");
+            const profileResp = await fetch("{{ hanko_public_url }}/api/profile/me", { credentials: "include" });
```
caiotheodoro, line 268: risk: this fetch to the external Hanko profile endpoint runs on every `hanko-login` event (i.e. every page load with a session, per the comment above), not once per session. Only the reload side-effect is deduped via `ftm_profile_lang` - the network call itself isn't gated the way the profile-sync fetch above is with `ftm_profile_synced`. Worth wrapping in the same sessionStorage-once check, both to cut the extra external round-trip on every navigation and to match what the PR description claims ("both guarded so navigation doesn't re-trigger them").

caiotheodoro (commented):
One finding inline: the language-adoption fetch isn't actually guarded per navigation the way the PR description says.

